PR #1, opened 2017-11-22
Tickets/dm 11479

mfisherlevine:


python/lsst/cp/pipe/cpTask.py
```diff
@@ -0,0 +1,594 @@
...
+
+        Parameters
+        ----------
+        butler : daf.persistence.butler
+            Butler for the repo containg the eotest data to be used
+        run : string or int
+            Optional run number, to be used for repos containing multiple runs
+        """
+        self.log.info("Running eotest routines direct")
+        runs = butler.queryMetadata('raw', ['run'])
```
PaulPrice: I generally consider it a bad idea for variable names to differ by a single character, as it's easy to confuse them.

python/lsst/cp/pipe/cpTask.py
```diff
@@ -0,0 +1,594 @@
...
+        """
+        return {int(amp): gain for amp, gain in pSet.toDict().items()}
+
+    def testMethod(self):  # TODO: XXX remove this whole method, just useful for debugging/dev work
+        """TODO: Remove this docstring that only exists to remove flake8 error."""
+        self.log.warn("Test warning message")
+        print(self.config.requireAllEOTests)
+
+    # @pipeBase.timeMethod #xxx re-include this
+    def runEotestDirect(self, butler, run=None):
```
PaulPrice: This is a "super method": it's trying to do too much. In general, a method should do one thing, but this is trying to do the dispatch for each of the individual parts. That means that if you want to subclass and override this method, you need to copy a lot of functional code, which is bad.

I suggest turning this method into a collection of `if` statements that call other methods to do the dispatch:

    if self.config.doFe55:
        self.runFe55(run, ...)
    if self.config.doReadNoise:
        self.runReadNoise(run, ...)

There's a lot of code duplication between the various dispatches, which you should factor out.

You should also ask yourself if this approach is really a good idea --- do you actually run each of the individual parts on every file? But this isn't the `run` method. Where *is* the `run` method? Without it, this `CmdLineTask` won't do *anything* when run. And it looks like you're not actually iterating over individual files (more like all files of a certain kind), so `CmdLineTask` (at least, as set up by default) isn't what you want. This code (as it currently stands) is in no-man's land.

python/lsst/cp/pipe/cpTask.py
```diff
@@ -0,0 +1,594 @@
...
+        given a butler for a TS8 (raft-test) repo. It can contain multiple runs, but must correspond to
+        only a single raft/RTM.
+
+        - Run all eotest tasks possible, using the butler to gather the data
+        - Write outputs in eotest format
+
+        In order to replicate the canonical eotest analysis, the tasks should be run in a specific order.
+        This is given/defined in the "Steps" section here:
+        http://lsst-camera.slac.stanford.edu/eTraveler/exp/LSST-CAMERA/
+               displayProcess.jsp?processPath=1179
```
PaulPrice: Don't split the URL.

python/lsst/cp/pipe/cpTask.py
```diff
@@ -0,0 +1,594 @@
...
+
+        List of tasks that exist in the eotest package but aren't mentioned on the above link:
+        --------
+        # linearityTask()
+        # fe55CteTask()
+        # eperTask()
+        # crosstalkTask()
+        # persistenceTask()
+
+        # TODO: For each eotest task, find out what the standard raft testing does for the optional params.
```
PaulPrice: `TODO` needs a ticket number.

python/lsst/cp/pipe/cpTask.py
```diff
@@ -0,0 +1,594 @@
...
+        ----------
+        butler : daf.persistence.butler
+            Butler for the repo containg the eotest data to be used
+        run : string or int
+            Optional run number, to be used for repos containing multiple runs
+        """
+        self.log.info("Running eotest routines direct")
+        runs = butler.queryMetadata('raw', ['run'])
+        if isinstance(run, int):
+            run = str(run)
```
PaulPrice: Why not just always call `run = str(run)`?

python/lsst/cp/pipe/cpTask.py
```diff
@@ -0,0 +1,594 @@
...
+        if len(runs) != 1 and run is None:  # lots found and we don't know which one to choose
+            raise RuntimeError("Butler query found %s for runs. eotest datasets must have a run numbers, and"
```
PaulPrice: "a run numbers"

python/lsst/cp/pipe/cpTask.py
```diff
@@ -0,0 +1,594 @@
...
+                               " must specify which run to use if a respoitory contains multiple runs."%runs)
```
PaulPrice: Space either side of `%` operator, here and multiple places below.

python/lsst/cp/pipe/cpTask.py
```diff
@@ -0,0 +1,594 @@
...
+        elif run is not None and run not in runs:  # Could be specifying one of many, or one of one here
```
PaulPrice: Don't use `elif` after `if ... raise` --- you know the first `if` wasn't followed.

python/lsst/cp/pipe/cpTask.py
```diff
@@ -0,0 +1,594 @@
...
+            raise RuntimeError("Butler query found %s for runs, but the run specified (%s) "
+                               "was not among them."%(runs, run))
+        elif run is None:  # we know it's OK now
+            run = butler.queryMetadata('raw', ['run'])[0]
```
PaulPrice: Haven't you already done that query and saved it in `runs`?

python/lsst/cp/pipe/cpTask.py
```diff
@@ -0,0 +1,594 @@
...
+        del runs  # we have run defined now, so remove this to avoid potential confusion later
+
+        if not os.path.exists(self.config.eotestOutputPath):
+            os.mkdir(self.config.eotestOutputPath)
```
PaulPrice: Use `os.makedirs` instead of `os.mkdir`.

python/lsst/cp/pipe/cpTask.py
```diff
@@ -0,0 +1,594 @@
...
+            for ccd in ccds:
+                if 'FE55' not in testTypes:  # TODO: Remove not to test functionality
+                    msg = "No Fe55 tests found. Available data: %s"%testTypes
+                    if self.config.requireAllEOTests:
+                        raise RuntimeError(msg)
+                    else:
+                        self.log.warn(msg + "\nSkipping Fe55 task")
+                        break
+                fe55Filenames = [butler.get('raw_filename', dataId={'visit': visit,
+                                                                    'ccd': ccd})[0][:-3]
```
PaulPrice: Why the `[:-3]`?

python/lsst/cp/pipe/cpTask.py
```diff
@@ -0,0 +1,594 @@
...
+                                 for visit in butler.queryMetadata('raw', ['visit'], dataId=fe55TaskDataId)]
+                self.log.trace("Fe55Task: Processing %s with %s files"%(ccd, len(fe55Filenames)))
+                maskFiles = self._getMaskFiles(self.config.eotestOutputPath, ccd)
+                gains = self.fe55.run(sensor_id=ccd, infiles=fe55Filenames, mask_files=maskFiles)
+                gainsPropSet = dafBase.PropertySet()
+                for amp, gain in gains.items():  # there is no propSet.fromDict() method so make like this
+                    gainsPropSet.addDouble(str(amp), gain)
+                butler.put(gainsPropSet, 'eotest_gain', dataId={'ccd': ccd, 'run': run})
```
PaulPrice: Why save them as a `PropertySet`? The only reason to use that here is if you're going to need to access it from C++. Otherwise, save them as a `dict` using `PickleStorage` in the mapper policy.

python/lsst/cp/pipe/cpTask.py
```diff
@@ -0,0 +1,594 @@
...
+
+        # TODO: validate the results above, and/or change code to (be able to) always run
+        # over all files instead of stopping at the "required accuracy"
+        # This will require making changes to the eotest code.
+
+        ################################
+        ################################
+        # Run the Noise task
+        if self.config.doReadNoise:
+            # self.readNoise.config.output_dir = '/home/mfl/thisisbad/'
```
PaulPrice: Remove commented code.

python/lsst/cp/pipe/cpTask.py
```diff
@@ -0,0 +1,594 @@
...
+            self.log.info("Starting readNoise task")
+            noiseTaskDataId = {'run': run, 'testType': 'FE55', 'imageType': 'BIAS'}
+            for ccd in ccds:
+                if ('FE55' not in testTypes) or ('BIAS' not in imTypes):
+                    msg = "Required data for readNoise unavailable. Available data:\
+                           \ntestTypes: %s\nimageTypes: %s"%(testTypes, imTypes)
+                    if self.config.requireAllEOTests:
+                        raise RuntimeError(msg)
+                    else:
+                        self.log.warn(msg + "\nSkipping Fe55 task")
```
PaulPrice: Cut/paste error: a sure sign that your code should be refactored to remove the duplication.

python/lsst/cp/pipe/cpTask.py
```diff
@@ -0,0 +1,594 @@
...
+        # This will require making changes to the eotest code.
+
+        ################################
+        ################################
+        # Run the Noise task
+        if self.config.doReadNoise:
+            # self.readNoise.config.output_dir = '/home/mfl/thisisbad/'
+            # note that LCA-10103 defines the Fe55 bias frames as the ones to use here
+            self.log.info("Starting readNoise task")
+            noiseTaskDataId = {'run': run, 'testType': 'FE55', 'imageType': 'BIAS'}
```
PaulPrice: Why do images have `testType=FE55` but `imageType=BIAS`? There's a fundamental difference between Fe55 images and bias images. Wouldn't it make more sense to just have `imageType=FE55`?

python/lsst/cp/pipe/cpTask.py
```diff
@@ -0,0 +1,594 @@
...
+                    if self.config.requireAllEOTests:
+                        raise RuntimeError(msg)
+                    else:
+                        self.log.warn(msg + "\nSkipping bright pixel task")
+                        break
+                darkFilenames = [butler.get('raw_filename', dataId={'visit': visit,
+                                                                    'ccd': ccd})[0][:-3]
+                                 for visit in butler.queryMetadata('raw', ['visit'],
+                                                                   dataId=brightTaskDataId)]
+                self.log.trace("BrigtTask: Processing %s with %s files"%(ccd, len(darkFilenames)))
```
PaulPrice: "BrigtTask"

python/lsst/cp/pipe/cpTask.py
```diff
@@ -0,0 +1,594 @@
...
+                                     for visit in butler.queryMetadata('raw', ['visit'],
+                                                                       dataId=flatPairDataId)]
+                # Note that eotest needs the original filenames as that is the only place the flat
+                # pair-number is kept, so we have to resolve links and pass in the *original* paths here :(
+                # Also, there is no "flat-pair" test type, so all FLAT/FLAT imType/testType will appear here
+                # so we need to filter these for only the pair acquisitions (as the eotest code looks like it
+                # isn't totally thorough on rejecting the wrong types of data here)
+                # TODO: adding a translator to obs_comCam and ingesting this would allow this to be done
+                # by the butler instead of here.
+                flatPairFilenames = [os.path.realpath(_) for _ in flatPairFilenames if
```
PaulPrice: Don't use `_` as a bona fide variable name (it's used for throwaway variables).

python/lsst/cp/pipe/cpTask.py
```diff
@@ -0,0 +1,594 @@
...
+                        break
+                flatPairFilenames = [butler.get('raw_filename', dataId={'visit': visit,
+                                                'ccd': ccd})[0][:-3]
+                                     for visit in butler.queryMetadata('raw', ['visit'],
+                                                                       dataId=flatPairDataId)]
+                # Note that eotest needs the original filenames as that is the only place the flat
+                # pair-number is kept, so we have to resolve links and pass in the *original* paths here :(
+                # Also, there is no "flat-pair" test type, so all FLAT/FLAT imType/testType will appear here
+                # so we need to filter these for only the pair acquisitions (as the eotest code looks like it
+                # isn't totally thorough on rejecting the wrong types of data here)
```
PaulPrice: Can the camera team be persuaded to write the required data in the header?

You should explain what you mean by the "original" filename. I think you mean the filename written by the data acquisition system, which gets changed when we ingest into the data repo?

PaulPrice: Why do you need to rely on "flat1" and "flat2"? Can't you figure out pairs using the exposure time or something like that?

tests/test_cp_pipe.py
```diff
@@ -0,0 +1,69 @@
...
+
+    def testExample(self):
+        pass
+
+    @unittest.skipIf(noEotest, noEotestMsg)
+    def testImport(self):
+        import lsst.cp.pipe as cpPipe
+
+    @unittest.skipIf(noEotest, noEotestMsg)
+    def testClassInstantiation(self):
```
PaulPrice: You test for instantiation, but don't actually run anything? I guess that's better than nothing, but not by much.

ups/cp_pipe.build
```diff
@@ -0,0 +1,2 @@
+@LSST BUILD@
+build_lsst @PRODUCT@ @VERSION@ @REPOVERSION@
```
PaulPrice: You can dump this file --- it's not used any more.

ups/cp_pipe.cfg
```diff
@@ -0,0 +1,26 @@
...
+import lsst.sconsUtils
+
+# Dependencies that provide header files and or libraries should be included here.
+# Pure-Python dependencies do not need to be included.
+# Packages that use pybind11 or boost_tests should declare them as build dependencies.
+# Otherwise, the rules for which packages to list here are the same as those for
+# table files.
+dependencies = {
+    "required": ["utils", "ndarray"],
+    "buildRequired": ["boost_test", "pybind11", "ndarray", "pex_exceptions"],
```
PaulPrice: You need to revise these lists. Only include those that you use directly.

ups/cp_pipe.table
```diff
@@ -0,0 +1,19 @@
...
+# version number.
+# Very common third-party packages (boost, python, swig, doxygen) 
+# and very low-level LSST packages can be assumed to be recursively
+# included by low-level LSST packages such as utils or daf_base.
+# Any other package whose interface is used should be listed explicitly
+# rather than assuming it will be included recursively.
+setupRequired(boost >= 1.47.0)
+setupRequired(utils >= 4.6.0.0)
+setupRequired(pex_exceptions >= 4.6.0.0)
+setupRequired(ndarray)
```
PaulPrice: You need to revise the list of packages you require. Only include those that you use directly.
Get rid of the `>=` and everything following.

COPYRIGHT
```diff
@@ -0,0 +1,22 @@
+#
```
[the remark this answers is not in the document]
mfisherlevine: This comes straight from the LSST template repo. I think it's to facilitate pasting into python files, but I see no harm in keeping it as-is.

python/lsst/__init__.py
```diff
@@ -0,0 +1,26 @@
...
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+# 
+# You should have received a copy of the LSST License Statement and 
+# the GNU General Public License along with this program.  If not, 
+# see <https://www.lsstcorp.org/LegalNotices/>.
+#
+
+import pkgutil, lsstimport
```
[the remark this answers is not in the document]
mfisherlevine: Huh. I didn't write this line, and there are 64 other instances of this in the stack. I _guess_ I should change it though?!

https://github.com/search?p=7&q=org%3Alsst+%22import+pkgutil%2C+lsstimport%22&type=Code&utf8=%E2%9C%93

python/lsst/cp/pipe/SConscript
```diff
@@ -0,0 +1,3 @@
+# -*- python -*-
```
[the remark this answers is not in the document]
mfisherlevine: Good point. We really should clean up our template packages though, as this is confusing to see when you're creating something new. (Obviously not your job though)

python/lsst/cp/pipe/cpTask.py
```diff
@@ -0,0 +1,594 @@
...
+import os
+import glob
+
+
+class CpTaskConfig(pexConfig.Config):
+    """Config class for the calibration products production (CP) task."""
+
+    fe55 = pexConfig.ConfigurableField(
+        target=sensorTest.Fe55Task,
+        doc="The eotest Fe55 analysis task. Should not be retragetted.",
```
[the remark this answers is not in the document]
mfisherlevine: Apologies, these will be re-targeted (or "retragetted" :P), so I've just removed that sentence.

python/lsst/cp/pipe/cpTask.py
```diff
@@ -0,0 +1,594 @@
...
+    doFlatPair = pexConfig.Field(
+        dtype=bool,
+        doc="Run the flatPair task?",
+        default=True,
+    )
+    eotestOutputPath = pexConfig.Field(
+        dtype=str,
+        doc="Path to which to write the eotest output results. Madatory runtime arg for running eotest.",
+        default='',
+    )
```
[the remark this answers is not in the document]
mfisherlevine: I had a way of generating this at run-time from the butler, but it was very hacky, and more importantly, because it needs to be set in all the subTasks at instantiation-time rather than run-time there really isn't a way around this as they're frozen.

I agree it's not good like this, but this won't be run from the command line, and nobody will suffer needlessly from this because running eotest is something that won't be done much at all.

python/lsst/cp/pipe/cpTask.py
```diff
@@ -0,0 +1,594 @@
...
+    requireAllEOTests = pexConfig.Field(
```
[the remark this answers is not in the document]
mfisherlevine: Again, I agree it's not great like this, but this won't be run from the command line.

python/lsst/cp/pipe/cpTask.py
```diff
@@ -0,0 +1,594 @@
...
+                log.warn("OVERWRITING: Found a user defined output path of %s for %sTask. "
+                         "This has been overwritten with %s, as individually specified output paths for "
+                         "subTasks are not supported at present"%(getattr(self, task).output_dir,
+                                                                  task, self.eotestOutputPath))
+            getattr(self, task).output_dir = self.eotestOutputPath
+
+
+class CpTask(pipeBase.CmdLineTask):
+    """
+    Calibration Products Production (CP) task.
```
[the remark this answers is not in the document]
mfisherlevine: Everyone hates using CPP due to C++, see RFC-415 (and your own comments). Have changed this to read "Calibration (Products) Production (CP) task." to mitigate slightly though 🙂

python/lsst/cp/pipe/cpTask.py
```diff
@@ -0,0 +1,594 @@
...
+
+    This task is used to produce the calibration products required to calibrate cameras.
+    Examples of such operations are as follows:
+    Given a set of flat-field images, find the dark pixels and columns.
```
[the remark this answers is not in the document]
mfisherlevine: Just because each of those tasks potentially does more than just what it says in the description, and I don't want to be obligated to say every little thing it does.

python/lsst/cp/pipe/cpTask.py
```diff
@@ -0,0 +1,594 @@
...
+    Given a set of darks, find the bright pixels and columns.
+    Given a set of Fe55 exposures, calulate the gain of the readout chain, in e-/ADU
+    Given a set of Fe55 exposures, calulate the instrinsic PSF of the silicon, and the degradation of
+    the PSF due to CTE.
+    Given a set of flat-pairs, measure the photon transfer curve (PTC).
+    Given a set of bias frames, calculate the read noise of the system in e-.
+    Given a set of pocket-pumping exposures, find charge-traps in the silicon.
+
+    The CpTask.runEotestDirect() is only applicable to LSST sensors, and only for a specific type of dataset
```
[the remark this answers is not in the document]
mfisherlevine: Because `eotest` is the analysis suite for the LSST-specific electro-optical tests run at BNL for sensor acceptance testing for the LSST project, and therefore doesn't concern itself with anything else. It is the polar-opposite of camera-agnostic.

python/lsst/cp/pipe/cpTask.py
```diff
@@ -0,0 +1,594 @@
...
+
+    See http://ls.st/ldm-151 Chapter 4, Calibration Products Production for further details
+    regarding the inputs and outputs.
+    """
+
+    ConfigClass = CpTaskConfig
+    _DefaultName = "cp"
+
+    def __init__(self, *args, **kwargs):
+        """
```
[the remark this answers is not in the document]
mfisherlevine: Was just copying what I'd seen elsewhere, but point taken, have removed.

python/lsst/cp/pipe/cpTask.py
```diff
@@ -0,0 +1,594 @@
...
+        **kwargs :
+            a dictionary of keyword arguments passed on to the Task constructor
+        """
+        pipeBase.CmdLineTask.__init__(self, *args, **kwargs)
+
+        # Note - we can't currently call validate on the subTask configs, as they are NOT valid
+        # due to state of eotest. However, we override validate() and call it here
+        # and use it to set the output dir config parameter in the subTasks.
+        self.config.validate()
+        self.config.freeze()
```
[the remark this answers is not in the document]
mfisherlevine: Only it's called from the command line.

python/lsst/cp/pipe/cpTask.py
```diff
@@ -0,0 +1,594 @@
...
+        self.makeSubtask("readNoise")
+        self.makeSubtask("brightPixels")
+        self.makeSubtask("darkPixels")
+        self.makeSubtask("traps")
+        self.makeSubtask("cte")
+        self.makeSubtask("flatPair")
+        self.makeSubtask("ptc")
+
+        # Do we want to switch to the better-named logger below or use the one we get from cmdLineTask?
+        # e.g. using self.log = lsstLog.Log.getLogger("cp.pipe.cpTask")?
```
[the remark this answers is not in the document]
mfisherlevine: It's not too bad, but it loses all the prefixes and just calls it cpTask I think, but sure, this should be OK if this is the done thing.

python/lsst/cp/pipe/cpTask.py
```diff
@@ -0,0 +1,594 @@
...
+
+    def _getMaskFiles(self, path, ccd):
```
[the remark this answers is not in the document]
mfisherlevine: Yes, this is indeed the intent. Nobody should ever need to call this from outside here.

python/lsst/cp/pipe/cpTask.py
```diff
@@ -0,0 +1,594 @@
...
+        as more and more tests run, and saves having to have clever logic for if some tasks fail.
+        Parameters
+        ----------
+        path : string
+            Path on which to find the mask files
+        ccd : string/int
+            Name/identifier of the CCD
+        Returns
+        -------
+        maskFiles : list/tuple
```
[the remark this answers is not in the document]
mfisherlevine: Nice 🙂

python/lsst/cp/pipe/cpTask.py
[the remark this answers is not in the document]
PaulPrice: All the more reason to jettison `CmdLineTask`.

python/lsst/cp/pipe/cpTask.py
```diff
@@ -0,0 +1,594 @@
...
+    def _cleanupEotest(self, path):
+        """Delete all the medianed files left behind after eotest has run.
+
+        Running eotest generates a lot of interim medianed files, so this just cleans them up.
+        Parameters
+        ----------
+        path : string
+           Path on which to delete all the eotest medianed files.
+        """
+        for filename in glob.glob(os.path.join(path, '*median*')):
```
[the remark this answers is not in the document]
mfisherlevine: Good call, I made it `'*_median_*.fits'`, that the narrowest it can be I think, but should be much safer.

python/lsst/cp/pipe/cpTask.py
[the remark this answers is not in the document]
mfisherlevine: Sorry, I'm not sure I follow why that adds weight to the jettisoning of `CmdLineTask`. Could you explain a little please?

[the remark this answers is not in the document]
PaulPrice: `CmdLineTask` is all about camera-agnosticism.

python/lsst/cp/pipe/cpTask.py
mfisherlevine: Huh, I could have sworn that I did that because it exceeded the 110 char limit, but apparently it doesn't. Sorry!

python/lsst/cp/pipe/cpTask.py
mfisherlevine: Indeed, but this is why I `del` it as soon as I'm done with it. Given that, I think it's OK. Shout if you disagree.

python/lsst/cp/pipe/cpTask.py
PaulPrice: Even if it exceeds the 110 char limit, you shouldn't split a URL.

mfisherlevine: 😱

mfisherlevine: Good to know though, thanks!

python/lsst/cp/pipe/cpTask.py
mfisherlevine: Because it comes in defaulted to `None`, and I don't want to turn it into `'None'`

mfisherlevine: Found a better fix though:
`if run is not None: run = str(run)`

python/lsst/cp/pipe/cpTask.py
mfisherlevine: I get where you're coming from, I just wanted these to be considered one logical block, and was using the `elif`s to tie them together. I felt strongly about the tying together, and so have rewritten this section to make it one piece, and more coherent/readable into the bargain.

python/lsst/cp/pipe/cpTask.py
mfisherlevine: Yes, but without the `[0]`. Also though, given the comment above, this code no longer exists 🙂

python/lsst/cp/pipe/cpTask.py
mfisherlevine: I don't really know why, but the butler returns a list of filenames which looks like:
`['/path/to/file/filename1.fits[0]', '/path/to/file/filename2.fits[0]']`, so that seems to be necessary 😬

python/lsst/cp/pipe/cpTask.py
mfisherlevine: There's two things to note here. The first is that the intent of the people writing the files was to note that these are bias frames taken during an Fe55 test run, so the _testType_ is Fe55, but the _imageType_ is BIAS.

The second point to note is that this code exists to interface with a pre-defined dataset structure which we have no control over, and so this is just dealing with the way things are.

python/lsst/cp/pipe/cpTask.py
mfisherlevine: The camera team can't change anything, they are in "production mode" and nothing will be changed at this point.

I have improved the explanation of what is meant by "original filename".

You could indeed work it out like that, but that is not how the 3rd party code works, and this code just exists to feed those tasks.

python/lsst/cp/pipe/cpTask.py
mfisherlevine: Apologies, it's a Robertism I caught but which I quite like(d). Very useful in py2 to prevent accidental use (due to scope-leakage), but fully pointless in py3 🙂

tests/test_cp_pipe.py
mfisherlevine: Yeah, I know, it's just really hard to run anything without a ton of test data, so I spoke to some locals and we thought that at least setting things up was better than nothing.

ups/cp_pipe.cfg
mfisherlevine: Sorry about this, this is another instance of our template package misleading people.

python/lsst/cp/pipe/cpTask.py
mfisherlevine: I had no idea about {{PickleStorage}} and have switched to it, it's a much better method for this, trying to force it into a {{PropertySet}} was never something I _wanted_ to do! (Seems that virtually nobody knows about this, there are only 3 uses anywhere in the stack!)

PR #4, opened 2018-07-17
First cut at the task

mfisherlevine:


python/lsst/cp/pipe/bfCoeffCalc.py
```diff
@@ -0,0 +1,1272 @@
...
+
+            if self.debug.enabled:
+                try:
+                    frameId1 = exp1.getMetadata().get("FRAMEID")
+                    frameId2 = exp2.getMetadata().get("FRAMEID")
+                    frameId = '_diff_'.join(frameId1, frameId2)
+                except:
+                    frameId = 'Im1 diff Im2'
+
+            # depending on level this is either one pass or n_amps
```
r-owen: This is cryptic and duplicates the comment on the next line. and, to some extent, the comment at the top of the loop
I suggest just one comment that is clearer, or one here and one at the top of the loop. I think you are trying to say something like this (though I may have it wrong and you certainly should reword it as you see fit):

    # Compute cross-correlation and means at the appropriate config.level:
    # - "DETECTOR": compare the two visits
    # - "AMP": compare each amplifier of one visit to ... (what? each amplifier of the other visit? the same amplifier of the other visit?)

python/lsst/cp/pipe/bfCoeffCalc.py
```diff
@@ -0,0 +1,1272 @@
...
+        parser = pipeBase.ArgumentParser(name=cls._DefaultName)
+        parser.add_argument("--visitPairs", help="The list of visit pairs to use, as a list of tuples "
+                            "enclosed in quotes e.g. \"(123,456),(789,987),(654,321)\""
+                            "NB: must be comma-separated-tuples with no spaces, enclosed in quotes!")
+        parser.add_id_argument("--id", datasetType="bfKernelNew", ContainerClass=BfDataIdContainer,
+                               help="The ccds to use, e.g. --id ccd=0..100")
+        return parser
+
+    @pipeBase.timeMethod
+    def run(self, dataRef, visitPairs):
```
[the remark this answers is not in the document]
yalsayyad: Can confirm.  This method should be renamed to `runDataRef` if you want to this to work as a  CmdlineTask after DM-2639 gets merged. (Remember that all "dataRef" means in `runDataRef` is Gen2 butler objects vs Gen3 butler objects.)

python/lsst/cp/pipe/bfCoeffCalc.py
```diff
@@ -0,0 +1,1272 @@
...
+    xcorrCheckRejectLevel = pexConfig.Field(
+        dtype=float,
+        doc="Sanity check level for the sum of the input cross-correlations. Arrays which "
+        "sum to greater than this are discarded before the clipped mean is calculated.",
+        default=1.0  # xxx change this back once the problem is fixed!
+        # default=0.2
+    )
+    maxIterSOR = pexConfig.Field(
+        dtype=int,
+        doc="The maximum number of iterations allowed for the successive over-relaxation method",
```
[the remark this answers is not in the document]
mfisherlevine: Yes, agreed on all fronts. Have changed all instance of `SOR ` to `SuccessiveOverRelaxation`, and changed the private method from `_SOR()` to public `successiveOverRelax()` and put the description in its docstring. If you feel this should be moved elsewhere in the codebase let me know where and I can put it there, though I feel that there is likely to be pushback if it goes in one of the appropriate places as it's a numerical method with potentially quite a few iterations, implemented in python (and probably not even very efficiently at that), so would prefer to keep this hidden away in here for now if possible.

python/lsst/cp/pipe/bfCoeffCalc.py
```diff
@@ -0,0 +1,1272 @@
...
+        default=10000
+    )
+    eLevelSOR = pexConfig.Field(
+        dtype=float,
+        doc="The target residual error for the successive over-relaxation method",
+        default=5.0e-14
+    )
+    nSigmaClipKernelGen = pexConfig.Field(
+        dtype=float,
+        doc="Number of sigma to clip to during pixel-wise clipping when generating the kernel",
```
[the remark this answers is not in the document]
mfisherlevine: I've changed the `_generateKernel()` to be public (for a few reasons) and have said to see its docstring for more info. I hope that's sufficient.

python/lsst/cp/pipe/bfCoeffCalc.py
```diff
@@ -0,0 +1,1272 @@
...
+        default=4
+    )
+    nSigmaClipXCorr = pexConfig.Field(
+        dtype=float,
+        doc="Number of sigma to clip when calculating means for the cross correlation",
+        default=5
+    )
+    maxLag = pexConfig.Field(
+        dtype=int,
+        doc="The maximum lag to use when calculating the cross correlation/kernel",
```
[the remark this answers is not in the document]
mfisherlevine: Yes, sorry, it was obvious to me, but perhaps not those unfamiliar. It's correlating two images in detector-space, so the answer is just pixels. Have added, and will tweak the docstring too.

python/lsst/cp/pipe/bfCoeffCalc.py
```diff
@@ -0,0 +1,1272 @@
...
+        default=8
+    )
+    nPixBorderGainCalc = pexConfig.Field(
+        dtype=int,
+        doc="The number of border pixels to exclude when calculating the gain",
+        default=10
+    )
+    nPixBorderXCorr = pexConfig.Field(
+        dtype=int,
+        doc="The number of border pixels to exclude when calculating the cross correlation/kernel",
```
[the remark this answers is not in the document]
mfisherlevine: No no, quite right, "and" makes more sense.

python/lsst/cp/pipe/bfCoeffCalc.py
```diff
@@ -0,0 +1,1272 @@
...
+        default=10
+    )
+    biasCorr = pexConfig.Field(
+        dtype=float,
+        doc="A scary, empirically determined correction-factor correcting for sigma-clipping" +
+        " a non-Gaussian distribution",
```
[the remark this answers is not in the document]
mfisherlevine: I've done the first bits (the phrasing etc), but the explanation is a bit tricky. Essentially this is something that probably shouldn't be touched except by experts, but I made it a config option because having a hard-coded number in the code filled me with horror, especially as I can imagine this being different in some cases. How it's calculated should indeed be added, and probably methods for doing so added to this module, and therefore those referenced here, but given that hasn't been done yet, I'm going to have to punt on this for now.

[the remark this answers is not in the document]
mfisherlevine: I think what I'll do is ticket adding that code to this, and add the ticket number to this docstring, saying that it should be update once DM-XXXXX is done. Does that sound like a reasonable solution?

python/lsst/cp/pipe/bfCoeffCalc.py
```diff
@@ -0,0 +1,1272 @@
...
+        dtype=bool,
+        doc="Contrain the fit of the PTC to go through the origin?",
+        default=True
+    )
+    level = pexConfig.ChoiceField(
+        doc="The level at which to calculate kernels",
+        dtype=str, default="CCD",
+        allowed={
+            "AMP": "Every amplifier treated separately",
+            "CCD": "One kernel per CCD",
```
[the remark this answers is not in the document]
mfisherlevine: Haha, I hope this too! It was me commenting that on RFC-484, and who made those changes to `obs_lsstCam`, so this is quite amusing that I went wrong here 😛

So yes, totally agree, and only 68 instance to change in this file 🤪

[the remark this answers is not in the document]
mfisherlevine: This might, however, lead to some weirdness/slight inconsistency, as the dataIds will still contain CCD as most obs_packages haven't been changed yet. I'll change it where I can, and we'll see where we end up is probably best. I'll let you know when it's time to rereview.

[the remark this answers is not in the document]
mfisherlevine: OK, this has now been done. I think all the places that ccd remains are more or less appropriate. It's slightly ironic that the convention we've chose to allow for dynamic keying of this is `ccdKey` rather than `detectorKey`, but changing this here and not elsewhere is both asking for trouble and well beyond the scope I think. If there are any instances you still object to do shout though.

python/lsst/cp/pipe/bfCoeffCalc.py
```diff
@@ -0,0 +1,1272 @@
...
+        self.config.validate()
+        self.config.freeze()
+
+    @classmethod
+    def _makeArgumentParser(cls):
+        """Augment argument parser for the BfTask."""
+        parser = pipeBase.ArgumentParser(name=cls._DefaultName)
+        parser.add_argument("--visitPairs", help="The list of visit pairs to use, as a list of tuples "
+                            "enclosed in quotes e.g. \"(123,456),(789,987),(654,321)\""
+                            "NB: must be comma-separated-tuples with no spaces, enclosed in quotes!")
```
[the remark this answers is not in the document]
mfisherlevine: Oh. My. God. THIS! 🤯 This is the answer I was looking for all along! How did nobody else mention that I was reinventing the wheel rather than shouting at me about my lack of regex knowledge! 😄

Thank you, this is indeed the elegant (and pre-existing solution) I needed. `nargs="*"` for the win! 🎉

It looks to me like I still need a (now very slimmed down) taskRunner to pass through the argument though, is that right? It now just looks like

```
    @staticmethod
    def getTargetList(parsedCmd, **kwargs):
        """Pass the visit list through explicitly."""
        return pipeBase.TaskRunner.getTargetList(parsedCmd, visitPairs=parsedCmd.visitPairs, **kwargs)
```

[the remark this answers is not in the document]
mfisherlevine: I had a little play seeing if I could get it through with `**kwargs` but failed. Could you just take a look at 16beda01764f3370818306624f2552d9feb93a7d and check that that is the best that can be done, _i.e._ I can't get rid of the TaskRunner altogether please?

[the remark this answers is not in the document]
mfisherlevine: I tried passing them through un-parsed with `**kwargs` and doing the parsing later so that the whole TaskRunner could be done away with but couldn't get it working. Could you confirm that you think this is the best that can be done please?

python/lsst/cp/pipe/bfCoeffCalc.py
```diff
@@ -0,0 +1,1272 @@
...
+        means = {}
+        kernels = {}
+
+        # setup necessary objects
+        ccdNum = dataRef.dataId[self.config.ccdKey]
+        if self.config.level == 'CCD':
+            xcorrs = {ccdNum: []}
+            means = {ccdNum: []}
+        elif self.config.level == 'AMP':
+            # NB: don't use dataRef.get('raw_detector') due to composites
```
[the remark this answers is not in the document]
mfisherlevine: I have changed this to read:
```
            # NB: don't use dataRef.get('raw_detector') due to composites
            # because of the way composite objects are handled/constructed (i.e. LSST images)
            # these need to be retrieved from the camera and dereferenced rather than accessed directly
```
Is that a good description? Not really. But the problem is that I don't really know what causes the problem; it's likely a bug, and this is a workaround, but without work digging into some of the messiest bits of the stack it's neither fixable nor possible to give a great description of why this is. Hopefully Gen3 will improve this 😬 🤞

python/lsst/cp/pipe/bfCoeffCalc.py
```diff
@@ -0,0 +1,1272 @@
...
+            detector = dataRef.get('camera')[dataRef.dataId[self.config.ccdKey]]
+            ampInfoCat = detector.getAmpInfoCatalog()
+            ampNames = [amp.getName() for amp in ampInfoCat]
+            xcorrs = {key: [] for key in ampNames}
+            means = {key: [] for key in ampNames}
+
+        # calculate or retrieve the gains
+        if self.config.doCalcGains:
+            self.log.info('Beginning gain estimation for CCD %s'%ccdNum)
```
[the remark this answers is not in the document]
mfisherlevine: Mainly pushing back just because it seems like a reasonable amount of work given I've always done it this way, so might not nail it the first time, and as you said, harder to test, plus it will be virtually no saving, as there are very few messages, on code that will be run very infrequently. It will take many thousands of times longer to fix (probably literally billions actually) than it will save in runtime, so leaving unless you shout. Will put the spaces in though of course.

python/lsst/cp/pipe/bfCoeffCalc.py
```diff
@@ -0,0 +1,1272 @@
...
+            gains, nomGains = self.estimateGains(dataRef, visitPairs)
+            dataRef.put(gains, datasetType='bfGain')
+            self.log.info('Finished gain estimation for CCD %s'%ccdNum)
+        else:
+            gains = dataRef.get('bfGain')
+            if not gains:
+                self.log.fatal('Failed to retrieved gains for CCD %s'%ccdNum)
+                raise RuntimeError("Must either calculate or supply gains for %s"%ccdNum)
```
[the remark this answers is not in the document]
mfisherlevine: I actually don't know why I was logging a FATAL as well as raising the RuntimeError tbh. I moved the message from the log into the RuntimeError so that it's accurate, and removed the logger message. I think that should cover this.

python/lsst/cp/pipe/bfCoeffCalc.py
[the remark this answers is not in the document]
r-owen: I think having a custom task runner is very reasonable because of the need to pass extra information. As to where to parse the visit pairs (converting them from a string `"123,456"` to a pair of integers `(123, 456)`): clearly it should be done before calling the task, which I think means it has to be done as part of parsing the arguments or in the task runner. Doing it while parsing the command optimal, if you can pull it off, as that catches any problems as soon as possible and allows you to print the usual "I could not parse this command, here is the command help" message. The argument parser does this for other arguments using a callback specified by the `action` argument to `add_argument`. See `class IdValueAction` in `argumentParser.py` for an example of a callback, and how it is used. There are a number of similar classes defined nearby.

python/lsst/cp/pipe/bfCoeffCalc.py
mfisherlevine: Yes, this was just a garbage comment, thanks. It not reads:

```
            # Compute the cross-correlation and means at the appropriate config.level:
            # - "DETECTOR": one key, so compare the two visits to each other
            # - "AMP": n_amp keys, comparing each amplifier of one visit to the same amplifier in other visit```

File: python/lsst/cp/pipe/bfCoeffCalc.py (hunk 7 above)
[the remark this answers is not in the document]
Comment (r-owen) [reply python/lsst/cp/pipe/bfCoeffCalc.py]: Yes

File: python/lsst/cp/pipe/bfCoeffCalc.py [hunk 13]
```diff
@@ -0,0 +1,1272 @@
...
+    level = pexConfig.ChoiceField(
+        doc="The level at which to calculate kernels",
+        dtype=str, default="CCD",
+        allowed={
+            "AMP": "Every amplifier treated separately",
+            "CCD": "One kernel per CCD",
+        }
+    )
+
+    def validateIsrConfig(self, logger):
```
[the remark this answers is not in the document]
r-owen: If there are parts that can usefully validate the config without logging then they should probably be moved to `validate` (or a method called by `validate` in the config. But otherwise your solution seems best and it will certainly suffice.

python/lsst/cp/pipe/bfCoeffCalc.py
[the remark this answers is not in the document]
r-owen: I agree. No point logging and then raising.

python/lsst/cp/pipe/bfCoeffCalc.py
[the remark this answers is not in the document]
r-owen: That seems a very reasonable approach.

python/lsst/cp/pipe/bfCoeffCalc.py
```diff
@@ -0,0 +1,1272 @@
...
+                continue
+            self.refList += refList
+
+
+class BfTask(pipeBase.CmdLineTask):
+    """Bright-fatter effect coefficient calculation task.
+
+    See http://ls.st/ldm-151 Chapter 4, Calibration Products Production for further details
+    regarding the inputs and outputs.
+    """
```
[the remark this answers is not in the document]
mfisherlevine: Rewritten. But given that it's quite long and from scratch I imagine there will be comments on it.

python/lsst/cp/pipe/makeBrighterFatterKernel.py
```diff
@@ -0,0 +1,1303 @@
...
+    """Brighter-fatter effect correction-kernel calculation task.
+
+    A command line task for calculating the brighter-fatter correction
+    kernel from pairs of flat-field images (with the same exposure length).
+
+    The following operations are performed:
+
+    - The configurable isr task is called, which unpersists and assembles the
+      raw images, and performs the selected instrument signature removal tasks.
+      The appropriacy of the selected isr components is checked.
```
r-owen: "appropriacy" is an great word (one new to me), but the definitions I found don't work here. Perhaps it would be more helpful to say that certain ISR steps must be performed. If they can be briefly explained then do it here. Then point to the config fields that specify the checks.

python/lsst/cp/pipe/makeBrighterFatterKernel.py
```diff
@@ -0,0 +1,1303 @@
...
+
+    - The gain of the each amplifier in the detector is calculated using
+      the PTC method and used to correct the images so that all calculations
```
r-owen: Please spell out "PTC"

python/lsst/cp/pipe/makeBrighterFatterKernel.py
```diff
@@ -0,0 +1,1303 @@
...
+      are done in units of electrons, and so that the level across amplifier
+      boundaries is continuous. Outliers in the PTC are iteratively rejected
+      before fitting, with the nSigma rejection level set by
+      config.nSigmaClipRegression. Individual pixels are masked from
+      the image based on config.nSigmaClipGainCalc
+
+    - Each image is then cross-correlated with the one it's paired with
+      (with the pairing defined by the --visitPairs command line argument),
+      which is done either the whole-image to whole-image,
+      or amplifier-by-amplifier, depending on the config.level.
```
r-owen: I suggest removing "the" before "config.level", but overall this is excellent -- very clear and helpful

python/lsst/cp/pipe/makeBrighterFatterKernel.py
```diff
@@ -0,0 +1,1303 @@
...
+
+    - Once the cross-correlations have been calculated for each visit pair,
+      these are used to generate the correction kernel.
+      Outlier values in these cross-correlations are rejected by using a
+      pixel-wise sigma-clipped thresholding to each cross-correlation in
+      the visitPair-length stack or cross-correlations.
+      The number of sigma clipped to is set by config.nSigmaClipKernelGen.
+      The maximum lag used, in pixels, and hence the size of the half-size
+      of the kernel generated, is given by config.maxLag,
```
r-owen: I suggest removing "in pixels" here, since it's documented in the config

python/lsst/cp/pipe/makeBrighterFatterKernel.py
```diff
@@ -0,0 +1,1303 @@
...
+      i.e. a value of 10 will result in a kernel of size 2n-1 = 19x19 pixels.
+
+    - Once DM-15277 has been completed, a method will exist to calculate the
+      empirical correction factor, config.biasCorr.
+      TODO: DM-15277 update this part of the docstring once the ticket is done.
+    """
```
r-owen: This new task documentation is great!

python/lsst/cp/pipe/makeBrighterFatterKernel.py
```diff
@@ -0,0 +1,1303 @@
...
+            # Compute the cross-correlation and means at the appropriate config.level:
+            # - "DETECTOR": one key, so compare the two visits to each other
+            # - "AMP": n_amp keys, comparing each amplifier of one visit to the same amplifier in other visit
+            for det_object in _scaledMaskedIm1.keys():
+                xcorrs[det_object].append(self._crossCorrelate(_scaledMaskedIm1[det_object],
+                                                               _scaledMaskedIm2[det_object]))
+                means[det_object].append([_means1[det_object], _means2[det_object]])
+
+            # this is awkward now that code is refactored
+            # not sure how best to make this work without passing everything around
```
r-owen: I'm not seeing the issue; can you please explain more (here in github is fine, no need to modify the code)

python/lsst/cp/pipe/makeBrighterFatterKernel.py
```diff
@@ -0,0 +1,1303 @@
...
+        self.log.info('Finished generating kernel(s) for %s' % detNum)
+        return pipeBase.Struct(exitStatus=0)
+
+    def _prepareImage(self, exp, gains, level):
+        """Prepare images for cross-correlation calculation.
+
+        Each amp has borders applied, is rescaled by the gain, and has the sigma-clipped mean subtracted.
+        If the level is 'DETECTOR' then this is done for the whole image in-place so that it can be
+        cross-correlated.
+        If the level is 'AMP' then this is done per-amplifier, and each amp-image returned.
```
r-owen: Please consider rewriting in active voice (the usual Python standard for doc strings) e.g. "For each amp apply borders (what does that mean?) rescale by gain, and subtract sigma-clipped mean"...

python/lsst/cp/pipe/makeBrighterFatterKernel.py
```diff
@@ -0,0 +1,1303 @@
...
+        """Estimate the gains of the amplifiers in the detector using the specified visits.
+
+        Given a dataRef and list of flats of varying intensity, calculate the gain for each
+        detector specified using the PTC method.
+
+        The fixPtcThroughOrigin config option determines whether the iterative fitting is
+        forced to go through the origin or not. This defaults to True, fitting var=1/gain * mean,
+        if set to False then var=1/g * mean + const is fitted.
+
+        This is really a PTC gain measurement task. See DM-14063 for results from of a comparison between
```
r-owen: Please spell out PTC

python/lsst/cp/pipe/makeBrighterFatterKernel.py
```diff
@@ -0,0 +1,1303 @@
...
+        diff = maskedIm0.clone()
+        diff -= maskedIm1.getImage()
+        diff = diff[border:-border, border:-border]
+
+        if self.debug.writeDiffImages:
+            filename = '_'.join(['diff', 'detector', detId, frameId, '.fits'])
+            diff.writeFits(os.path.join(self.debug.debugDataPath, filename))
+
+        # Subtract background.  It should be a constant, but it isn't always
+        # xxx do we want some logic here for whether to subtract or not? Or a config option?
```
r-owen: Any chance of figuring this out before merging? If not, please file a ticket and reference it here.

python/lsst/cp/pipe/bfCoeffCalc.py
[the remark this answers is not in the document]
mfisherlevine: OK, glad I'm not doing it too wrong then. It seems like this one liner is very simple, and much more understandable (to me at least) than using callbacks etc, so as long as you don't think this needs moving elsewhere I'll just leave it as-is.

python/lsst/cp/pipe/makeBrighterFatterKernel.py
mfisherlevine: OK, will fix. But for reference, the Google definition I get is:
```
appropriacy
əˈprəʊprɪəsi/
noun
noun: appropriacy; plural noun: appropriacies

    the extent to which something is suitable or proper in the circumstances.
    "the appropriacy of the methodology employed"
```
which is the way in which I intended it, and seems to fit well, to my ears at least 🙂

mfisherlevine: I'm not sure what you mean by "the config fields that specify the checks" though, as there are none - the isr config is always checked. I guess you mean "the config fields that are specified _by_ the checks", _i.e._ which fields are checked, right? If so, I'll mention the most important and most-likely-misconfigured ones, as the list is quite long, and the code can define the rest (they live in three very self-explanatory lists after all).

python/lsst/cp/pipe/makeBrighterFatterKernel.py
mfisherlevine: I dunno... I think that it reads as it should, and might save people searching around if they're confused. I think it adds clarity, sounds right, and as it will never change having it defined in multiple places is no problem. Unless you feel strongly that it should go I'd rather leave that there. However, as this is defining the shape of the output of the whole task I've moved it up in this section, and I think the whole thing reads a lot better now. Fro convenience, this section now reads:
```
    - Once the cross-correlations have been calculated for each visit pair,
      these are used to generate the correction kernel.
      The maximum lag used, in pixels, and hence the size of the half-size
      of the kernel generated, is given by config.maxLag,
      i.e. a value of 10 will result in a kernel of size 2n-1 = 19x19 pixels.
      Outlier values in these cross-correlations are rejected by using a
      pixel-wise sigma-clipped thresholding to each cross-correlation in
      the visitPair-length stack of cross-correlations.
      The number of sigma clipped to is set by config.nSigmaClipKernelGen.
```
Let me know if that's OK, or if you want changes (or the mention of "in pixels" removed).

python/lsst/cp/pipe/makeBrighterFatterKernel.py
mfisherlevine: 😊

python/lsst/cp/pipe/bfCoeffCalc.py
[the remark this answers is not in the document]
mfisherlevine: I've done something a little different, but it now reads well, thanks for that.

python/lsst/cp/pipe/bfCoeffCalc.py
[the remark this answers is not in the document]
mfisherlevine: Yeah, I'm torn. It's not _really_ validation is the thing; they're not intrinsically invalid, they're just ensuring that people aren't Doing It Wrong™️, but they're not testing for a non-compliant pexConfig, so I think it should probably stay as-is really.

python/lsst/cp/pipe/makeBrighterFatterKernel.py
mfisherlevine: The solution here is going to be pull out all the commented code and ticket enhancing the debug plot functionality I think. But to explain though:

the issue is that to generate the correct titles for the debug plots you need the objects, but you don't want to be passing everything around just to be able to query things that aren't really used. However, a massive block of code to generate the titles doesn't really belong in the main routine, and even less so as it mostly won't even fire.

I don't think there's a nice solution to this, and we'll probably see this code reappear in this place, though hopefully a little more cleanly/concisely, if/when that ticket gets done, but for now I think just removing it is fine.

python/lsst/cp/pipe/makeBrighterFatterKernel.py
mfisherlevine: I have made it log warnings the background is found to be above a configurable level, so people in the future will know if it's subtracting something large. I think this should do it.

python/lsst/cp/pipe/bfCoeffCalc.py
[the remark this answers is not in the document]
r-owen: I'd say if it raises an exception then it's validation. We do this sort of thing in other places -- check that something was processed with required config settings before being allowed to proceed -- and we call it validation there.

However, I am not saying you have to split this into `xConfig.validate` (with no logging) and the rest of it. If it feels like a coherent bit of code to you then it's probably best to leave as a single method.

python/lsst/cp/pipe/makeBrighterFatterKernel.py
```diff
@@ -0,0 +1,1380 @@
...
+    and this class parses that, and passes the parsed version
+    to the run() method.
+
+    See pipeBase.TaskRunner for more information.
+    """
+
+    @staticmethod
+    def getTargetList(parsedCmd, **kwargs):
+        """Parse the visit list and pass through explicitly."""
+        visitPairs = [(int(visit.split(",")[0]), int(visit.split(",")[1])) for visit in parsedCmd.visitPairs]
```
r-owen: I strongly suggest better error checking and reporting.

Heres is a version with good error checking that uses the callback mechanism. I **have** tested this code and hope you will consider using it. It parses the data in the usual location, and thus produces really nice error messages that don't include distracting tracebacks.
```
class VisitPairAction(argparse.Action):
    """argparse action callback to process a visit pair into
    a tuple 2 floats
    """
    def __call__(self, parser, namespace, values, option_string):
        """Parse ``--visitPairs`` data and append results to
        ``namespace.<argument>.visitPairs``.

        Parameters
        ----------
        parser : `ArgumentParser`
            Argument parser.
        namespace : `argparse.Namespace`
            Parsed command (an instance of argparse.Namespace).
            The following attributes are updated:

            - ``visitPairs``
        values : `list`
            A list of visit pairs, each of the form "visit,visit"
        option_string : `str`
            Option value specified by the user. Ignored.
        """
        if namespace.visitPairs is None:
            namespace.visitPairs = []
        for visitStringPair in values:
            visitStrings = visitStringPair.split(",")
            if len(visitStrings) != 2:
                parser.error("Found {} visits in {} instead of 2".format(len(visitStrings), visitStringPair))
            try:
                visits = [int(visit) for visit in visitStrings]
            except Exception:
                parser.error("Could not parse {} as two integer visit numbers".format(visitStringPair))
            namespace.visitPairs.append(visits)
```
If you are really keen to parse the data in the task runner then you can easily adapt the above, but error messages will produce tracebacks and will not print help for the task.

Doing it in a loop instead of a list comprehension is slower and more verbose, but speed should not be an issue and good error reporting is important. For instance users will occasionally forget and put a space after a comma: "--visit-pairs 1, 2 3, 4" and a good error message is helpful. You could also accept that, but it would require more code -- especially when using the callback mechanism, as you'd have to keep track of "did the last value end with a comma, if so the next value should be a single integer". I don't think it is necessary or expected.

python/lsst/cp/pipe/makeBrighterFatterKernel.py
```diff
@@ -0,0 +1,1380 @@
...
+        """Compute refList based on idList.
+
+        This method must be defined as the dataset does not exist before this
+        task is run.
+
+        Parameters
+        ----------
+        namespace
+            Results of parsing command-line
+            (with ``butler`` and ``log`` elements).
```
r-owen: I think you are saying that the "butler" and "log" elements are required. If so, it would be nice to say so explicitly, e.g. as per the original class (though you could replace "set" with "present"):
```
        namespace : `argparse.Namespace`
            Results of parsing command-line. The ``butler`` and ``log``
            elements must be set.
```

python/lsst/cp/pipe/makeBrighterFatterKernel.py
```diff
@@ -0,0 +1,1380 @@
...
+      warnings are logged.
+
+    - The gain of the each amplifier in the detector is calculated using
+      the photon transfer curve (PTC) method and used to correct the images
+      so that all calculations are done in units of electrons, and so that the
+      level across amplifier boundaries is continuous.
+      Outliers in the PTC are iteratively rejected
+      before fitting, with the nSigma rejection level set by
+      config.nSigmaClipRegression. Individual pixels are masked from
+      the image based on config.nSigmaClipGainCalc
```
r-owen: I'm afraid I don't understand this last sentence. Is it practical to be a bit more specific about what calculation is used to mask these pixels? Also please document the following here: is the mask of the exposure updated by this operation? If so, which mask plane is updated (and if it's one that you add, say that as well)?

python/lsst/cp/pipe/makeBrighterFatterKernel.py
```diff
@@ -0,0 +1,1380 @@
...
+        pipeBase.CmdLineTask.__init__(self, *args, **kwargs)
+        self.makeSubtask("isr")
+
+        self.debug = lsstDebug.Info(__name__)
+        if self.debug.enabled:
+            self.log.info("Running with debug enabled...")
+            # if we're displaying, test it works and save the displays for later
+            # it's worth testing here as displays are flaky and sometimes
+            # can't be contacted, and given processing takes a while
+            # it's a shame to fail late due to display issues
```
r-owen: A very helpful comment. Please break into sentences with periods (I think the first line is one sentence
and the rest is another).

python/lsst/cp/pipe/makeBrighterFatterKernel.py
```diff
@@ -0,0 +1,1380 @@
...
+            if self.debug.display:
+                try:
+                    afwDisp.setDefaultBackend(self.debug.displayBackend)
+                    afwDisp.Display.delAllDisplays()
+                    self.disp1 = afwDisp.Display(0, open=True)
+                    self.disp2 = afwDisp.Display(1, open=True)
```
r-owen: Thanks for including the keyword

python/lsst/cp/pipe/makeBrighterFatterKernel.py
```diff
@@ -0,0 +1,1380 @@
...
+        self.checkIsrConfig()
+        self.config.validate()
+        self.config.freeze()
+
+    @classmethod
+    def _makeArgumentParser(cls):
+        """Augment argument parser for the MakeBrighterFatterKernelTask."""
+        parser = pipeBase.ArgumentParser(name=cls._DefaultName)
+        parser.add_argument("--visitPairs", nargs="*",
+                            help="Visit pairs to use. Each pair must be of the form INT,INT e.g. 123,456")
```
r-owen: Sorry I didn't catch this earlier, but all our existing command line options are lowercase with hyphens between words so please do that for the visit pairs argument. Use the dest keyword to specify the name for the parsed data, e.g.: `add_argument("--visit-pairs", dest="visitPairs", nargs="*", ...)`. This change will require a one or two tweaks to existing task documentation.

python/lsst/cp/pipe/makeBrighterFatterKernel.py
```diff
@@ -0,0 +1,1380 @@
...
+
+    def checkIsrConfig(self):
+        """Check that appropriate ISR settings are being used
+        for brighter-fatter kernel calculation."""
+        mandatory = ['doAssembleCcd', 'doOverscan']  # raise if False
+        forbidden = ['doApplyGains', 'normalizeGains', 'doFlat', 'doFringe', 'doAddDistortionModel',
+                     'doBrighterFatter', 'doUseOpticsTransmission', 'doUseFilterTransmission',
+                     'doUseSensorTransmission', 'doUseAtmosphereTransmission', 'doGuider', 'doStrayLight',
+                     'doTweakFlat']  # raise if True
+        desirableTrue = ['doBias', 'doDark', 'doCrosstalk', 'doDefect', 'doLinearize']  # WARN if False
```
r-owen: I suggest making these three fields config parameters, so an obs package can plug in the appropriate variant ISR task and tweak the tests accordingly. (I thought they already were config fields, which explains an earlier confusion -- I had suggested that a help string point to the config fields which specify these tests).

Unfortunately the field name should probably include "isr", e.g. "isrMandatorySteps", "isrForbiddenSteps" and "isrDesirableSteps" -- with a possible future "isrUndesirableSteps".

If you make this change then I also strongly suggest failing if a specified ISR config field doesn't exist (e.g. e.g. by testing and raising -- or skipping the test if the resulting error message is clear enough). Otherwise you risk failing to test something important due to a spelling error. In fact that is another argument for making the tests more visible by making them part of the config.

python/lsst/cp/pipe/makeBrighterFatterKernel.py
```diff
@@ -0,0 +1,1380 @@
...
+            means = {key: [] for key in ampNames}
+        else:
+            raise RuntimeError("Unsupported level: {}".format(self.config.level))
+
+        # calculate or retrieve the gains
+        if self.config.doCalcGains:
+            self.log.info('Beginning gain estimation for detector %s' % detNum)
+            gains, nomGains = self.estimateGains(dataRef, visitPairs)
+            dataRef.put(gains, datasetType='brighterFatterGain')
+            self.log.info('Finished gain estimation for detector %s' % detNum)
```
r-owen: Consider making this debug level. I worry about users complaining about excessive verbosity. In which case the "beginning" log message may want to be "Compute gain estimation for detector %s"

python/lsst/cp/pipe/makeBrighterFatterKernel.py
```diff
@@ -0,0 +1,1380 @@
...
+        # generate the kernel(s)
+        self.log.info('Generating kernel(s) for %s' % detNum)
+        for det_object in xcorrs.keys():  # looping over either detectors or amps
+            if self.config.level == 'DETECTOR':
+                objId = 'detector %s' % det_object
+            elif self.config.level == 'AMP':
+                objId = 'detector %s AMP %s' % (detNum, det_object)
+            kernels[det_object] = self.generateKernel(xcorrs[det_object], means[det_object], objId)
+        # writing the level might be useful, but might also mess things up xxx
+        # kernels['level'] = self.config.level
```
r-owen: Please don't leave in commented-out code (at least not without a ticket #). It is an interesting issue. It seems like a great idea to include the info, but it's weird to mix data with metadata.

python/lsst/cp/pipe/makeBrighterFatterKernel.py
```diff
@@ -0,0 +1,1380 @@
...
+            maxIter = self.config.maxIterSuccessiveOverRelaxation
+        if not eLevel:
+            eLevel = self.config.eLevelSuccessiveOverRelaxation
+
+        # initialise, and set boundary conditions
+        func = np.zeros([source.shape[0]+2, source.shape[1]+2])
+        resid = np.zeros([source.shape[0]+2, source.shape[1]+2])
+        rhoSpe = np.cos(np.pi/source.shape[0])  # Here a square grid is assummed
+
+        inError = 0
```
r-owen: I suggest deleting this assignment as it is not being used

python/lsst/cp/pipe/makeBrighterFatterKernel.py
```diff
@@ -0,0 +1,1380 @@
...
+    for i in range(L+1):
+        for j in range(L+1):
+            XCORR[i+L, j+L] = xcorr.getArray()[i, j]
+            XCORR[-i+L, j+L] = xcorr.getArray()[i, j]
+            XCORR[i+L, -j+L] = xcorr.getArray()[i, j]
+            XCORR[-i+L, -j+L] = xcorr.getArray()[i, j]
+    return xcorr, np.sum(means1)
+
+
+def simulateXcorr(fluxLevels, imageShape, addCorrelations=False, correlationStrength=0.1, nSigma=5, border=3,
```
r-owen: I don't see any use of this in the package. Can you get rid of it, along with _crossCorrelateSimulate above? If you have unresolved issues to work out then at least file a ticket and add a TODO message to the doc string for both of these functions and consider moving it to a separate module.

python/lsst/cp/pipe/makeBrighterFatterKernel.py
```diff
@@ -0,0 +1,1380 @@
...
+        return means, variances, coVars
+
+    def _plotXcorr(self, xcorr, mean, zmax=0.05, title=None, fig=None, saveToFileName=None):
+        """Used to plot the correlation functions."""
+        try:
+            xcorr = xcorr.getArray()
+        except:
+            pass
+
+        xcorr /= float(mean)
```
r-owen: You don't need the `float` in Python 3

python/lsst/cp/pipe/makeBrighterFatterKernel.py
```diff
@@ -0,0 +1,1380 @@
...
+            # note the shape of these returns depends on level
+            _scaledMaskedIm1, _means1 = self._prepareImage(exp1, gains, self.config.level)
+            _scaledMaskedIm2, _means2 = self._prepareImage(exp2, gains, self.config.level)
+
+            if self.debug.enabled:
+                try:
+                    frameId1 = exp1.getMetadata().get("FRAMEID")
+                    frameId2 = exp2.getMetadata().get("FRAMEID")
+                    frameId = '_diff_'.join(frameId1, frameId2)
+                except:
```
r-owen: Please don't use bare "except" and please run flake8

python/lsst/cp/pipe/makeBrighterFatterKernel.py
```diff
@@ -0,0 +1,1380 @@
...
+            diffAmpIm = diff[amp.getBBox()].clone()
+            diffAmpImCrop = diffAmpIm[border:-border-maxLag, border:-border-maxLag, afwImage.LOCAL]
+            diffAmpImCrop -= afwMath.makeStatistics(diffAmpImCrop, afwMath.MEANCLIP, sctrl).getValue()
+            w, h = diffAmpImCrop.getDimensions()
+            xcorr = np.zeros((maxLag + 1, maxLag + 1), dtype=np.float64)
+
+            # calculate the cross-correlation
+            for xlag in range(maxLag + 1):
+                for ylag in range(maxLag + 1):
+                    dim_xy = diffAmpIm[border+xlag:border+xlag + w, border+ylag: border+ylag + h, afwImage.LOCAL].clone()
```
r-owen: This line is too long and the formatting is confusing and not compliant with our standards (e.g. spaces around "+"). I suggest:
```
                    dim_xy = diffAmpIm[border + xlag: border + xlag + w,
                                       border + ylag: border + ylag + h,
                                       afwImage.LOCAL].clone()
```

python/lsst/cp/pipe/makeBrighterFatterKernel.py
```diff
@@ -0,0 +1,1380 @@
...
+        diff[box, afwImage.LOCAL] -= bkgd.getImageF(afwMath.Interpolate.CUBIC_SPLINE,
+                                                    afwMath.REDUCE_INTERP_ORDER)
+
+        variances = {}
+        coVars = {}
+        for amp in detector:
+            ampName = amp.getName()
+
+            diffAmpIm = diff[amp.getBBox()].clone()
+            diffAmpImCrop = diffAmpIm[border:-border-maxLag, border:-border-maxLag, afwImage.LOCAL]
```
r-owen: Please add spaces after ":" and around "-" (or before, in the case of negating `border`). I suggest a global regex search for + and - without surrounding spaces

python/lsst/cp/pipe/makeBrighterFatterKernel.py
```diff
@@ -0,0 +1,1380 @@
...
+        dataRef.dataId['visit'] = v1
+        exp1 = self.isr.runDataRef(dataRef).exposure
+        dataRef.dataId['visit'] = v2
+        exp2 = self.isr.runDataRef(dataRef).exposure
+        del dataRef.dataId['visit']
+        exps = [exp1, exp2]
+        self._checkExpLengthEqual(exp1, exp2, v1, v2)
+
+        detector = exps[0].getDetector()  # note we lose the detector in the next line as they belong to exp
+        ims = [self._convertImagelikeToFloatImage(exp) for exp in exps]
```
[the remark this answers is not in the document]
r-owen: I see that for visit pairs they had better have compatible detectors or the code will fail. So I have much less strong feelings about changing the code. But if you leave it as is then please consider deleting the code comment; I found it confusing, and once I figured out what it meant I wondered why it was present.

python/lsst/cp/pipe/bfCoeffCalc.py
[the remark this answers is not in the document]
mfisherlevine: OK, I've just renamed the method to `validateIsrConfig()` config then, and left it as it was (after the changes I made before of course, so it still lives in the main class).

python/lsst/cp/pipe/makeBrighterFatterKernel.py
```diff
@@ -0,0 +1,1380 @@
...
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the LSST License Statement and
+# the GNU General Public License along with this program.  If not,
+# see <https://www.lsstcorp.org/LegalNotices/>.
+#
```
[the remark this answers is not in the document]
mfisherlevine: 👍 to the `__all__`, but other modules put this after the imports, and indeed I get an E402 error for every module level import that follows this, as they have to be first according to that rule. Is there a reason you ask for this to be first?

python/lsst/cp/pipe/makeBrighterFatterKernel.py
mfisherlevine: Actually, I wasn't trying to say anything really, just following the original docs I think. It now just reads `Results of parsing command-line.`.

python/lsst/cp/pipe/makeBrighterFatterKernel.py
mfisherlevine: Sorry, I see how this was confusing - this is actually a comment on the input images, and I've changed it to say that. Also, nothing is masked or updated, and I've also changed the text to not use the work "mask".

python/lsst/cp/pipe/makeBrighterFatterKernel.py
mfisherlevine: I think both of those are good suggestions, and I especially like the latter. However, I will just remove for now, and if/when we decide we want that I can make a ticket and new datasets etc, as we're not really losing any code here.

python/lsst/cp/pipe/makeBrighterFatterKernel.py
```diff
@@ -0,0 +1,1380 @@
...
+    def _prepareImage(self, exp, gains, level):
+        """Prepare images for cross-correlation calculation.
+
+        For each amp, crop by the border amount, specified by
+        config.nPixBorderXCorr, then rescale by the gain
+        and subtract the sigma-clipped mean.
+        If the level is 'DETECTOR' then this is done
+        for the whole image in-place so that it can be cross-correlated.
+        If the level is 'AMP' then this is done per-amplifier,
+        and each amp-image returned.
```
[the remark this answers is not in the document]
mfisherlevine: 100% agree. I don't know why it said this. I think it's because these docs were adapted from the original code a little too blindly, this was never the case, and certainly code should never work like that!

python/lsst/cp/pipe/makeBrighterFatterKernel.py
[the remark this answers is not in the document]
r-owen: PEP8 changed at some point. It used to require that some imports went before `__all__` and then changed (I think a few years ago). Here is the current text: 
https://www.python.org/dev/peps/pep-0008/#id24 or look up PEP8 "Module Level Dunder Names"

python/lsst/cp/pipe/makeBrighterFatterKernel.py
```diff
@@ -0,0 +1,1380 @@
...
+            kernels[det_object] = self.generateKernel(xcorrs[det_object], means[det_object], objId)
+        # writing the level might be useful, but might also mess things up xxx
+        # kernels['level'] = self.config.level
+        dataRef.put(kernels)
+
+        self.log.info('Finished generating kernel(s) for %s' % detNum)
+        return pipeBase.Struct(exitStatus=0)
+
+    def _prepareImage(self, exp, gains, level):
+        """Prepare images for cross-correlation calculation.
```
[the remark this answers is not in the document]
mfisherlevine: Yeah, I agree, and I'm not too worried about the singular return length with that method name. I've gone for `_makeCroppedExposures()`.

python/lsst/cp/pipe/makeBrighterFatterKernel.py
mfisherlevine: OK, I've moved all this to configs as you said. For the checking, what I've done is to remove the protecting blocks that were in there for the forbidden/mandatory catagories so that you get an error if they're not found, I think that should suffice. This made sense too as they're now overridable per-obs_package, so no reason to be forgiving if they're not there.

If you could just take a look and check that the way I've gone about this is OK with you I think that would be good, thanks.

python/lsst/cp/pipe/makeBrighterFatterKernel.py
```diff
@@ -0,0 +1,1380 @@
...
+        ax.set_zlabel(r"$\langle{(F_i - \bar{F})(F_i - \bar{F})}\rangle/\bar{F}$")
+
+        if title:
+            fig.suptitle(title)
+        if saveToFileName:
+            fig.savefig(saveToFileName)
+
+    @staticmethod
+    def _getNameOfSet(vals):
+        """Convert a list of numbers to a string, merging consecutive values."""
```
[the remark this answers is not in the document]
mfisherlevine: Have removed all this code as it was actually unused (it only features in making titles for debug plots), and have just dumped it in DM-15305 for now.

python/lsst/cp/pipe/makeBrighterFatterKernel.py
```diff
@@ -0,0 +1,1380 @@
...
+        xcorrList = []
+        sctrl = afwMath.StatisticsControl()
+        sctrl.setNumSigmaClip(self.config.nSigmaClipKernelGen)
+
+        for corrNum, ((mean1, mean2), corr) in enumerate(zip(means, corrs)):
+            corr[0, 0] -= (mean1+mean2)
+            if corr[0, 0] > 0:
+                self.log.warn('Skipped item %s due to unexpected value of (variance-mean)' % corrNum)
+                continue
+            corr /= -float(1.0*(mean1**2+mean2**2))
```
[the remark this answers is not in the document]
mfisherlevine: I think that if the means are identically zero then things are sufficiently bad that a division by zero is OK actually.

python/lsst/cp/pipe/makeBrighterFatterKernel.py
mfisherlevine: OK, this has been tidied up significantly. The code must live on, and I think this is the right place for it, but it certainly needs a little more work, as there are some known problems. I've written a new function that unifies the generating images part and allows switching which code is called to do the calculations, and filed DM-15756 to finish up that bit of work, which will then remove the vestigial function here I think. If you have any comments on how you think that should go (including if you disagree that it should be in this module) do please leave them on the ticket, but I hope this is OK for now.

python/lsst/cp/pipe/makeBrighterFatterKernel.py
mfisherlevine: Done! (I think)

python/lsst/cp/pipe/makeBrighterFatterKernel.py
```diff
@@ -0,0 +1,1427 @@
...
+            if nIter%2 == 0:
+                for i in range(1, func.shape[0] - 1, 2):
+                    for j in range(1, func.shape[0] - 1, 2):
+                        resid[i, j] = float(func[i, j-1] + func[i, j + 1] + func[i - 1, j] +
+                                            func[i + 1, j] - 4.0*func[i, j] - dx*dx*source[i - 1, j - 1])
+                        func[i, j] += omega*resid[i, j]*.25
+                for i in range(2, func.shape[0] - 1, 2):
+                    for j in range(2, func.shape[0] - 1, 2):
+                        resid[i, j] = float(func[i, j - 1] + func[i, j + 1] + func[i - 1, j] +
+                                            func[i + 1, j] - 4.0*func[i, j] - dx*dx*source[i - 1, j - 1])
```
r-owen: Both here and elsewhere in this method you would probably get better speed and possibly code clarity by using `scipy.signal.convolve` to apply one kernel to `func` and one to `source`: `resid = func convolved with [0, 1, 0] [1, 4, 1] [0, 1, 0]] - source convolved with [[0, 0, 0] [0, 0, 0] [dx**2, 0, 0]]`. It's a 1-liner (modulo defining the two kernels, which are always the same and can be created once and re-used).

python/lsst/cp/pipe/makeBrighterFatterKernel.py
```diff
@@ -0,0 +1,1427 @@
...
+        # We perform two sweeps per cycle,
+        # updating 'odd' and 'even' points separately
+        nIter = 0
+        omega = 1.0
+        dx = 1.0
+        while nIter < maxIter*2:
+            outError = 0
+            if nIter%2 == 0:
+                for i in range(1, func.shape[0] - 1, 2):
+                    for j in range(1, func.shape[0] - 1, 2):
```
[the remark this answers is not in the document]
mfisherlevine: Fair point. I think I will add an assert though as well, because a non-square kernel sounds like madness to me, in this case at the very least.

python/lsst/cp/pipe/makeBrighterFatterKernel.py
mfisherlevine: As discussed OOB I'm going to leave this as-is for now because despite being both ugly and slow, it will only save milliseconds of compute to change it, and testing that the behaviour is identical will take a lot longer.

python/lsst/cp/pipe/makeBrighterFatterKernel.py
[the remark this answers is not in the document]
mfisherlevine: Now done.

python/lsst/cp/pipe/makeBrighterFatterKernel.py
mfisherlevine: Thanks for this, and yes, I had indeed just missed it before. I've split the difference and implemented what you put here (and many thanks for that!), but not in the argparser as this avoids callbacks which I think aren't strictly necessary and would make running the task more magic than it needs to be.

python/lsst/cp/pipe/makeBrighterFatterKernel.py
```diff
@@ -0,0 +1,1380 @@
...
+            # Do sanity checks; if these are failed more investigation is needed
+            breaker = 0
+            for amp in detector:
+                ampName = amp.getName()
+                if _means[ampName]*10 < _vars[ampName] or _means[ampName]*10 < _covars[ampName]:
+                    msg = 'Sanity check failed; check visit pair %s amp %s' % (visPair, ampName)
+                    self.log.warn(msg)
+                    breaker += 1
+            if breaker:
+                continue
```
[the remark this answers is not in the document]
mfisherlevine: I don't think this is ignored; if this is set then the continue applies to the outer-loop, _i.e._ serves to skip this visit pair, no?

python/lsst/cp/pipe/makeBrighterFatterKernel.py
```diff
@@ -0,0 +1,1380 @@
...
+
+            # having made sanity checks
```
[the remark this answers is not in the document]
mfisherlevine: Will remove if I'm wrong above, but I think it's OK...

python/lsst/cp/pipe/makeBrighterFatterKernel.py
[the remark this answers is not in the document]
r-owen: You are right. My mistake!

python/lsst/cp/pipe/makeBrighterFatterKernel.py
```diff
@@ -0,0 +1,1380 @@
...
+        # NB: don't use dataRef.get('raw_detector') due to composites
+        detector = dataRef.get('camera')[dataRef.dataId[self.config.ccdKey]]
+
+        ampMeans = {}
+
+        dataRef.dataId['visit'] = v1
+        exp1 = self.isr.runDataRef(dataRef).exposure
+        dataRef.dataId['visit'] = v2
+        exp2 = self.isr.runDataRef(dataRef).exposure
+        del dataRef.dataId['visit']
```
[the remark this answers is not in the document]
mfisherlevine: Yes, this was nasty, and I missed this, thanks for catching that. I've done it as you said now.

python/lsst/cp/pipe/makeBrighterFatterKernel.py
r-owen: I just looked at the code. The new config parameters look fine (though I'm disappointed that one still needs to be hard-coded -- assembleCcd.doTrim; if you used dotted names for subtasks you could avoid that). The code I found using these is the `validateIsrConfig` method and it still logs a warning if a config parameter in one of these lists is missing -- rather than failing. So I'm not sure why you said "you get an error if they are not found". I'd prefer an error, but I think what you have will do.

python/lsst/cp/pipe/makeBrighterFatterKernel.py
```diff
@@ -0,0 +1,1380 @@
...
+
+            means = {}
+            for amp in detector:
+                ampName = amp.getName()
+                means[ampName] = ampMeans[ampName][0] + ampMeans[ampName][1]
+
+        return means, variances, coVars
+
+    def _plotXcorr(self, xcorr, mean, zmax=0.05, title=None, fig=None, saveToFileName=None):
+        """Used to plot the correlation functions."""
```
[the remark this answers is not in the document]
mfisherlevine: Yeah, that was straight from the original code. You should have seen some of the others!

python/lsst/cp/pipe/makeBrighterFatterKernel.py
```diff
@@ -0,0 +1,1380 @@
...
+        -----------
+        corrs : `list` of `numpy.ndarray`, (Ny, Nx)
+            A list of the quarter-image cross-correlations
+        means : `dict` of `tuples` of `floats`
+            The means of the input images for each corr in corrs
+        rejectLevel : `float`, optional
+            This is essentially is a sanity check parameter.
+            If this condition is violated there is something unexpected
+            going on in the image, and it is discarded from the stack before
+            the clipped-mean is calculated.
```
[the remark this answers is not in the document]
mfisherlevine: Not sure, but I think you just need to scroll down. I see:
```        rejectLevel : `float`, optional
            This is essentially is a sanity check parameter.
            If this condition is violated there is something unexpected
            going on in the image, and it is discarded from the stack before
            the clipped-mean is calculated.
            If not provided, a default value of 0.2 is picked up from the
            task config.```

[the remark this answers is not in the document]
Comment (r-owen) [reply python/lsst/cp/pipe/makeBrighterFatterKernel.py]: I suggest: If not provided then config.xcorrCheckRejectLevel is used. It is risky to document the default value of a config parameter anywhere (except in the config itself, where it is set) because it may change, and of course obs packages may override it.

File: python/lsst/cp/pipe/makeBrighterFatterKernel.py (hunk 34 above)
Comment (r-owen) [reply python/lsst/cp/pipe/makeBrighterFatterKernel.py]: That sounds fine to me.

File: python/lsst/cp/pipe/makeBrighterFatterKernel.py (hunk 39 above)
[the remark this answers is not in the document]
Comment (mfisherlevine) [reply python/lsst/cp/pipe/makeBrighterFatterKernel.py]: Great, thanks, I've just removed the confusing comment in that case.

File: python/lsst/cp/pipe/makeBrighterFatterKernel.py (hunk 30 above)
Comment (mfisherlevine) [reply python/lsst/cp/pipe/makeBrighterFatterKernel.py]: I think this is the last point remaining, but I've been struggling to understand here. What do you mean by "if you used dotted names for subtasks"? I tried to work it out but failed I'm afraid. Do you think you could give me a tiny example? I think regardless of whether it really needs changing I would learn something here.

Regarding the behaviour; mandatory/forbidden now raise, but desirable WARNs if wrong polarity, and logs an INFOs if the key is not found. I like this behaviour because it suggests to the user that they should configure the equivalent keys in the obs_package config, but this could easily be changed to a `raise` if not found if you think that's better (though I think it has to WARN if wrong, otherwise it's not "desirable", it's mandatory.) Does that sound good?

I'm totally happy to change this until you're happy though, so if you can confirm what you'd like to see instead I'll go ahead and make the changes (especially if I can move the hard-coded `isr.assembleCcd.doTrim` into the `mandatory` config).

Comment (r-owen) [reply python/lsst/cp/pipe/makeBrighterFatterKernel.py]: I just mean that the name in the config could be "assembleCcd.doTrim". Parsing that is a bit more work, as `getattr` and `setattr` cannot handle multiple levels in one call. However, a simple loop like the following would suffice (this is only lightly tested):
```
def getDottedAttr(obj, name):
    """Get an attribute that may be several levels deep, e.g. foo.bar.baz
    """
    if not name:
        raise ValueError("Name is blank")
    try:
        nameList = name.split(".")
        value = getattr(obj, nameList[0])
        for subName in nameList[1:]:
            value = getattr(value, subName)
        return value
    except Exception as e:
        raise AttributeError("Could not find {} in {}: {}".format(name, obj, e))
```

r-owen: And just to make it clear: I suggest using this function for all values you are trying to get, not just ones you know are dotted. The loop is not run at all if nameList has only one element, so it'll be quite efficient.

PR #6, opened 2018-10-25
tickets/DM-16295 - Get cp_pipe etc working for ts8 for bootcamp

mfisherlevine:


mfisherlevine:
Yes, good point, on both counts. Let's discuss at some point, but I suspect the answer will be just to not do it 🙂

mfisherlevine:
Yes, good point, on both counts. Let's discuss at some point, but I suspect the answer will be just to not do it 🙂

PR #7, opened 2018-11-05
tickets/DM-15772 - Rename the task, its configClass, and all mentions of its name

mfisherlevine:


python/lsst/cp/pipe/runEotestTask.py
```diff
@@ -179,7 +179,7 @@ def validate(self):
             getattr(self, task).output_dir = self.eotestOutputPath
 
 
-class CpTask(pipeBase.CmdLineTask):
+class RunEotestTask(pipeBase.CmdLineTask):
     """
     Calibration (Products) Production (CP) task.
```
erykoff: Should this topline description be updated too?

mfisherlevine: Yes, yes it should.

mfisherlevine: And, come to think of it, a little more of that docstring too actually...

PR #11, opened 2019-03-28
Merge into lsst:tickets/DM-18683

mfisherlevine:


timj:
@mfisherlevine can you rebase first? There is a bad merge commit in the history.

PR #12, opened 2019-05-10
tickets/DM-19535 Rename MakeBrighterFatterKernelTaskRunner and move to utils

mfisherlevine:
Other parts of cp_pipe need paired visits, so move this
to a utils file and give a generic name, PairedVisitTaskRunner


python/lsst/cp/pipe/utils.py
```diff
@@ -0,0 +1,61 @@
+#
+# LSST Data Management System
+#
+# Copyright 2008-2019  AURA/LSST.
```
TallJimbo: Should switch this to the new copyright form.  I always steal from the `geom` package as some good soul already updated all of those.

PR #14, opened 2019-07-25
tickets/DM-19709 Write findDetects task

mfisherlevine:


tests/test_defects.py
```diff
@@ -0,0 +1,270 @@
+#!/usr/bin/env python
```
parejkoj: Tests don't have a sha-bang. (and below) See RFC-215

tests/test_defects.py
```diff
@@ -0,0 +1,270 @@
...
+import numpy as np
+import copy
+
+import lsst.utils
+import lsst.utils.tests
+
+import lsst.cp.pipe as cpPipe
+import lsst.ip.isr.isrMock as isrMock
+from lsst.afw.geom import Box2I, Point2I, Extent2I
+from lsst.cp.pipe.utils import countMaskedPixels
```
parejkoj: Please order these so the cp pipe stuff goes together. I've started to settle on ordering import blocks alphabetically, but I'm not wedded to the idea. But at least keep similar bits grouped.

tests/test_defects.py
```diff
@@ -0,0 +1,270 @@
...
+from __future__ import absolute_import, division, print_function
+import unittest
+import numpy as np
+import copy
+
+import lsst.utils
+import lsst.utils.tests
+
+import lsst.cp.pipe as cpPipe
+import lsst.ip.isr.isrMock as isrMock
```
parejkoj: This is usually phrased as `from lsst.ip.isr import isrMock`.

Similarly in the other test file.

tests/test_defects.py
```diff
@@ -0,0 +1,270 @@
...
+        mockImageConfig.isTrimmed = True
+
+        self.flatExp = isrMock.FlatMock(config=mockImageConfig).run()
+        (shapeY, shapeX) = self.flatExp.getDimensions()
+
+        # x, y, size tuples
+        # always put edge defects at the start and change the value of nEdge
+        self.brightDefects = [(0, 15, 3), (100, 123, 1), (77, 90, 3)]
+        self.darkDefects = [(25, 0, 1), (33, 62, 2), (52, 21, 2)]
+        nEdge = 1  # NB update if more edge defects are included
```
parejkoj: Maybe it's just me, but I always have to think about what `NB` means in a code comment (is it a person's initials? Wait, something about Latin...). `NOTE:` seems clearer.

tests/test_defects.py
```diff
@@ -0,0 +1,270 @@
...
+        mockImageConfig.flatDrop = 0.99999
+        mockImageConfig.isTrimmed = True
+
+        self.flatExp = isrMock.FlatMock(config=mockImageConfig).run()
+        (shapeY, shapeX) = self.flatExp.getDimensions()
+
+        # x, y, size tuples
+        # always put edge defects at the start and change the value of nEdge
+        self.brightDefects = [(0, 15, 3), (100, 123, 1), (77, 90, 3)]
+        self.darkDefects = [(25, 0, 1), (33, 62, 2), (52, 21, 2)]
```
parejkoj: It looks like these are all square: should non-square defects also be tested? Probably worth noting that the tested defects are all square (I had to look at the bboxes to figure it out).

tests/test_defects.py
```diff
@@ -0,0 +1,270 @@
...
+        self.darkExp = self.flatExp.clone()
+        self.spareImage = self.flatExp.clone()  # for testing edge bits and misc
+
+        self.flatExp.image.array[:] = flatData
+        self.darkExp.image.array[:] = darkData
+
+        self.defaultTask = cpPipe.defects.FindDefectsTask(config=self.defaultConfig)
+
+    def tearDown(self):
+        del self
```
parejkoj: You shouldn't need this method at all.

tests/test_defects.py
```diff
@@ -0,0 +1,270 @@
...
+
+        shouldBeMissed = self.darkBBoxes[self.onlyEdges] + self.brightBBoxes[self.onlyEdges]
+        for defect in defects:
+            self.assertNotIn(defect.getBBox(), shouldBeMissed)
+
+    def test_postProcessDefectSets(self):
+        defects = self.defaultTask.findHotAndColdPixels(self.flatExp, 'flat')
+
+        # minimal test
+        merged = self.defaultTask._postProcessDefectSets([defects], self.flatExp.getDimensions(), 'FRACTION')
```
parejkoj: minimal test how so?

tests/test_defects.py
```diff
@@ -0,0 +1,270 @@
...
+
+        shouldBeFound = self.darkBBoxes[self.noEdges] + self.brightBBoxes[self.noEdges]
+        for defect in defects:
+            self.assertIn(defect.getBBox(), shouldBeFound)
+
+        shouldBeMissed = self.darkBBoxes[self.onlyEdges] + self.brightBBoxes[self.onlyEdges]
+        for defect in defects:
+            self.assertNotIn(defect.getBBox(), shouldBeMissed)
+
+    def test_postProcessDefectSets(self):
```
parejkoj: This method does many things: please add a docstring saying what sorts of tests it does.

It wasn't clear to me how the different cases below actually test different aspects of the code, since all of the `asserts` are the same. What would make one of those asserts fail?

tests/test_defects.py
```diff
@@ -0,0 +1,270 @@
...
+        defects = self.defaultTask.findHotAndColdPixels(self.flatExp, 'flat')
+
+        # minimal test
+        merged = self.defaultTask._postProcessDefectSets([defects], self.flatExp.getDimensions(), 'FRACTION')
+        self.assertEqual(defects, merged)
+
+        # should always be true regardless of config
+        merged = self.defaultTask._postProcessDefectSets([defects, defects], self.flatExp.getDimensions(),
+                                                         'FRACTION')
+        self.assertEqual(defects, merged)
```
parejkoj: How is this different from the above test?

tests/test_defects.py
```diff
@@ -0,0 +1,270 @@
...
+
+        defectList[7] = defectList[7][:-1]
+        merged = task._postProcessDefectSets(defectList, self.flatExp.getDimensions(), 'AND')
+        self.assertNotEqual(defects, merged)
+
+        merged = task._postProcessDefectSets(defectList, self.flatExp.getDimensions(), 'OR')
+        self.assertEqual(defects, merged)
+
+    def test_pixelCounting(self):
```
parejkoj: Please add a docstring: the name of the test doesn't make what it does obvious.

tests/test_defects.py
```diff
@@ -0,0 +1,270 @@
...
+            defectArea += defect.getBBox().getArea()
+
+        crossCheck = 0
+        for x, y, s in self.brightDefects:
+            crossCheck += s**2
+        for x, y, s in self.darkDefects:
+            crossCheck += s**2
+
+        self.assertEqual(defectArea, task._nPixFromDefects(defects))  # the real function being tested
+        self.assertEqual(defectArea, crossCheck)  # but calculate the sum two ways
```
parejkoj: I don't understand what advantage these two asserts give? If `getBBox().getArea()` is wrong, we have bigger problems...

tests/test_defects.py
```diff
@@ -0,0 +1,270 @@
...
+
+    def test_getNumGoodPixels(self):
```
parejkoj: Please add a docstring for this one too.

tests/test_defects.py
```diff
@@ -0,0 +1,270 @@
...
+
+        self.assertEqual(imageSize, nGood)
+
+        NODATABIT = mi.mask.getPlaneBitMask("NO_DATA")
+
+        noDataBox = Box2I(Point2I(31, 49), Extent2I(3, 6))
+        testImage.mask[noDataBox] |= NODATABIT
+
+        self.assertEqual(imageSize - noDataBox.getArea(), self.defaultTask._getNumGoodPixels(mi))
+        self.assertEqual(imageSize, self.defaultTask._getNumGoodPixels(mi, 'BAD'))  # check for misfire
```
parejkoj: "misfire" of what?

tests/test_defects.py
```diff
@@ -0,0 +1,270 @@
...
+        hEdge = self.defaultConfig.nPixBorderLeftRight
+        vEdge = self.defaultConfig.nPixBorderUpDown
+        xSize, ySize = mi.getDimensions()
+
+        nEdge = xSize*vEdge*2 + ySize*hEdge*2 - hEdge*vEdge*4
+
+        self.assertEqual(countMaskedPixels(mi, 'EDGE'), nEdge)
+
+    def testImport(self):
+        import lsst.cp.pipe as cpPipe  # noqa: F401
```
parejkoj: This is entirely superfluous.

tests/test_defects.py
```diff
@@ -0,0 +1,270 @@
...
+
+    def testClassInstantiation(self):
```
parejkoj: Also superfluous: you instantiate these in `setUp`, and tests that something doesn't blow up on instantiation aren't generally useful without also testing some functionality.

tests/test_defects.py
```diff
@@ -0,0 +1,270 @@
...
+        testImage.mask[noDataBox] ^= NODATABIT  # XOR to reset what we did
+        self.assertEqual(imageSize, nGood)
+
+        BADBIT = mi.mask.getPlaneBitMask("BAD")
+        badBox = Box2I(Point2I(85, 98), Extent2I(4, 7))
+        testImage.mask[badBox] |= BADBIT
+
+        self.assertEqual(imageSize - badBox.getArea(), self.defaultTask._getNumGoodPixels(mi, 'BAD'))
+
+    def test_edgeMasking(self):
```
parejkoj: Please document this test as well. You check that no pixels are masked, mask some pixels, then check that they got masked?

tests/test_utils.py
```diff
@@ -0,0 +1,151 @@
...
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the LSST License Statement and
+# the GNU General Public License along with this program.  If not,
+# see <https://www.lsstcorp.org/LegalNotices/>.
+#
+"""Test cases for cp_pipe."""
```
parejkoj: These a really tests of lsst.cp.pipe.utils, no?

tests/test_defects.py
```diff
@@ -0,0 +1,270 @@
...
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the LSST License Statement and
+# the GNU General Public License along with this program.  If not,
+# see <https://www.lsstcorp.org/LegalNotices/>.
+#
+"""Test cases for cp_pipe."""
```
parejkoj: These are really tests of `FindDefectsTask`, yes?

tests/test_utils.py
```diff
@@ -0,0 +1,151 @@
...
+
+        with self.assertRaises(RuntimeError):
+            cpUtils.checkExpLengthEqual(exp1, exp2, raiseWithMessage=True)
+
+    def test_validateIsrConfig(self):
+
+        mand = ['doAssembleCcd']
+        forb = ['doFlat', 'doFringe']
+        des = ['doBias', 'doDark']
+        undes = ['doLinearize', 'doBrighterFatter']
```
parejkoj: `mand forb des`. The new boardgame from Fantasy Flight!

Please give them non-abbreviated names.

python/lsst/cp/pipe/utils.py
```diff
@@ -20,12 +20,27 @@
...
+
+import re
+import numpy as np
 
 import lsst.pipe.base as pipeBase
+import lsst.ip.isr as ipIsr
+import lsst.log as lsstLog
+
 
+def countMaskedPixels(maskedIm, maskPlane):
```
parejkoj: No docstring. Also, we don't already have this functionality in `afw.image`?

python/lsst/cp/pipe/defects.py
```diff
@@ -0,0 +1,691 @@
...
+        for amp in exp.getDetector():
+            ampImg = maskedIm[amp.getBBox()].clone()
+
+            # crop ampImage depending on where the amp lies in the image
+            if self.config.nPixBorderLeftRight:
+                if ampImg.getX0() == 0:
+                    ampImg = ampImg[self.config.nPixBorderLeftRight:, :, afwImage.LOCAL]
+                else:
+                    ampImg = ampImg[:-self.config.nPixBorderLeftRight, :, afwImage.LOCAL]
+            if self.config.nPixBorderUpDown:
```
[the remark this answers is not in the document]
RobertLuptonTheGood: There's almost certainly a method on amp to get the BBox (and you should use it).  Which pixels are you looking for here?

python/lsst/cp/pipe/defects.py
```diff
@@ -0,0 +1,691 @@
...
+    _DefaultName = "findDefects"
+
+    def __init__(self, *args, **kwargs):
+        pipeBase.CmdLineTask.__init__(self, *args, **kwargs)
+        self.makeSubtask("isrForFlats")
+        self.makeSubtask("isrForDarks")
+        # change these back once you use the INFO level logging to check
+        # that all the right things are being done, and only the right things!
+        # self.isrForFlats.log.setLevel(lsstLog.WARN)
+        # self.isrForDarks.log.setLevel(lsstLog.WARN)
```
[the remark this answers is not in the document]
mfisherlevine: Sorry, yes, that all just needed deleting. All done 🙂

python/lsst/cp/pipe/defects.py
```diff
@@ -0,0 +1,691 @@
...
+        msg = "Combining %s defect sets from darks for detector %s"
+        self.log.info(msg, len(defectLists['dark']), detNum)
+        mergedDefectsFromDarks = self._postProcessDefectSets(defectLists['dark'], exp.getDimensions(),
+                                                             self.config.combinationMode)
+        msg = "Combining %s defect sets from flats for detector %s"
+        self.log.info(msg, len(defectLists['flat']), detNum)
+        mergedDefectsFromFlats = self._postProcessDefectSets(defectLists['flat'], exp.getDimensions(),
+                                                             self.config.combinationMode)
+
+        self.log.info("Combining bright and dark defect sets")
```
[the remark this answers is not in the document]
mfisherlevine: I think it's in keeping with all the other messages in the task. I will add the identifiers and keep it as an `INFO` level so that it's coherent. This task is not very verbose, nor is it long or slow, so it's nice to see what's happening.

python/lsst/cp/pipe/defects.py
```diff
@@ -0,0 +1,691 @@
...
+        brightDarkPostMerge = [mergedDefectsFromDarks, mergedDefectsFromFlats]
+        allDefects = self._postProcessDefectSets(brightDarkPostMerge, exp.getDimensions(), mode='OR')
+
+        self._writeData(dataRef, allDefects)
+
+        self.log.info("Finished finding defects in detector %s" % detNum)
```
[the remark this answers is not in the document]
mfisherlevine: Nobody ever runs at `DEBUG` level as far as I can see, and it's nice to see when it finishes. Yes, it _should_ raise before then, but I don't see any harm in getting some nice user-friendly confirmation that everything went as expected.

python/lsst/cp/pipe/defects.py
```diff
@@ -0,0 +1,691 @@
...
+        defects : `lsst.meas.algorithms.Defect`
+            The defects to be written.
+        """
+        filename = dataRef.getUri(write=True)  # does not guarantee that full path exists
+        dirname = os.path.dirname(filename)
+        if not os.path.exists(dirname):
+            os.makedirs(dirname)
+
+        msg = "Writing defects to %s in format: %s"
+        self.log.info(msg, os.path.splitext(filename)[0], self.config.writeAs)
```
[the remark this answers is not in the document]
mfisherlevine: The extension changes according to what the type written is. I could put log messages in each of the write blocks, but that seems unnecessarily messy. Would you strongly prefer that?

python/lsst/cp/pipe/makeBrighterFatterKernel.py
```diff
@@ -760,7 +760,7 @@ def estimateGains(self, dataRef, visitPairs):
         return gains, nomGains
 
     @staticmethod
-    def _checkExpLengthEqual(exp1, exp2, v1=None, v2=None):
+    def _checkExpLengthEqual(exp1, exp2, v1=None, v2=None):  # TODO: refactor this to use the utils version
```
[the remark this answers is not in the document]
mfisherlevine: Was quicker to just do the work instead 🙂

python/lsst/cp/pipe/utils.py
```diff
@@ -20,12 +20,27 @@
...
+    maskBit = maskedIm.mask.getPlaneBitMask(maskPlane)
+    # bit = afwImage.Mask.getPlaneBitMask(maskPlane)
+    # nPix = len(np.where(np.bitwise_and(maskedIm.mask.array, maskBit))[0])
```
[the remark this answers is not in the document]
mfisherlevine: But what about the other way of doing it? 😭 Which is better? How will I remember what past-Merlin was contemplating?! 😄 OK, done.

python/lsst/cp/pipe/utils.py
```diff
@@ -20,12 +20,27 @@
...
+__all__ = ['PairedVisitListTaskRunner', 'SingleVisitListTaskRunner',
+           'NonexistentDatasetTaskDataIdContainer', 'parseCmdlineNumberString',
+           'countMaskedPixels', 'checkExpLengthEqual']
+
+import re
+import numpy as np
 
 import lsst.pipe.base as pipeBase
+import lsst.ip.isr as ipIsr
+import lsst.log as lsstLog
```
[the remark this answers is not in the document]
mfisherlevine: I thought this was the standard import for this, like `import lsst.afw.image as afwImage` - it's in several places. Should I actually change this?

python/lsst/cp/pipe/utils.py
```diff
@@ -58,3 +73,217 @@ def getTargetList(parsedCmd, **kwargs):
             visitPairs.append(visits)
 
         return pipeBase.TaskRunner.getTargetList(parsedCmd, visitPairs=visitPairs, **kwargs)
+
+
+def parseCmdlineNumberString(inputString):
+    """Parse command line numerical expression sytax and return as list of int
```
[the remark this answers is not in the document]
mfisherlevine: I couldn't find it. This is basically a copy & paste from there, so yes, it really _should_ exist in a way that's importable for others, but no, I had to pull it out and hack it a little.

python/lsst/cp/pipe/utils.py
```diff
@@ -58,3 +73,217 @@ def getTargetList(parsedCmd, **kwargs):
...
+
+
+class SingleVisitListTaskRunner(pipeBase.TaskRunner):
+    """Subclass of TaskRunner for the MakeBrighterFatterKernelTask.
+
+    This transforms the processed arguments generated by the ArgumentParser
+    into the arguments expected by makeBrighterFatterKernelTask.run().
+
+    makeBrighterFatterKernelTask.run() takes a two arguments,
+    one of which is the dataRef (as usual), and the other is the list
```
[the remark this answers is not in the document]
mfisherlevine: Agreed. Copy & paste hangover. At least it was there because I was being a good-boy and refactoring nicely for once 🙂

python/lsst/cp/pipe/defects.py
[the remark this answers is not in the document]
timj: Also, you don't even know what extension is going to be written if you are writing to text file.

python/lsst/cp/pipe/utils.py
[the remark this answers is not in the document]
timj: The standard is definitely not to include "lsst" in the imported package name. Here `as log` is fine. I think @parejkoj tends to prefer using `lsst.log` everywhere to be explicit (and it's only one character longer than you have above).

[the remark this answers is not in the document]
mfisherlevine: OK, I've just changed it to `import lsst.log` and left it as `lsst.log` in the code. There are many other places in the stack this transgression is occurring though...

python/lsst/cp/pipe/defects.py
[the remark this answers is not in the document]
mfisherlevine: In which case I'm just marking as resolved.

python/lsst/cp/pipe/utils.py
```diff
@@ -58,3 +73,217 @@ def getTargetList(parsedCmd, **kwargs):
...
+        """Parse the visit list and pass through explicitly."""
+        assert len(parsedCmd.visitList) == 1  # even if you give two on the cmdline you only get 1 here
+        visits = parseCmdlineNumberString(parsedCmd.visitList[0])
+
+        return pipeBase.TaskRunner.getTargetList(parsedCmd, visitList=visits, **kwargs)
+
+
+class NonexistentDatasetTaskDataIdContainer(pipeBase.DataIdContainer):
+    """A DataIdContainer for the tasks for which the output does
+    not yet exist."""
```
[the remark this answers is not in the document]
mfisherlevine: Me too...

python/lsst/cp/pipe/utils.py
```diff
@@ -58,3 +73,217 @@ def getTargetList(parsedCmd, **kwargs):
...
+            if configParam not in configDict:
+                raise RuntimeError(f"Mandatory parameter {configParam} not found in the isr configuration.")
+            if configDict[configParam] is False:
+                raise RuntimeError(f"Must set config.isr.{configParam} to True for this task.")
+
+    if forbidden:
+        for configParam in forbidden:
+            if configParam not in configDict:
+                log.warn(f"Failed to find key {configParam} in the isr config. This is set to be " +
+                         "forbidden, so you should ensure that the equivalent for your obs_package to False.")
```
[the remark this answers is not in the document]
mfisherlevine: If it's there, and you didn't set it to False, then fail.
If it's not there, then it can't be true, so that's kind of OK, but something might be fishy, so warn the user.
It's weird, but I think it's clear - I thought about this part reasonably carefully.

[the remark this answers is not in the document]
mfisherlevine: There are reasons why some are `WARN`, some are `INFO` and some are `raise`. I think it's as it should be.

python/lsst/cp/pipe/utils.py
```diff
@@ -58,3 +73,217 @@ def getTargetList(parsedCmd, **kwargs):
...
+                         " It is probably desirable to have this set to True")
+    if undesirable:
+        for configParam in undesirable:
+            if configParam not in configDict:
+                log.info(f"Failed to find key {configParam} in the isr config. You probably want" +
+                         " to set the equivalent for your obs_package to False.")
+                continue
+            if configDict[configParam] is True:
+                log.warn(f"Found config.isr.{configParam} set to True for this task." +
+                         " It is probably desirable to have this set to False")
```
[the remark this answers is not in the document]
mfisherlevine: I went with undesirable because it's a word, and the opposite of desirable, whereas (un)recommended isn't. It is indeed supposed to be a finger wagging, hence only being at info/warn level.

tests/test_defects.py
mfisherlevine: Yeah, it's not optimal, but I think it's fine for now. Have noted the implicit squareness.

tests/test_defects.py
mfisherlevine: Dunno - weird comment, has been removed.

tests/test_defects.py
mfisherlevine: Have documented this now. There is potential for logic errors when doing this combination step, so test it carefully for the different modes that are possible.

tests/test_defects.py
mfisherlevine: One is "merging" a set with nothing (the list has length one), the second is merging it with itself. This was failing at one point, so I know it's not identical!

mfisherlevine: I've now added a comment to point that out though 👍

tests/test_defects.py
mfisherlevine: Summing the areas is not the same as looking at the number of defective pixels in an image, for several reasons: overlapping bboxes, pixels not making the cut when defect sets are merged, maybe others. Definitely doesn't hurt to assert twice - if it's always true then no harm, and if it's not then it should be looked at.

tests/test_defects.py
mfisherlevine: Have clarified in the comment, which now reads
`# check for misfire; we're setting NO_DATA here, not BAD`

tests/test_defects.py
mfisherlevine: Yeah, this was written before the substantive tests. Removed.

tests/test_utils.py
mfisherlevine: Yes. Changed.

tests/test_defects.py
mfisherlevine: Yes, changed.

tests/test_utils.py
mfisherlevine: This was necessary for the line lengths. I originally had it like that, but it was _such_ a mess. I don't think this is worth it. I've added comments to the lines instantiating these to make it clear what they are though, I think that should suffice.

python/lsst/cp/pipe/utils.py
mfisherlevine: Not that I know of.

python/lsst/cp/pipe/defects.py
```diff
@@ -0,0 +1,691 @@
...
+              The exit code
+
+        """
+
+        detNum = dataRef.dataId[self.config.ccdKey]
+        self.log.info("Calculating defects using %s visits for detector %s" % (visitList, detNum))
+
+        defectLists = {'dark': [], 'flat': []}
+
+        if self.config.mode == 'MASTER':
```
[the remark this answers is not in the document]
mfisherlevine: I am leaving this as-is because @RobertLuptonTheGood has a ticket pending (off this branch - DM-20940) that will address this, so a) it would be duplicating effort that's already been done, and b) will leave him in a rebase nightmare. Consider there an invisible `TODO: DM-20940` on this that's already in progress 🙂

python/lsst/cp/pipe/defects.py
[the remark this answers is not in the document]
mfisherlevine: Many points here:
* This isn't working on the amps, this is working on the whole detector
* There are indeed methods to get the bbox, and yes, you could then dilate/contract it, and you could make copies, and do gross stuff ORing them or whatever, but it's going to be horrible
* Even then I don't see how you'd easily get a rectangle on each edge easily
* And if you somehow did, it would be really muddy to them break the symmetry to support different sized borders at the top/bottom vs the left/right

So yes, there probably _is_ a way to do it with bboxes, but I think it's going to be much worse than what is currently here.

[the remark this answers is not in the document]
mfisherlevine: Basically, applying asymmetric amp-wise borders is not something I know how to do elegantly, and doesn't come up that often I don't think. I do agree that it should be supported by lower level code, but moving that stuff there is well beyond the scope of this ticket.

python/lsst/cp/pipe/utils.py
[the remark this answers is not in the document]
parejkoj: `as log` means you can't use `log` anywhere in the code. Collapsing that by removing the period makes no sense to me at all, and I wonder where it came from.

python/lsst/cp/pipe/utils.py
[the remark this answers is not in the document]
parejkoj: Can you please ask on #dm-middleware about whether such cmdline parsing will be more accessible in gen3?

python/lsst/cp/pipe/utils.py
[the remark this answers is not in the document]
parejkoj: I think the "it's in this list, but not in the Config" messages should make it more clear that that is where the discrepancy is (and there's also not an equivalent warning for the `required` list. Something like:

"Failed to find forbidden key {configParam} in the isr config. The keys in the forbidden list should each have an associated Field in IsrConfig: check that there is not a typo in this case."

python/lsst/cp/pipe/utils.py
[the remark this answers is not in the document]
parejkoj: Maybe change "It is probably desirable..." to "The cp_pipe Config recommends setting this to False."?

tests/test_defects.py
parejkoj: I think the "the real function being tested" comment threw me off some. Maybe remove those two inline comments and replace them with a comment over the whole thing:

"Test the result of `_nPixFromDefects` via two different ways of calculating area."

tests/test_utils.py
parejkoj: "This was necessary for the line lengths." That's not good. Looking at the lines below, changing e.g. `cpUtils.validateIsrConfig(task, mand + ['test'], forb, des, undes)` to:

```
cpUtils.validateIsrConfig(task,
                          mandatory + ['test'],
                          forbidden,
                          desirable,
                          undesirable)
```

seems like a win on several fronts (full words, breaking up long parameter list).

python/lsst/cp/pipe/defects.py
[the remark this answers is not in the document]
parejkoj: Ok. Each of those two new functions feels like a perfect case for a unittest of exactly that function (instead of the integration tests you have below).

python/lsst/cp/pipe/utils.py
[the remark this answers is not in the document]
mfisherlevine: Dunno ¯\_(ツ)_/¯ But I dislike `log` on its own for exactly that reason too 🙂

python/lsst/cp/pipe/utils.py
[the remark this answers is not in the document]
mfisherlevine: Done. 👍

python/lsst/cp/pipe/utils.py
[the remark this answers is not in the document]
mfisherlevine: Done.

python/lsst/cp/pipe/utils.py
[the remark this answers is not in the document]
mfisherlevine: Done.

tests/test_defects.py
mfisherlevine: Done.

tests/test_utils.py
mfisherlevine: Whatever floats your goat. 🐐

python/lsst/cp/pipe/defects.py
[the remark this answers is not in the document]
mfisherlevine: I trust Robert will do exactly that 🙂

PR #17, opened 2019-10-01
DM-19903: Add columnar defect identification to findDefectsTask in cp_pipe please

plazas:


python/lsst/cp/pipe/defects.py
```diff
@@ -529,6 +539,48 @@ def findHotAndColdPixels(self, exp, imageType, setMask=False):
...
+
+                # Next few lines: don't mask pixels in column if gap between
+                # two consecutive bad pixels is too large.
+                diffIndex = np.where(np.diff(multipleY) > gapThreshold)[0]
+                if len(diffIndex) != 0:
+                    limits = [minY]  # put the minimum first
+                    for gapIndex in diffIndex:
+                        limits.append(multipleY[gapIndex])
+                        limits.append(multipleY[gapIndex+1])
+                    limits.append(maxY)  # maximum last
```
parejkoj: Could you do the above loop with a `sort(multipleY[diffIndex])` instead? I'm having trouble visualizing it.

It also seems like you could create the defects themselves as part of this loop, and get rid of the `for i in np.arange` below.

parejkoj:
Also, please update the PR title to match the standard:

https://developer.lsst.io/work/flow.html#make-a-pull-request

python/lsst/cp/pipe/defects.py
```diff
@@ -129,6 +130,16 @@ class FindDefectsTaskConfig(pexConfig.Config):
         doc="Number of pixels to exclude from left & right of image when looking for defects.",
         default=7,
     )
+    colThreshold = pexConfig.Field(
+        dtype=int,
+        doc="Minimum number of bad pixels in a column.",
```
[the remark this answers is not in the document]
plazas: What you wrote in the third paragraph is the explanation of what this parameter does, thanks. I have copied that explanation into the "doc" field of "colThreshold".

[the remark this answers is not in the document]
plazas: I have adopted your naming suggestion, and changed the "doc" string to the following: 
"If BPC is the set of all the bad pixels in a given column (not necessarily consecutive), and the size of BPC is at least 'badPixelColumnThreshold', all the pixels between the two pixels that satisfy minY (BPC) and maxY (BPC) will be marked as bad, with 'Y' being the long axis of the amplifier (and 'X' the other axis, which for a column is a constant for all pixels in the set BPC). If there are more than 'goodPixelColumnGapThreshold' consecutive non-bad pixels in BPC, an exception to the above is made and those consecutive 'goodPixelColumnGapThreshold' are not marked as bad.",

python/lsst/cp/pipe/defects.py
```diff
@@ -129,6 +130,16 @@ class FindDefectsTaskConfig(pexConfig.Config):
...
+        default=50,
+    )
+    gapThreshold = pexConfig.Field(
+        dtype=int,
+        doc="Size, in pixels, of usable pixels in a column with on and off bad pixels.",
```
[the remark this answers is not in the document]
plazas: I have adopted your naming suggestion, and included a reference to this parameter in the "doc" tring of "badPixelColumnThreshold" above.   In the example that you give, there would be 98 bad pixels in the column in total (not necessarily consecutive), so all the pixels in between those two with the minimum and maximum "Y" coordinate amongst the 98 would be marked as bad. There would be an exception for the 31 pixel that you mention, assuming that they are consecutive.

python/lsst/cp/pipe/defects.py
```diff
@@ -305,7 +316,6 @@ def runDataRef(self, dataRef, visitList):
                 imageType = butler.queryMetadata('raw', self.config.imageTypeKey, dataId={'visit': visit})[0]
                 imageType = imageType.lower()
                 dataRef.dataId['visit'] = visit
-
```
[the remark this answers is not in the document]
plazas: I don't know. @mfisherlevine , do you know?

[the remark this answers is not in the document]
mfisherlevine: I'm on vacation today and will be offline for the rest of the day, but I'm not sure why you're asking _me_ why _you_ removed a single blank line...

python/lsst/cp/pipe/defects.py
```diff
@@ -539,48 +546,68 @@ def findHotAndColdPixels(self, exp, imageType, setMask=False):
...
         for defect in defects:
             bbox = defect.getBBox()
+            if not (bbox.getMinX() == bbox.getMaxX()): # assume defect region is only one pixel: single column
+                continue
             x.append(bbox.getMinX())
             y.append(bbox.getMinY())
         x = np.array(x)
         y = np.array(y)
-
+        stop
```
mfisherlevine: Errrr 🤨

python/lsst/cp/pipe/defects.py
```diff
@@ -539,48 +546,68 @@ def findHotAndColdPixels(self, exp, imageType, setMask=False):
...
+        goodPixelColumnGapThreshold = self.config.goodPixelColumnGapThreshold  # (not "bad" pixels, see below).
+        for x0 in multipleX:
+            index = np.where(x == x0)
+            multipleY = y[index]  # multipleY and multipleX are in 1-1 correspondence.
+            minY, maxY = np.min(multipleY), np.max(multipleY)
+
+            # Next few lines: don't mask pixels in column if gap of good pixels between
+            # two consecutive bad pixels is larger or equal than 'goodPixelColumnGapThreshold'.
+            diffIndex = np.where(np.diff(multipleY) >= goodPixelColumnGapThreshold)[0]
+            print ("diffIndex: ", diffIndex)
```
mfisherlevine: We don't use `print()` functions in production code - replace with logger messages.

python/lsst/cp/pipe/defects.py
```diff
@@ -539,48 +546,68 @@ def findHotAndColdPixels(self, exp, imageType, setMask=False):
...
+            if len(diffIndex) != 0:
+                limits = [minY]  # put the minimum first
+                for gapIndex in diffIndex:
+                    limits.append(multipleY[gapIndex])
+                    limits.append(multipleY[gapIndex+1])
+                limits.append(maxY)  # maximum last
+                print ("limits: ", limits)
```
mfisherlevine: As above.

timj: And this is a PEP8 violation as well (no space before the `(`). You have many flake8 errors that will need fixing.

python/lsst/cp/pipe/defects.py
```diff
@@ -529,6 +539,48 @@ def findHotAndColdPixels(self, exp, imageType, setMask=False):
...
         defects = measAlg.Defects.fromFootprintList(footprintList)
+
+        # If there's a column with on and off defects, mask all the pixels in between.
+        # Get the x, y values of the defects.
+        x, y = [], []
+        for defect in defects:
+            bbox = defect.getBBox()
+            x.append(bbox.getMinX())
+            y.append(bbox.getMinY())
+        x = np.array(x)
```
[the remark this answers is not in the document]
mfisherlevine: I imagine that this is the kind of thing where the amount of developer time is >> the wall-time that will ever be saved when implementing, likely by >3 orders of magnitude, if not 6. You're not wrong, but I just don't think it will actually matter in reality. If I'm missing something about why this might be important though, please do let me know, I might just not have thought about it correctly.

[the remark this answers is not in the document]
timj: The code is also not what you want. Getting a list of lower left coordinates of all the bounding boxes does not tell you anything about the pixels that are masked. What you need is to get the mask from the bounding box (and `Defects` has a method for that).

mfisherlevine:
> To be clear, I haven't actually reviewed anything, I just glanced at that one push out of curiosity and saw the first `stop` line, and was curious enough to read the rest of the commit.

Oh yeah, and the other 24 flake8 violations...

python/lsst/cp/pipe/defects.py
plazas: Thank you. I am not done yet addressing the comments and I have not yet checked for flake8 errors; I plan to do that before formally coming back to the reviewer.

mfisherlevine: I mean, OK, but don't you run a linter in your editor? Wouldn't that be a good idea?

mfisherlevine: ~Wouldn't that be a good idea?~ That _would be a good idea_.

python/lsst/cp/pipe/defects.py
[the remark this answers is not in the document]
plazas: I'm confused. Even in the case that the defect has only one pixel width it is true that the bounding boxes do not tell us anything about the pixels that are masked? The code was written to handle bad pixels in columns.  What do they bounding boxes us about? In any case, I will look for the mask of the bounding box that you talk about and see if I can use that information in this case. Thanks.

python/lsst/cp/pipe/defects.py
plazas: Yes, I run it when I'm done.

mfisherlevine: Do. Not. Do. That. Want to know why? Wait for my next response on here...

mfisherlevine:
There are so many things wrong with findBadColumns() and the subsequent code I don't know where to start.

Firstly, just code-wise:

At the end of `findBadColumns()`, you call `_markOneBadColumn()` with `(x, y, x0, multipleX, multipleY, defects)` where neither `x0` nor `multipleY` are even defined, implying that code has never been run.

Then _markOneBadColumn() doesn't have `self` as the first variable, nor does this have a `@staticmethod` decorator, showing that the next function has also never been run, not to mention that in that function you're appending to a list called `defects` which is also not defined.

Obviously, a *linter* would make sure that *all your variables are defined*, so mistakes like this wouldn't happen. Then again, perhaps this is the first time in history where I'm actually glad you're _not_ running one, because it reveals the clear evidence of pushing code that has never been _run_, let alone tested.

Scientifically:

Why do you `continue` when `bbox.getMinX() != bbox.getMaxX()`? This implies that you're both not interested in any double-or-more columns, nor potentially even bad columns where the adjoining column has a bad pixel in it. This doesn't seem right - if a whole column is bad, it's _more_ likely to have somewhat bad columns right next to it, so this seems like it will _exclude_ the _very_ situations we're wanting to look at.

Again - this doesn't constitute a full review. Clearly, this code has not been run as-is, so I think these problems need to be fixed before more time is spent reading the code. (When functions have no docstrings and are accessing variables that don't even exist, it's super hard to work out what it is trying to do).

plazas:
Merlin, I'm still working on addressing the comments by the reviewer. This code has not been run, of course, and it's not ready to be reviewed yet.

parejkoj:
Beyond Merlin's comments, I will re-iterate that you really need to write some tests that define what you intend the code to do, so that when you write the code you can confirm that it really does that. There is already a `FindDefectsTaskTestCase` that you can build on. Writing those tests should be your first priority, since the tests would help you be sure that any other changes you make keep the functionality unchanged from what is intended.

plazas:
Yes, thanks. I'm actually working right now on the tests, as you initially suggested. I agree with you.

plazas:
@parejkoj: I have put the new code in functions ("maskBlocksIfIntermitentBadPixelsInColumn" and "_markBlocksInBadColumn"), with more explanations in the doc strings and examples. I have added a test function called "test_maskBlocksIfIntermitentBadPixelsInColumn".  I have run the code and tests, and the tests pass.  As ""maskBlocksIfIntermitentBadPixelsInColumn"  just calls     "_markBlocksInBadColumn", a separate test for the latter would be redundant.  

To clarify, the new code added in this ticket is not meant to find bad columns. That happens already when running the code before. The purpose of this extra code, as described in the associated Jira ticket, is "given that sometimes columns have on-and-off bad pixels, it would be good to mask whole blocks of pixels in the column as bad, unless there is a certain amount of consecutive good pixels  between two bad pixels in the column". 

Thank you for your comments and please let mew know what you think.

python/lsst/cp/pipe/defects.py
plazas: In the end I left the code as it was, because I thought it was easy to understand as it is. I have added clarifications in the comments and doc-string; please let me know if further clarification is needed.

python/lsst/cp/pipe/defects.py
```diff
@@ -529,6 +546,141 @@ def findHotAndColdPixels(self, exp, imageType, setMask=False):
...
+            bbox = defect.getBBox()
+            x.append(bbox.getMinX())
+            y.append(bbox.getMinY())
+            dimensions = bbox.getDimensions()
+            extensionX.append(dimensions[0])
+        x = np.array(x)
+        y = np.array(y)
+        extensionX = np.array(extensionX)
+        # Find the defects with same "x" (vertical) coordinate (column).
+        unique, counts = np.unique(x, return_counts=True)
```
timj: Maybe I'm misreading the code but I'm still a bit worried that the code cares about the edge of the defect region and that if some defects were bigger than single pixels and started a few pixels over that they would be missed. If the bad pixels were converted into that mask and that mask was collapsed into a 1-D histogram, would that be enough to tell you if a majority of pixels in the column were masked out so you could mask the whole column? You could then fix up the mask itself and then generate a new set of defects from the mask.

python/lsst/cp/pipe/defects.py
```diff
@@ -673,7 +673,7 @@ def _markBlocksInBadColumn(self, x, y, extensionX, multipleX, defects):
                     limits.append(multipleY[gapIndex])
                     limits.append(multipleY[gapIndex+1])
                 limits.append(maxY)  # maximum last
-                assert (len(limits)%2 == 0)  # `limits` it's even by design, but check anyways
+                assert (len(limits)%2 == 0, 'limits is even by design, but check anyways')
```
timj: This code is not correct. I think you get a warning from Python about it always being true and flake8 also complains. You need to remove the parentheses.

python/lsst/cp/pipe/defects.py
plazas: I added a few lines that loop over each defect bounding box to obtain the coordinates (x,y) of each individual bad pixel in the amp.

timj: This still all looks really complicated. Did you discount the approach of generating the mask and then scanning quickly through each column looking for gaps?  Do you know you handle overlapping bounding boxes properly? The tests don't look to me like they deal with cases other than your hand crafted situation.

plazas: Thank you for your comment. I adopted this approach because it seemed simpler to me than generating the mask. I am assuming that going through all the boxes and then going through each box to get the (x,y) coordinates of all the bad pixels would make sure that no bad pixel was missed, regardless whether their original bounding boxes overlapped or not. Perhaps I am not understanding your concern about overlapping bounding boxes. Would you mind elaborating on that, please?  Regarding the test, I designed it in order to address the charge of the ticket (blocking on-and-off bad pixels in a column) and the comments above related to possible multiple columns (which was a concern when I was using the origin coordinate of the boxes---x_min and y_min---but that is addressed when using single pixel coordinates). Do you have another case (or other cases) that perhaps I could test?

tests/test_defects.py
```diff
@@ -67,16 +67,42 @@ def setUp(self):
...
+                              [55, 28, 1, 1], [55, 30, 1, 1],
+                              (60, 1, 1, 18), (60, 20, 1, 10), (61, 2, 2, 2), (61, 6, 2, 8),
+                              (70, 1, 1, 18), (70, 20, 1, 10), (68, 2, 2, 2), (68, 6, 2, 8),
+                              (75, 1, 1, 18), (75, 20, 1, 10), (73, 2, 2, 2), (73, 6, 2, 8), (76, 2, 2, 2),
+                              (76, 6, 2, 8),
+                              (80, 1, 1, 18), (80, 20, 1, 10), (81, 2, 2, 2), (81, 8, 2, 8),
+                              (87, 1, 1, 18), (87, 20, 1, 10), (85, 2, 2, 2), (85, 8, 2, 8),
+                              (93, 1, 1, 12), (93, 15, 1, 20), (91, 2, 2, 2), (91, 7, 2, 2),
+                              (94, 2, 2, 2), (94, 7, 2, 2),
+                              (91, 18, 2, 3), (91, 24, 2, 3), (94, 18, 2, 3), (94, 24, 2, 3)]
```
parejkoj: I think it would be clearer if each specific column test had exactly its the list of defects that it was testing defined in the test. So, The `(15,1,1,50)` column shouldn't be here, just in that one test. It would make it easier to see what's going on. And then you could note that the `50` length is because of splitting along the mock amp boundary.

tests/test_defects.py
```diff
@@ -103,30 +131,442 @@ def setUp(self):
...
+        defects = self.allDefectsList
+
+        defectsWithColumns = task.maskBlocksIfIntermitentBadPixelsInColumn(defects)
+
+        boxesMeasured = []
+        for defect in defectsWithColumns:
+            boxesMeasured.append(defect.getBBox())
+
+        for boxInput in expectedDefects:
+            self.assertIn(boxInput, boxesMeasured)
```
parejkoj: You could clean this up quite a bit by having a `check_maskBlocks` method that takes the input bad data and the expected defects, and then creates the image, calls `maskBlocks` and does the test. Then each individual test would look like this:

```
test_maskBlocks_long_column(self):
    badPixels = [(1,15,1,50), ...]
    expected = [Box2I(...)]
    check_maskBlocks(badPixels, expected)
```

plazas:
@parejkoj Thank you for your comments. I have addressed them, and updated the code.

tests/test_defects.py
```diff
@@ -103,30 +108,472 @@ def setUp(self):
...
+
+        Input defects for this test: (60, 1, 1, 18), (60, 20, 1, 10), (61, 2, 2, 2), (61, 6, 2, 8)
+
+        Expected defects after the function runs: (60,1,1,31), (61, 2, 2, 14)
+
+        Plots can be found in DM-19903 on Jira.
+        """
+
+        expectedDefects = [Box2I(corner = Point2I(60, 1), dimensions = Extent2I(1, 29)),
+                           Box2I(corner = Point2I(61, 2), dimensions = Extent2I(1, 12))]
```
parejkoj: I don't think this is right: the extent of that region is 2 wide in 'x', so the output bad block should have `Extent(2,12)`, I believe. If this is what the code was producing, there may be a bug.

tests/test_defects.py
```diff
@@ -103,30 +108,472 @@ def setUp(self):
...
+
+        Input defects for this test: (70, 1, 1, 18), (70, 20, 1, 10), (68, 2, 2, 2), (68, 6, 2, 8)
+
+        Expected defects after the function runs: (70,1,1,31), (68, 2, 2, 14)
+
+        Plots can be found in DM-19903 on Jira.
+        """
+
+        expectedDefects = [Box2I(corner = Point2I(70, 1), dimensions = Extent2I(1, 29)),
+                           Box2I(corner = Point2I(68, 2), dimensions = Extent2I(1, 12))]
```
parejkoj: As above, I think this is incorrect and the Extent of the second item should be `2,12`

tests/test_defects.py
```diff
@@ -86,14 +89,16 @@ def setUp(self):
 
         # NOTE: darks and flats have same defects applied deliberately to both
         for defect in self.brightDefects:
-            y, x, s = defect
-            flatData[x:x+s, y:y+s] += self.nSigmaBright * flatWidth  # are these actually the numbers we want?
-            darkData[x:x+s, y:y+s] += self.nSigmaBright * darkWidth  # are these actually the numbers we want?
+            y, x, sy, sx = defect
+            # are these actually the numbers we want?
```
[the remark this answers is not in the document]
plazas: I don't know what the meaning of this question is. I believe it was Merlin who wrote it.

tests/test_defects.py
```diff
@@ -103,30 +108,472 @@ def setUp(self):
...
+        expectedDefects = [Box2I(corner = Point2I(20, 1), dimensions = Extent2I(1, 25))]
+        defects = self.allDefectsList
+        defects.append(Box2I(corner = Point2I(20, 1), dimensions = Extent2I(1, 25)))
+
+        self.check_maskBlocks(defects, expectedDefects)
+
+    def test_maskBlocks_short_column(self):
+        """A test for maskBlocksIfIntermitentBadPixelsInColumn (test 3).
+        Npix contiguous bad pixels in a column where Npix < badOnAndOffPixelColumnThreshold (10).
+        Tests that a contigous bad column does not get split by the code.
```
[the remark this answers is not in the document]
plazas: I have reworded the sentence as you suggest. I have removed the "input" and "expected" defects lists from all the docstrings, too.

tests/test_defects.py
```diff
@@ -103,30 +108,472 @@ def setUp(self):
...
+                     Box2I(corner = Point2I(35, 5), dimensions = Extent2I(1, 3)),
+                     Box2I(corner = Point2I(35, 11), dimensions = Extent2I(1, 2))]
+        for badBox in badPixels:
+            defects.append(badBox)
+
+        self.check_maskBlocks(defects, expectedDefects)
+
+    def test_maskBlocks_more_than_thresholds(self):
+        # Test 6: n discontiguous bad pixels in a column where n >= threshold, gap >= "good" threshold
+        # n=34 bad pixels total, 1 "good" gap big enough (13>=5 good pixels, from y=6 (1+5) to y=19)
```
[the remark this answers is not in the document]
plazas: I deleted this comment.

tests/test_defects.py
```diff
@@ -103,30 +108,472 @@ def setUp(self):
...
+        """
+
+        defects = self.allDefectsList
+        defects.append(Box2I(corner = Point2I(15, 1), dimensions = Extent2I(1, 50)))
+        expectedDefects = [Box2I(corner = Point2I(15, 1), dimensions = Extent2I(1, 50))]
+
+        self.check_maskBlocks(defects, expectedDefects)
+
+    def test_maskBlocks_long_column(self):
+        """A test for maskBlocksIfIntermitentBadPixelsInColumn (test 2).
```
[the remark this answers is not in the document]
plazas: I deleted this from all the docstrings too.

tests/test_defects.py
```diff
@@ -103,30 +108,472 @@ def setUp(self):
...
+        task = cpPipe.defects.FindDefectsTask(config=config)
+
+        defectsWithColumns = task.maskBlocksIfIntermitentBadPixelsInColumn(inputDefects)
+
+        boxesMeasured = []
+        for defect in defectsWithColumns:
+            boxesMeasured.append(defect.getBBox())
+
+        for boxInput in expectedDefects:
+            self.assertIn(boxInput, boxesMeasured)
```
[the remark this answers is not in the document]
plazas: Could you please elaborate on this? I think that the test needs to check that the expected defects are part of the set of measured defects, without requiring them to have them the same cardinality necessarily. In any case, I tried to implement the second option that you suggest "self.assertEqual(len(boxInput, boxesMeasured)", but I'm a bit confused because "boxInput" is just an element of the set  "expectedDefects".

tests/test_defects.py
plazas: Thanks for catching this. I did run the tests and they all passed before asking you to look at the code, so the code per se it's not incorrect, but rather incomplete in the sense that there was a column missing from the "expectedDefects" set. For the case of this test, I have added this extra expected defect (Box2I(corner = Point2I(62, 2), dimensions = Extent2I(1, 12))),  but I also added a few more for tests 11, 12, and 15, for the same reason.

tests/test_defects.py
plazas: See above. In this case, I added the defect "Box2I(corner = Point2I(69, 2), dimensions = Extent2I(1, 12))" to the "expectedDefects" set.

plazas: (I clicked on "resolve conversation" for this comment by accident---2020JAN09, 9:04 p.m Central time---then on "unresolve", but now it says "Outdated").

plazas:
Thank you again for your feedback. I have addressed your comments, have asked for clarification in one of them, and reran the tests code: all the tests pass.

tests/test_defects.py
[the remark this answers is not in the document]
mfisherlevine: `git blame` says it wasn't me.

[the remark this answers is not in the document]
jdswinbank: `git blame` is wrong. :-) Try [here](https://github.com/lsst/cp_pipe/commit/0712f5971e724ea276a0e532723169cdb135ef69#diff-24c51ebe1c07d9827b3a5873b938d54bR90).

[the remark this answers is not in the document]
mfisherlevine: _You_, sir, are the one who is wrong - I put them on a totally different line in the commit you reference! The fact that someone (who git correctly blames) moved them so that they now refer to entirely different code (my code that it referred to having been deleted) means that I am indeed _not_ to blame. So there! 🎩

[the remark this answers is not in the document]
mfisherlevine: Anyway, shenanigans and 💩-posting aside, I think these are the ~droids~ numbers you're looking for. They'll certainly do for now, anyway.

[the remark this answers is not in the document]
mfisherlevine: So feel free to delete that comment.

tests/test_defects.py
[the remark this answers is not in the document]
parejkoj: You should know exactly what defects should be produced by each test, without that you miss things like the test errors I described below (where the "blob" was broken up but you didn't check for the second part of it).

In both of my examples, replace `boxInput` with `expectedDefects`: I've updated the comment. Sadly, `Defect`s are not sortable, so my `sorted` test wouldn't work (and I think we want sorting there, because the order of the output shouldn't matter).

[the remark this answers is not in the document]
plazas: I think that each of the tests added for this ticket should check that the set of expected defects in each test is contained entirely within the set of measured defects, which will be a set of defects found by the general defect finding code and by the extra columnar code added in this branch. As a consequence of this,  the extra condition that you suggest be added, `self.assertEqual(len(expectedDefects), len(boxesMeasured))`, will not be satisfied (I assumed that this is what you meant since in your comment I see `self.assertEqual(len(expectedDefects), boxesMeasured)`, but please correct me if I am wrong).  The `boxesMeasured` set will contain more elements than the `expectedDefects` set.  For example, in the test where we have bad pixels every other pixel along the whole column (`test_maskBlocks_every_other_pixel_bad_greater_than_threshold`), the defect-finding code that runs before the columnar code of this ticket will first find the intermittent bad pixels (and also the bad pixels defined for other tests in the code, such as the ones in lines 73 and 75 ), then, the columnar code of this branch will mark the whole column as a bad region because the two conditions for doing so are triggered (namely, (a) the number of bad pixels in the column is greater than the threshold1=10, and (b) (given (a)  is satisfied), the number of consecutive good pixels is not greater than threshold2=5). This is indeed what happens when I run the code for this test: 

`EXPECTED:  [Box2I(minimum=Point2I(50, 10), dimensions=Extent2I(1, 31))]

MEASURED:  [Box2I(minimum=Point2I(0, 15), dimensions=Extent2I(3, 3)), Box2I(minimum=Point2I(100, 123), dimensions=Extent2I(1, 1)), Box2I(minimum=Point2I(5, 0), dimensions=Extent2I(1, 1)), Box2I(minimum=Point2I(7, 62), dimensions=Extent2I(2, 2)), Box2I(minimum=Point2I(50, 10), dimensions=Extent2I(1, 1)), Box2I(minimum=Point2I(50, 12), dimensions=Extent2I(1, 1)), Box2I(minimum=Point2I(50, 14), dimensions=Extent2I(1, 1)), Box2I(minimum=Point2I(50, 16), dimensions=Extent2I(1, 1)), Box2I(minimum=Point2I(50, 18), dimensions=Extent2I(1, 1)), Box2I(minimum=Point2I(50, 20), dimensions=Extent2I(1, 1)), Box2I(minimum=Point2I(50, 22), dimensions=Extent2I(1, 1)), Box2I(minimum=Point2I(50, 24), dimensions=Extent2I(1, 1)), Box2I(minimum=Point2I(50, 26), dimensions=Extent2I(1, 1)), Box2I(minimum=Point2I(50, 28), dimensions=Extent2I(1, 1)), Box2I(minimum=Point2I(50, 30), dimensions=Extent2I(1, 1)), Box2I(minimum=Point2I(50, 32), dimensions=Extent2I(1, 1)), Box2I(minimum=Point2I(50, 34), dimensions=Extent2I(1, 1)), Box2I(minimum=Point2I(50, 36), dimensions=Extent2I(1, 1)), Box2I(minimum=Point2I(50, 38), dimensions=Extent2I(1, 1)), Box2I(minimum=Point2I(50, 40), dimensions=Extent2I(1, 1)), Box2I(minimum=Point2I(50, 10), dimensions=Extent2I(1, 31))]`

I think that the critical thing is to make sure that all the defects in `EXPECTED` (in this case only one: `Box2I(minimum=Point2I(50, 10), dimensions=Extent2I(1, 31))`), are found in `MEASURED~, which is the case, and that's why the condition is 

`for boxInput in expectedDefects:
            self.assertIn(boxInput, boxesMeasured)`

The errors that you found in the previous comment had to do with an incomplete `EXPECTED` set (this set is determined by hand knowing what we would want the code in this branch to do, and hence the images in Jira ticket DM-19903 to aid in this), but they did not indicate that the code was not working (the code worked, it just I forgot to check for a few more extra boxes).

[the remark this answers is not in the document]
jdswinbank: Hi both —

Since I understand this issue is blocking this ticket being reviewed, we need to resolve it.

As I understand it, there are two issues here:

- @parejkoj is correct that we should test that `maskBlocksIfIntermitentBadPixelsInColumn` doesn't mask *more* than it is supposed to. By my (superficial) reading of the code, that's certainly a possibility — `boxesMeasured` could contain boxes that cover effectively the whole image in addition to the `expectedDefects` and the test would pass. I agree with him that this is less than ideal.

- @plazas is correct that `boxesMeasured` doesn't just contain the columnar defects, but *also* the inputs to `maskBlocksIfIntermittentBadPixelsInColumn`. I therefore don't think that checking for equality between `boxesMeasured` and `expectedDefects` is helpful.

Given that, I think the right checks are:

- Iterate over `boxesMeasured`, and, for *every* entry, ensure that it appears in *either* `inputDefects` *or* `expectedDefects`.
- Iterate over `expectedDefects`, and, for *every* entry, ensure that it appears in `boxesMeasured` (this is what's already happening).

I think that's just another extra couple of lines of code.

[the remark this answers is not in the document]
plazas: Thank you, @jdswinbank. I have implemented your suggestions. @parejkoj, please let me know if things look good for you.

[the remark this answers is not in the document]
plazas: (and the tests pass)

tests/test_defects.py
```diff
@@ -140,6 +140,17 @@ def check_maskBlocks(self, inputDefects, expectedDefects):
         for boxInput in expectedDefects:
             self.assertIn(boxInput, boxesMeasured)
 
+        # Union set of input and expected defects, as boxes.
```
parejkoj: This is not a set, it is a list; did you mean for it to be a set? I don't remember if BBox is hashable.

I think you should move a version of the comment below up to here. Maybe something like this?

```
# Check that the code did not mask anything outside extra by
# looking in both the input list and "expanded-column" list.
```

timj: BBox is not hashable. We really have to consider how much of this code should really be inside `Defects`  -- the fact that internally it's all `BBox` is really an implementation detail.  This feels like it's not code that is specific to cp_pipe but more generally something that is comparing Defects.

jdswinbank: Let's not consider that on this ticket. :-)

If this is functional, let's clean up that comment and merge it. If you think we need to consider re-homing it, please file another ticket to cover that.

plazas: Thanks, @parejkoj. I just had in mind the regular definition of a set in naïve set theory as a collection of objects, nothing technical. I'll use your suggested comment instead.

PR #20, opened 2019-11-20
DM-21212: Update existing cp_pipe tasks to pipelineTasks

czwa:
Tickets/dm 21212

This ticket converts the existing pipe_drivers constructCalibs.py tasks to Gen3, and uses pipelines to handle multiprocessing, configuration, and task handling.

### DIFF
--- a/bin.src/blessCalibration.py
+++ b/bin.src/blessCalibration.py
@@ -1,0 +1,77 @@
+#!/usr/bin/env python
+#
+# This file is part of cp_pipe.
+#
+# Developed for the LSST Data Management System.
+# This product includes software developed by the LSST Project
+# (https://www.lsst.org).
+# See the COPYRIGHT file at the top-level directory of this distribution
+# for details of code ownership.
+#
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <https://www.gnu.org/licenses/>.
+#
+"""Mark cp_pipe generated calibrations as valid, and register them with
+the butler with an appropriate use date range.
+"""
+from lsst.cp.pipe.cpCertify import BlessCalibration
+
+import argparse
+import logging
+
+import lsst.log
+from lsst.log import Log
+
+from lsst.daf.butler import Butler
+
+if __name__ == "__main__":
+    parser = argparse.ArgumentParser(
+        description=__doc__,
+        formatter_class=argparse.RawDescriptionHelpFormatter
+    )
+    parser.add_argument("root", help="Path to butler to use")
+    parser.add_argument("inputCollection", help="Input collection to pull from.")
+    parser.add_argument("outputCollection", help="Output collection to add to.")
+    parser.add_argument("datasetTypeName", help="Dataset type to bless.")
+
+    parser.add_argument("-v", "--verbose", action="store_const", dest="logLevel",
+                        default=Log.INFO, const=Log.DEBUG,
+                        help="Set the log level to DEBUG.")
+    parser.add_argument("-b", "--beginDate",
+                        help="Start date for using the calibration")
+    parser.add_argument("-e", "--endDate",
+                        help="End date for using the calibration")
+    parser.add_argument("-s", "--skipCalibrationLabel", action="store_true",
+                        default=False, dest="skipCL",
+                        help="Do not attempt to register the calibration label.")
+
+    args = parser.parse_args()
+    log = Log.getLogger("lsst.daf.butler")
+    log.setLevel(args.logLevel)
+
+    # DM-22527: Clean up syntax/log handling in gen3 repo scripts.
+    lgr = logging.getLogger("lsst.daf.butler")
+    lgr.setLevel(logging.INFO if args.logLevel == Log.INFO else logging.DEBUG)
+    lgr.addHandler(lsst.log.LogHandler())
+
+    butler = Butler(args.root, run=args.inputCollection)
+
+    # Do the thing.
+    crozier = BlessCalibration(butler=butler,
+                               inputCollection=args.inputCollection,
+                               outputCollection=args.outputCollection)
+
+    crozier.findInputs(args.datasetTypeName)
+    if not args.skipCL:
+        crozier.addCalibrationLabel(beginDate=args.beginDate, endDate=args.endDate)
+    crozier.registerCalibrations(args.datasetTypeName)
--- a/pipelines/cpBias.yaml
+++ b/pipelines/cpBias.yaml
@@ -1,0 +1,30 @@
+description: cp_pipe BIAS calibration construction
+tasks:
+  isr:
+    class: lsst.ip.isr.isrTask.IsrTask
+    config:
+      connections.ccdExposure: 'raw'
+      connections.outputExposure: 'cpBiasProc'
+      doBias: False
+      doVariance: True
+      doLinearize: False
+      doCrosstalk: False
+      doDefect: True
+      doNanMasking: True
+      doInterpolate: True
+      doBrighterFatter: False
+      doDark: False
+      doFlat: False
+      doApplyGains: False
+      doFringe: False
+  cpCombine:
+    class: lsst.cp.pipe.cpCombine.CalibCombineTask
+    config:
+      connections.inputExps: 'cpBiasProc'
+      connections.outputData: 'biasProposal'
+      calibrationType: 'bias'
+      exposureScaling: "None"
+contracts:
+  - isr.doBias == False
+  - cpCombine.calibrationType == "bias"
+  - cpCombine.exposureScaling == "None"
--- a/pipelines/cpDark.yaml
+++ b/pipelines/cpDark.yaml
@@ -1,0 +1,36 @@
+description: cp_pipe DARK calibration construction
+tasks:
+  isr:
+    class: lsst.ip.isr.isrTask.IsrTask
+    config:
+      connections.ccdExposure: 'raw'
+      connections.outputExposure: 'cpDarkIsr'
+      doBias: True
+      doVariance: True
+      doLinearize: True
+      doCrosstalk: True
+      doDefect: True
+      doNanMasking: True
+      doInterpolate: True
+      doBrighterFatter: False
+      doDark: False
+      doFlat: False
+      doApplyGains: False
+      doFringe: False
+  cpDark:
+    class: lsst.cp.pipe.cpDarkTask.CpDarkTask
+    config:
+      connections.inputExp: 'cpDarkIsr'
+      connections.outputExp: 'cpDarkProc'
+  cpCombine:
+    class: lsst.cp.pipe.cpCombine.CalibCombineTask
+    config:
+      connections.inputExps: 'cpDarkProc'
+      connections.outputData: 'darkProposal'
+      calibrationType: 'dark'
+      exposureScaling: "DarkTime"
+      python: config.mask.append("CR")
+contracts:
+  - isr.doDark == False
+  - cpCombine.calibrationType == "dark"
+  - cpCombine.exposureScaling == "DarkTime"
--- a/pipelines/cpFlat.yaml
+++ b/pipelines/cpFlat.yaml
@@ -1,0 +1,41 @@
+description: cp_pipe FLAT calibration construction optimized for single-CCD cameras
+tasks:
+  isr:
+    class: lsst.ip.isr.isrTask.IsrTask
+    config:
+      connections.ccdExposure: 'raw'
+      connections.outputExposure: 'cpFlatProc'
+      doBias: True
+      doVariance: True
+      doLinearize: True
+      doCrosstalk: True
+      doDefect: True
+      doNanMasking: True
+      doInterpolate: True
+      doDark: True
+      doBrighterFatter: False
+      doFlat: False
+      doFringe: False
+      doApplyGains: False
+  cpFlatMeasure:
+    class: lsst.cp.pipe.cpFlatNormTask.CpFlatMeasureTask
+    config:
+      connections.inputExp: 'cpFlatProc'
+      connections.outputStats: 'flatStats'
+  cpFlatNorm:
+    class: lsst.cp.pipe.cpFlatNormTask.CpFlatNormalizationTask
+    config:
+      connections.inputMDs: 'flatStats'
+      connections.outputScales: 'cpFlatNormScales'
+  cpCombine:
+    class: lsst.cp.pipe.cpCombine.CalibCombineTask
+    config:
+      connections.inputExps: 'cpFlatProc'
+      connections.inputScales: 'cpFlatNormScales'
+      connections.outputData: 'flatProposal'
+      calibrationType: 'flat'
+      calibrationDimensions: ['physical_filter']
+      exposureScaling: InputList
+      scalingLevel: AMP
+contracts:
+  - isr.doFlat == False
--- a/python/lsst/cp/pipe/cpCombine.py
+++ b/python/lsst/cp/pipe/cpCombine.py
@@ -104,7 +104,7 @@ class CalibCombineConnections(pipeBase.PipelineTaskConnections,
     inputScales = cT.Input(
         name="cpScales",
         doc="Input scale factors to use.",
-        storageClass="StructuredDataDictYaml",
+        storageClass="StructuredDataDict",
         dimensions=("instrument", ),
         multiple=False,
     )
--- a/python/lsst/cp/pipe/cpCombine.py
+++ b/python/lsst/cp/pipe/cpCombine.py
@@ -1,0 +1,581 @@
+# This file is part of cp_pipe.
+#
+# Developed for the LSST Data Management System.
+# This product includes software developed by the LSST Project
+# (https://www.lsst.org).
+# See the COPYRIGHT file at the top-level directory of this distribution
+# for details of code ownership.
+#
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+import numpy as np
+import time
+
+import lsst.pex.config as pexConfig
+import lsst.pipe.base as pipeBase
+import lsst.pipe.base.connectionTypes as cT
+import lsst.afw.math as afwMath
+import lsst.afw.image as afwImage
+
+from lsst.geom import Point2D
+from lsst.log import Log
+from astro_metadata_translator import merge_headers, ObservationGroup
+from astro_metadata_translator.serialize import dates_to_fits
+
+
+# CalibStatsConfig/CalibStatsTask from pipe_base/constructCalibs.py
+class CalibStatsConfig(pexConfig.Config):
+    """Parameters controlling the measurement of background statistics.
+    """
+    stat = pexConfig.Field(
+        dtype=int,
+        default=int(afwMath.MEANCLIP),
+        doc="Statistic to use to estimate background (from lsst.afw.math)",
+    )
+    clip = pexConfig.Field(
+        dtype=float,
+        default=3.0,
+        doc="Clipping threshold for background",
+    )
+    nIter = pexConfig.Field(
+        dtype=int,
+        default=3,
+        doc="Clipping iterations for background",
+    )
+    mask = pexConfig.ListField(
+        dtype=str,
+        default=["DETECTED", "BAD", "NO_DATA"],
+        doc="Mask planes to reject",
+    )
+
+
+class CalibStatsTask(pipeBase.Task):
+    """Measure statistics on the background
+
+    This can be useful for scaling the background, e.g., for flats and fringe frames.
+    """
+    ConfigClass = CalibStatsConfig
+
+    def run(self, exposureOrImage):
+        """Measure a particular statistic on an image (of some sort).
+
+        Parameters
+        ----------
+        exposureOrImage : `lsst.afw.image.Exposure`, `lsst.afw.image.MaskedImage`, or `lsst.afw.image.Image`
+           Exposure or image to calculate statistics on.
+
+        Returns
+        -------
+        results : float
+           Resulting statistic value.
+        """
+        stats = afwMath.StatisticsControl(self.config.clip, self.config.nIter,
+                                          afwImage.Mask.getPlaneBitMask(self.config.mask))
+        try:
+            image = exposureOrImage.getMaskedImage()
+        except Exception:
+            try:
+                image = exposureOrImage.getImage()
+            except Exception:
+                image = exposureOrImage
+
+        return afwMath.makeStatistics(image, self.config.stat, stats).getValue()
+
+
+class CalibCombineConnections(pipeBase.PipelineTaskConnections,
+                              dimensions=("instrument", "detector")):
+    inputExps = cT.Input(
+        name="cpInputs",
+        doc="Input pre-processed exposures to combine.",
+        storageClass="ExposureF",
+        dimensions=("instrument", "detector", "exposure"),
+        multiple=True,
+    )
+    inputScales = cT.Input(
+        name="cpScales",
+        doc="Input scale factors to use.",
+        storageClass="StructuredDataDictYaml",
+        dimensions=("instrument", ),
+        multiple=False,
+    )
+
+    outputData = cT.Output(
+        name="cpProposal",
+        doc="Output combined proposed calibration.",
+        storageClass="ExposureF",
+        dimensions=("instrument", "detector"),
+    )
+
+    def __init__(self, *, config=None):
+        super().__init__(config=config)
+
+        if config and len(config.calibrationDimensions) != 0:
+            newDimensions = tuple(config.calibrationDimensions)
+            newOutputData = cT.Output(
+                name=self.outputData.name,
+                doc=self.outputData.doc,
+                storageClass=self.outputData.storageClass,
+                dimensions=self.allConnections['outputData'].dimensions + newDimensions
+            )
+            self.dimensions.update(config.calibrationDimensions)
+            self.outputData = newOutputData
+
+            if config.exposureScaling == 'InputList':
+                newInputScales = cT.PrerequisiteInput(
+                    name=self.inputScales.name,
+                    doc=self.inputScales.doc,
+                    storageClass=self.inputScales.storageClass,
+                    dimensions=self.allConnections['inputScales'].dimensions + newDimensions
+                )
+                self.dimensions.update(config.calibrationDimensions)
+                self.inputScales = newInputScales
+
+
+# CalibCombineConfig/CalibCombineTask from pipe_base/constructCalibs.py
+class CalibCombineConfig(pipeBase.PipelineTaskConfig,
+                         pipelineConnections=CalibCombineConnections):
+    """Configuration for combining calib exposures.
+    """
+    calibrationType = pexConfig.Field(
+        dtype=str,
+        default="calibration",
+        doc="Name of calibration to be generated.",
+    )
+    calibrationDimensions = pexConfig.ListField(
+        dtype=str,
+        default=[],
+        doc="List of updated dimensions to append to output."
+    )
+
+    exposureScaling = pexConfig.ChoiceField(
+        dtype=str,
+        allowed={
+            "None": "No scaling used.",
+            "ExposureTime": "Scale inputs by their exposure time.",
+            "DarkTime": "Scale inputs by their dark time.",
+            "MeanStats": "Scale inputs based on their mean values.",
+            "InputList": "Scale inputs based on a list of values.",
+        },
+        default=None,
+        doc="Scaling to be applied to each input exposure.",
+    )
+    scalingLevel = pexConfig.ChoiceField(
+        dtype=str,
+        allowed={
+            "DETECTOR": "Scale by detector.",
+            "AMP": "Scale by amplifier.",
+        },
+        default="DETECTOR",
+        doc="Region to scale.",
+    )
+    maxVisitsToCalcErrorFromInputVariance = pexConfig.Field(
+        dtype=int,
+        default=5,
+        doc="Maximum number of visits to estimate variance from input variance, not per-pixel spread",
+    )
+
+    doVignette = pexConfig.Field(
+        dtype=bool,
+        default=False,
+        doc="Copy vignette polygon to output and censor vignetted pixels?"
+    )
+
+    mask = pexConfig.ListField(
+        dtype=str,
+        default=["SAT", "DETECTED", "INTRP"],
+        doc="Mask planes to respect",
+    )
+    combine = pexConfig.Field(
+        dtype=int,
+        default=int(afwMath.MEANCLIP),
+        doc="Statistic to use for combination (from lsst.afw.math)",
+    )
+    clip = pexConfig.Field(
+        dtype=float,
+        default=3.0,
+        doc="Clipping threshold for combination",
+    )
+    nIter = pexConfig.Field(
+        dtype=int,
+        default=3,
+        doc="Clipping iterations for combination",
+    )
+    stats = pexConfig.ConfigurableField(
+        target=CalibStatsTask,
+        doc="Background statistics configuration",
+    )
+
+
+class CalibCombineTask(pipeBase.PipelineTask,
+                       pipeBase.CmdLineTask):
+    """Task to combine calib exposures."""
+    ConfigClass = CalibCombineConfig
+    _DefaultName = 'cpCombine'
+
+    def __init__(self, **kwargs):
+        super().__init__(**kwargs)
+        self.makeSubtask("stats")
+
+    def runQuantum(self, butlerQC, inputRefs, outputRefs):
+        inputs = butlerQC.get(inputRefs)
+
+        dimensions = [exp.dataId.byName() for exp in inputRefs.inputExps]
+        inputs['inputDims'] = dimensions
+
+        outputs = self.run(**inputs)
+        butlerQC.put(outputs, outputRefs)
+
+    def run(self, inputExps, inputScales=None, inputDims=None):
+        """Combine calib exposures for a single detector.
+
+        Parameters
+        ----------
+        inputExps : `list` [`lsst.afw.image.Exposure`]
+            Input list of exposures to combine.
+        inputScales : `dict` [`dict` [`dict` [`float`]]], optional
+            Dictionary of scales, indexed by detector (`int`),
+            amplifier (`int`), and exposure (`int`).  Used for
+            'inputList' scaling.
+        inputDims : `list` [`dict`]
+            List of dictionaries of input data dimensions/values.
+            Each list entry should contain:
+
+            ``"exposure"``
+                exposure id value (`int`)
+            ``"detector"``
+                detector id value (`int`)
+
+        Returns
+        -------
+        combinedExp : `lsst.afw.image.Exposure`
+            Final combined exposure generated from the inputs.
+
+        Raises
+        ------
+        RuntimeError
+            Raised if no input data is found.  Also raised if
+            config.exposureScaling == InputList, and a necessary scale
+            was not found.
+        """
+        width, height = self.getDimensions(inputExps)
+        stats = afwMath.StatisticsControl(self.config.clip, self.config.nIter,
+                                          afwImage.Mask.getPlaneBitMask(self.config.mask))
+        numExps = len(inputExps)
+        if numExps < 1:
+            raise RuntimeError("No valid input data")
+        if numExps < self.config.maxVisitsToCalcErrorFromInputVariance:
+            stats.setCalcErrorFromInputVariance(True)
+
+        # Create output exposure for combined data.
+        combined = afwImage.MaskedImageF(width, height)
+        combinedExp = afwImage.makeExposure(combined)
+
+        # Apply scaling:
+        expScales = []
+        if inputDims is None:
+            inputDims = [dict() for i in inputExps]
+
+        for index, (exp, dims) in enumerate(zip(inputExps, inputDims)):
+            scale = 1.0
+            if exp is None:
+                self.log.warn("Input %d is None (%s); unable to scale exp.", index, dims)
+                continue
+
+            if self.config.exposureScaling == "ExposureTime":
+                scale = exp.getInfo().getVisitInfo().getExposureTime()
+            elif self.config.exposureScaling == "DarkTime":
+                scale = exp.getInfo().getVisitInfo().getDarkTime()
+            elif self.config.exposureScaling == "MeanStats":
+                scale = self.stats.run(exp)
+            elif self.config.exposureScaling == "InputList":
+                visitId = dims.get('exposure', None)
+                detectorId = dims.get('detector', None)
+                if visitId is None or detectorId is None:
+                    raise RuntimeError(f"Could not identify scaling for input {index} ({dims})")
+                if detectorId not in inputScales['expScale']:
+                    raise RuntimeError(f"Could not identify a scaling for input {index}"
+                                       f" detector {detectorId}")
+
+                if self.config.scalingLevel == "DETECTOR":
+                    if visitId not in inputScales['expScale'][detectorId]:
+                        raise RuntimeError(f"Could not identify a scaling for input {index}"
+                                           f"detector {detectorId} visit {visitId}")
+                    scale = inputScales['expScale'][detectorId][visitId]
+                elif self.config.scalingLevel == 'AMP':
+                    scale = [inputScales['expScale'][detectorId][amp.getName()][visitId]
+                             for amp in exp.getDetector()]
+                else:
+                    raise RuntimeError(f"Unknown scaling level: {self.config.scalingLevel}")
+            elif self.config.exposureScaling == 'None':
+                scale = 1.0
+            else:
+                raise RuntimeError(f"Unknown scaling type: {self.config.exposureScaling}.")
+
+            expScales.append(scale)
+            self.log.info("Scaling input %d by %s", index, scale)
+            self.applyScale(exp, scale)
+
+        self.combine(combined, inputExps, stats)
+
+        self.interpolateNans(combined)
+
+        if self.config.doVignette:
+            polygon = inputExps[0].getInfo().getValidPolygon()
+            VignetteExposure(combined, polygon=polygon, doUpdateMask=True,
+                             doSetValue=True, vignetteValue=0.0)
+
+        # Combine headers
+        self.combineHeaders(inputExps, combinedExp,
+                            calibType=self.config.calibrationType, scales=expScales)
+
+        # Return
+        return pipeBase.Struct(
+            outputData=combinedExp,
+        )
+
+    def getDimensions(self, expList):
+        """Get dimensions of the inputs.
+
+        Parameters
+        ----------
+        expList : `list` [`lsst.afw.image.Exposure`]
+            Exps to check the sizes of.
+
+        Returns
+        -------
+        width, height : `int`
+            Unique set of input dimensions.
+        """
+        dimList = [exp.getDimensions() for exp in expList if exp is not None]
+        return self.getSize(dimList)
+
+    def getSize(self, dimList):
+        """Determine a consistent size, given a list of image sizes.
+
+        Parameters
+        -----------
+        dimList : iterable of `tuple` (`int`, `int`)
+            List of dimensions.
+
+        Raises
+        ------
+        RuntimeError
+            If input dimensions are inconsistent.
+
+        Returns
+        --------
+        width, height : `int`
+            Common dimensions.
+        """
+        dim = set((w, h) for w, h in dimList)
+        if len(dim) != 1:
+            raise RuntimeError("Inconsistent dimensions: %s" % dim)
+        return dim.pop()
+
+    def applyScale(self, exposure, scale=None):
+        """Apply scale to input exposure.
+
+        This implementation applies a flux scaling: the input exposure is
+        divided by the provided scale.
+
+        Parameters
+        ----------
+        exposure : `lsst.afw.image.Exposure`
+            Exposure to scale.
+        scale : `float` or `list` [`float`], optional
+            Constant scale to divide the exposure by.
+        """
+        if scale is not None:
+            mi = exposure.getMaskedImage()
+            if isinstance(scale, list):
+                for amp, ampScale in zip(exposure.getDetector(), scale):
+                    ampIm = mi[amp.getBBox()]
+                    ampIm /= ampScale
+            else:
+                mi /= scale
+
+    def combine(self, target, expList, stats):
+        """Combine multiple images.
+
+        Parameters
+        ----------
+        target : `lsst.afw.image.Exposure`
+            Output exposure to construct.
+        expList : `list` [`lsst.afw.image.Exposure`]
+            Input exposures to combine.
+        stats : `lsst.afw.math.StatisticsControl`
+            Control explaining how to combine the input images.
+        """
+        images = [img.getMaskedImage() for img in expList if img is not None]
+        afwMath.statisticsStack(target, images, afwMath.Property(self.config.combine), stats)
+
+    def combineHeaders(self, expList, calib, calibType="CALIB", scales=None):
+        """Combine input headers to determine the set of common headers,
+        supplemented by calibration inputs.
+
+        Parameters
+        ----------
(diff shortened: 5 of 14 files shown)


timj:
If this is constructCalibs.py and pipe_drivers is no longer used, where is the code that reconstructs the output header so that header propagation from inputs to outputs works properly? (see DM-16292).

PaulPrice:
Can you clean up the commit history before I review, please?

mfisherlevine:
Was there a random RHL commit that got sucked into one of these on the cleanup? The changes to `ptc.py` seem  unrelated to this ticket. Not complaining, just wanting to make sure I'm understanding as I have changes pending to that file on DM-18683 and don't want to mess anything up (though they should be orthogonal to those changes).

python/lsst/cp/pipe/cpFunctions.py
```diff
@@ -0,0 +1,516 @@
...
+
+class BlessCalibration(pipeBase.Task):
+    """Create a way to bless existing calibration products.
+
+    The inputs are assumed to have been constructed via cp_pipe, and
+    already exist in the butler.
+
+    Parameters
+    ----------
+    kwargs : `dict`
```
PaulPrice: Why specify this as `**kwargs` instead of actual values? I think this would be better:

    def __init__(self, *, butler, inputCollection, outputCollection, **kwargs):

But then `kwargs` would be unused. Why is there no `Task.__init__(**kwargs)`?

python/lsst/cp/pipe/cpFunctions.py
```diff
@@ -0,0 +1,516 @@
...
+        Parameters
+        ----------
+        datasetTypeName : `str`
+            Dataset that will be blessed.
+        inputTypeString : `str`
+            Dataset name for the input datasets.
+
+        Raises
+        ------
+        RuntimeError :
```
PaulPrice: No colon.

python/lsst/cp/pipe/cpFunctions.py
```diff
@@ -0,0 +1,516 @@
...
+
+    def findInputs(self, datasetTypeName, inputTypeString=None):
+        """Find and prepare inputs for blessing.
+
+        Parameters
+        ----------
+        datasetTypeName : `str`
+            Dataset that will be blessed.
+        inputTypeString : `str`
+            Dataset name for the input datasets.
```
PaulPrice: State what the default setting is. Add ", optional" to the type.

python/lsst/cp/pipe/cpFunctions.py
```diff
@@ -0,0 +1,516 @@
...
+
+        Raises
+        ------
+        RuntimeError :
+            Raised if no input datasets found or if the calibration
+            label exists and is not empty.
+        """
+        if inputTypeString is None:
+            inputDatasetTypeName = datasetTypeName + "Proposal"
```
PaulPrice: What is `inputDatasetTypeName` if `inputTypeString` is not `None`?

python/lsst/cp/pipe/cpFunctions.py
```diff
@@ -0,0 +1,516 @@
...
+
+        self.inputValues = list(self.registry.queryDatasets(inputDatasetTypeName,
+                                                            collections=[self.inputCollection]))
+        if len(self.inputValues) == 0:
+            raise RuntimeError(f"No inputs found for dataset {inputDatasetTypeName} "
+                               f"in {self.inputCollection}")
+
+        # Construct calibration label and choose instrument to use.
+        self.calibrationLabel = "{}/{}".format(datasetTypeName,
+                                               self.inputCollection)
```
PaulPrice: Why split the line?
And you clearly know and like f-strings, so why not make this `f"{datasetTyopeName}/{self.inputCollection}"`?

python/lsst/cp/pipe/cpFunctions.py
```diff
@@ -0,0 +1,516 @@
...
+    """
+    def __init__(self, **kwargs):
+        self.butler = kwargs['butler']
+        self.registry = self.butler.registry
+
+        self.inputCollection = kwargs['inputCollection']
+        self.outputCollection = kwargs['outputCollection']
+
+        self.calibrationLabel = ''
+        self.instrument = ''
```
PaulPrice: Why choose empty string instead of `None`? `None` would cause more trouble if you use them accidentally.

python/lsst/cp/pipe/cpFunctions.py
```diff
@@ -0,0 +1,516 @@
...
+        self.inputValues = list(self.registry.queryDatasets(inputDatasetTypeName,
+                                                            collections=[self.inputCollection]))
+        if len(self.inputValues) == 0:
+            raise RuntimeError(f"No inputs found for dataset {inputDatasetTypeName} "
+                               f"in {self.inputCollection}")
+
+        # Construct calibration label and choose instrument to use.
+        self.calibrationLabel = "{}/{}".format(datasetTypeName,
+                                               self.inputCollection)
+        self.instrument = self.inputValues[0].dataId['instrument']
```
PaulPrice: Why are you setting instance attributes? This was _verboten_ in Gen2, and I don't think it's a good idea for Gen3 either.

python/lsst/cp/pipe/cpFunctions.py
```diff
@@ -0,0 +1,516 @@
...
+                raise RuntimeError(f"Existing entries for calibration_label {self.calibrationLabel}")
+        except LookupError:
+            pass
+
+        # Prepare combination of new data ids and object data:
+        self.newDataId = []
+        self.objects = []
+        for input in self.inputValues:
+            self.newDataId.append(input.dataId)
+            self.objects.append(self.butler.get(input))
```
PaulPrice: Classic case for using listcomps:
```
self.newDataId = [value.dataId for value in self.inputValues]
self.objects = [self.butler.get(value) for value in self.inputValues]
```

python/lsst/cp/pipe/cpFunctions.py
```diff
@@ -0,0 +1,516 @@
...
+
+    def registerCalibrations(self, datasetTypeName):
+        """Add blessed inputs to the output collection.
+
+        Parameters
+        ----------
+        datasetTypeName : `str`
+            Dataset type these calibrations will be registered for.
+        """
+
```
PaulPrice: Remove blank line.

python/lsst/cp/pipe/cpFunctions.py
```diff
@@ -0,0 +1,516 @@
...
+        run = self.registry.getRun(collection=self.outputCollection)
+        if run is None:
+            run = self.registry.makeRun(self.outputCollection)
+        self.butler.run = run
+        self.butler.collection = None
+
+        for newId, data in zip(self.newDataId, self.objects):
+            self.butler.put(data, datasetTypeName, dataId=newId,
+                            calibration_label=self.calibrationLabel,
+                            producer=None)
```
PaulPrice: Does this need to be atomic?

python/lsst/cp/pipe/cpFunctions.py
```diff
@@ -0,0 +1,516 @@
...
+        Parameters
+        ----------
+        name : `str`
+            A unique string for the calibration_label key.
+        instrument : `str`
+            Instrument this calibration is for.
+        beginDate : `str`
+            A hyphen delineated date string for the beginning of the valid date range.
+        endDate : `str`
+            A hyphen delineated date string for the end of the valid date range.
```
PaulPrice: "hyphen delineated" --> "ISO 8601". See https://xkcd.com/1179/

python/lsst/cp/pipe/cpFunctions.py
```diff
@@ -0,0 +1,516 @@
...
+        beginY, beginM, beginD = beginDate.split("-")
+        endY, endM, endD = endDate.split("-")
+
+        self.butler.registry.insertDimensionData(
+            "calibration_label",
+            {
+                "name": name,
+                "instrument": instrument,
+                "datetime_begin": datetime.datetime(int(beginY), int(beginM), int(beginD), 0, 0, 0),
+                "datetime_end": datetime.datetime(int(endY), int(endM), int(endD), 0, 0, 0)
```
PaulPrice: Use `datetime.fromisoformat(beginDate)`, etc.

python/lsst/cp/pipe/cpFunctions.py
```diff
@@ -0,0 +1,516 @@
...
+            }
+        )
+
+
+class CalibStatsConfig(pexConfig.Config):
+    """Parameters controlling the measurement of background statistics.
+    """
+    stat = pexConfig.Field(
+        dtype=int,
+        default=int(afwMath.MEANCLIP),
```
PaulPrice: I'm not sure this was ever a good idea, but it's my fault.

python/lsst/cp/pipe/cpFunctions.py
```diff
@@ -0,0 +1,516 @@
...
+
+class CalibStatsTask(pipeBase.Task):
+    """Measure statistics on the background
+
+    This can be useful for scaling the background, e.g., for flats and fringe frames.
+    """
+    ConfigClass = CalibStatsConfig
+
+    def run(self, exposureOrImage):
+        """!Measure a particular statistic on an image (of some sort).
```
PaulPrice: Remove the leading `!` (vestigial from doxygen).

python/lsst/cp/pipe/cpFunctions.py
```diff
@@ -0,0 +1,516 @@
...
+
+        Parameters
+        ----------
+        exposureOrImage : `lsst.afw.image.Exposure`, `lsst.afw.image.MaskedImage`, or `lsst.afw.image.Image`
+           Exposure or image to calculate statistics on.
+
+        Returns
+        -------
+        results : `lsst.afw.math.statistics`
```
PaulPrice: No, this returns a value, which is a `float`.

python/lsst/cp/pipe/cpSkyTask.py
```diff
@@ -0,0 +1,172 @@
...
+# along with this program.  If
+import numpy as np
+
+import lsst.pex.config as pexConfig
+import lsst.pipe.base as pipeBase
+import lsst.pipe.base.connectionTypes as cT
+import lsst.afw.math as afwMath
+import lsst.afw.image as afwImage
+
+from lsst.pipe.drivers.background import (FocalPlaneBackground, MaskObjectsTask, SkyMeasurementTask,
```
PaulPrice: pipe_drivers isn't listed as a dependency in the table file.
Which makes me wonder if you've used pipe_tasks anywhere, as that isn't there either.

python/lsst/cp/pipe/cpSkyTask.py
```diff
@@ -0,0 +1,172 @@
...
+            self.config.background.statistic
+        )
+
+        bgNew = afwMath.makeBackground(mi, ctrl)
+        icExpBkg = afwMath.BackgroundList((
+            bgNew,
+            afwMath.stringToInterpStyle(self.config.background.algorithm),
+            afwMath.stringToUndersampleStyle("REDUCE_INTERP_ORDER"),
+            afwMath.ApproximateControl.UNKNOWN,
+            0, 0, False  # CZW: It'd be nice to have some idea what these are.
```
PaulPrice: https://github.com/lsst/afw/blob/master/python/lsst/afw/math/backgroundList.py#L64-L65

python/lsst/cp/pipe/cpSkyTask.py
```diff
@@ -0,0 +1,172 @@
...
+                                                  camera)
+        bgModel.addCcd(inputExp)
+
+        scale = np.median(bgModel.getStatsImage().getArray())
+
+        # As constructCalibs SkyTask.scatterProcess
+        # Remeasure bkg with scaled version of the FP model removed
+        mi = inputExp.getMaskedImage()
+
+        bg = bgModel.toCcdBackground(inputExp.getDetector(), mi.getBBox())
```
PaulPrice: I think this has to be missing the crucial stage of gathering the `bgModel`s from all the CCDs in the visit. I very much doubt that this works as it's intended.

pipelines/cpBias.yaml
```diff
@@ -0,0 +1,26 @@
...
+      connections.outputExposure: 'cpBiasProc'
+      doBias: False
+      doDark: False
+      doFlat: False
+      doApplyGains: False
+      doFringe: False
+      doBrighterFatter: False
+      doVariance: True
+      doLinearize: True
+      doCrosstalk: True
```
PaulPrice: I think crosstalk should be switched off here.

Why is this list of configuration parameters in a different order from the others below?

pipelines/cpBias.yaml
```diff
@@ -0,0 +1,26 @@
...
+      doDefect: True
+      doNanMasking: True
+      doInterpolate: True
+  cpCombine:
+    class: lsst.cp.pipe.cpFunctions.CalibCombineTask
+    config:
+      connections.inputExps: 'cpBiasProc'
+      connections.outputData: 'biasProposal'
+      calibrationType: 'bias'
+      exposureScaling: "ExposureTime"
```
PaulPrice: Specifying these as configuration parameters doesn't seem right to me, because configuration parameters can be changed by the user. Is there another way? I'd rather see a scheme with subclasses of `CalibCombineTask`.

pipelines/cpFlat.yaml
```diff
@@ -0,0 +1,29 @@
+description: cp_pipe SKY calibration construction
```
PaulPrice: Everything below says "flat", not "sky".

pipelines/cpFlat.yaml
```diff
@@ -0,0 +1,29 @@
...
+      doApplyGains: False
+      doFringe: False
+  cpCombine:
+    class: lsst.cp.pipe.cpFunctions.CalibCombineTask
+    config:
+      connections.inputExps: 'cpFlatProc'
+      connections.outputData: 'flatProposal'
+      calibrationType: 'flat'
+      calibrationDimensions: ['physical_filter']
+      nIter: 10
```
PaulPrice: Why set this to 10? It's not set that way in the original.

pipelines/cpFlat.yaml
```diff
@@ -0,0 +1,29 @@
...
+      exposureScaling: "ExposureTime"
+      outputScaling: 1.0
```
PaulPrice: This removes the relative scaling between CCDs, which I think is very bad.

bin.src/blessCalibration.py
```diff
@@ -0,0 +1,74 @@
...
+                        default=False, dest="skipCL",
+                        help="Do not attempt to register the calibration label.")
+
+    args = parser.parse_args()
+    log = Log.getLogger("lsst.daf.butler")
+    log.setLevel(args.logLevel)
+
+    lgr = logging.getLogger("lsst.daf.butler")
+    lgr.setLevel(logging.INFO if args.logLevel == Log.INFO else logging.DEBUG)
+    lgr.addHandler(lsst.log.LogHandler())
```
PaulPrice: Is this logger tweaking desirable for production?

python/lsst/cp/pipe/cpFunctions.py
```diff
@@ -91,15 +91,6 @@ def findInputs(self, datasetTypeName, inputTypeString=None):
...
 
-        try:
-            existingValues = list(self.registry.queryDimensions(['calibration_label'],
-                                                                instrument=self.instrument,
-                                                                calibration_label=self.calibrationLabel))
-            if len(existingValues) != 0:
-                raise RuntimeError(f"Existing entries for calibration_label {self.calibrationLabel}")
-        except LookupError:
-            pass
-
```
PaulPrice: Why is this code now disappearing? If it's not supposed to be there, don't add it in the first place.

python/lsst/cp/pipe/cpFunctions.py
```diff
@@ -124,6 +115,12 @@ def registerCalibrations(self, datasetTypeName):
         self.butler.collection = None
 
         for newId, data in zip(self.newDataId, self.objects):
+            # Special case known special storageClasses.
+            if datasetTypeName in ('bias', 'dark'):
+                data = data.getImage()
+            elif datasetTypeName in ('flat'):
+                data = data.getMaskedImage()
+
```
PaulPrice: If this code should be there, put it there in the original.

python/lsst/cp/pipe/cpFunctions.py
```diff
@@ -124,6 +115,12 @@ def registerCalibrations(self, datasetTypeName):
         self.butler.collection = None
 
         for newId, data in zip(self.newDataId, self.objects):
+            # Special case known special storageClasses.
```
PaulPrice: Can I suggest putting this in a method so it's easy to supplement or override? This will make it easier to add new calibs in the future.

python/lsst/cp/pipe/ptc.py
```diff
@@ -199,19 +198,6 @@ def runDataRef(self, dataRef, visitPairs):
...
-        #
-        # Basically, the butler's trying to look up the fields in `raw_visit` which won't work
-        for name in dataRef.getButler().getKeys('bias'):
-            if name not in dataRef.dataId:
-                try:
-                    dataRef.dataId[name] = \
-                        dataRef.getButler().queryMetadata('raw', [name], detector=detNum)[0]
-                except OperationalError:
-                    pass
-
```
PaulPrice: Why is this part of this ticket?

python/lsst/cp/pipe/cpFunctions.py
czwa: This looks like a typo that wasn't detected by my testing.  Renaming inputTypeString to inputDatasetTypeName does what I want: allow searching for a default inputDatasetTypeName ("biasProposal" for example) when you're trying to bless the given datasetTypeName ("bias"), but let this be overridden if needed.

bin.src/blessCalibration.py
```diff
@@ -0,0 +1,74 @@
...
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <https://www.gnu.org/licenses/>.
+#
+
```
parejkoj: Please add a file docstring describing what this file is for.

You can use some kwargs to `ArgumentParser` to make the file docstring be the commandline docstring.

```
    parser = argparse.ArgumentParser(description=__doc__,
                                     formatter_class=argparse.RawDescriptionHelpFormatter)
```

python/lsst/cp/pipe/cpFringeTask.py
```diff
@@ -0,0 +1,133 @@
...
+
+        Returns
+        -------
+        outputExp : `lsst.afw.image.Exposure`
+            Fringe pre-processed frame.
+
+        """
+
+        """Detect +/-1 sigma sources/pixels, and mask them from the input
+        fringes.
```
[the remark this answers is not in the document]
czwa: Remnant from a now-rebased prior version.  Removed.

python/lsst/cp/pipe/cpSkyTask.py
```diff
@@ -0,0 +1,172 @@
...
+
+from lsst.pipe.drivers.background import (FocalPlaneBackground, MaskObjectsTask, SkyMeasurementTask,
+                                          FocalPlaneBackgroundConfig, BackgroundConfig)
+
+__all__ = ["CpSkyTask", "CpSkyTaskConfig"]
+
+
+class CpSkyConnections(pipeBase.PipelineTaskConnections,
+                       dimensions=("instrument", "physical_filter", "detector", "visit"),
+                       defaultTemplates={}):
```
natelust: dont need

python/lsst/cp/pipe/cpSkyTask.py
```diff
@@ -0,0 +1,172 @@
...
+    inputExp = cT.Input(
+        name="cpSkyISR",
+        doc="Input pre-processed exposures to combine.",
+        storageClass="ExposureF",
+        dimensions=("instrument", "visit", "detector"),
+        deferLoad=False,
+        multiple=False,
```
natelust: dont need these two

python/lsst/cp/pipe/cpSkyTask.py
```diff
@@ -0,0 +1,172 @@
...
+
+    outputExp = cT.Output(
+        name="cpSkyProc",
+        doc="Output combined proposed calibration.",
+        storageClass="ExposureF",
+        dimensions=("instrument", "physical_filter", "detector", "visit"),
+    )
+
+    def __init__(self, *, config=None):
+        super().__init__(config=config)
```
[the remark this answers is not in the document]
natelust: +1

czwa:
Resolved the phantom commit.  This should now contain only things from this branch.

python/lsst/cp/pipe/cpFunctions.py
czwa: I'm happy to think of the best option as part of a future ticket.  This should simply match existing code for the generation switch.

python/lsst/cp/pipe/cpFunctions.py
```diff
@@ -0,0 +1,516 @@
...
+            self.outputData = newOutputData
+
+
+class CalibCombineConfig(pipeBase.PipelineTaskConfig,
+                         pipelineConnections=CalibCombineConnections):
+    """Configuration for combining calib exposures.
+    """
+    calibrationType = pexConfig.Field(
+        dtype=str,
+        default="calibration",
```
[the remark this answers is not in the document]
czwa: I'm now questioning how much utility this option gives at all.  This currently only sets a header keyword based on this value, which is helpful to identify what the output product is if the butler is somehow unavailable.

python/lsst/cp/pipe/cpFunctions.py
```diff
@@ -0,0 +1,516 @@
...
+        default=None,
+        doc="Scaling to be applied to each input exposure.",
+    )
+    outputScaling = pexConfig.Field(
+        dtype=float,
+        default=float('nan'),
+        doc="Scaling value to apply to the output combined product.",
+    )
+
+    rows = pexConfig.Field(
```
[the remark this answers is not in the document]
czwa: I've removed rows, as that isn't currently implemented for the combine task (although it isn't an impossible thing to do).  Added as separate ticket: DM-22521.

python/lsst/cp/pipe/cpFunctions.py
czwa: Wrapped in transaction to be safe.

python/lsst/cp/pipe/cpDarkTask.py
```diff
@@ -0,0 +1,133 @@
...
+    _DefaultName = "cpDark"
+
+    def __init__(self, **kwargs):
+        super().__init__(**kwargs)
+        self.makeSubtask("repair")
+
+    def run(self, inputExp):
+        """Preprocess input exposures prior to DARK combination.
+
+        This task detects and repairs cosmic rays strikes.
```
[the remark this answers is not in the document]
czwa: Fixed.  The DarkTime scaling is needed to make repair.run() return a reasonable CR mask, but final scaling should be applied during combination.

pipelines/cpBias.yaml
czwa: Added contracts to fix pipeline-specific config values.

bin.src/blessCalibration.py
czwa: This duplicates other gen3 utility scripts.  Those scripts probably should be cleaned when they're made camera agnostic and moved to obs_base.  Ticket DM-22527 will do this work.

python/lsst/cp/pipe/cpFunctions.py
czwa: Peculiarities with SQLAlchemy is the short answer.  I thought the query would raise LookupError when called, but it only raises when the result is used.  The repack into an array does this.

python/lsst/cp/pipe/cpFunctions.py
czwa: Unexpected type mismatches between HSC (special cased when I made the gen3 repo two months ago) and LATISS (no special cases, so must properly type the data).

python/lsst/cp/pipe/cpFunctions.py
```diff
@@ -0,0 +1,571 @@
...
+        if run is None:
+            run = self.registry.makeRun(self.outputCollection)
+        self.butler.run = run
+        self.butler.collection = None
+
+        for newId, data in zip(self.newDataId, self.objects):
+            # Special case known special storageClasses.
+            if datasetTypeName in ('bias', 'dark'):
+                data = data.getImage()
+            elif datasetTypeName in ('flat'):
```
timj: I think you need a comma before the `)` here else you are looking for substrings within "flat" and not whether "flat" is in the tuple.

python/lsst/cp/pipe/cpFringeTask.py
```diff
@@ -0,0 +1,133 @@
...
+import lsst.cp.pipe.cpFunctions as cpFunctions
+import lsst.meas.algorithms as measAlg
+import lsst.afw.detection as afwDet
+
+
+__all__ = ["CpFringeTask", "CpFringeTaskConfig"]
+
+
+class CpFringeConnections(pipeBase.PipelineTaskConnections,
+                          dimensions=("instrument", "detector", "physical_filter", "visit"),
```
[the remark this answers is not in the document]
czwa: Updated in both connections.dimensions and output product dimensions.  Also updated to be less higgledy-piggledy in dimension lists.

python/lsst/cp/pipe/cpFunctions.py
```diff
@@ -0,0 +1,560 @@
...
+
+        if config and len(config.calibrationDimensions) != 0:
+            # CZW: this seems fake, but ok.
+            newOutputData = cT.Output(
+                name=self.outputData.name,
+                doc=self.outputData.doc,
+                storageClass=self.outputData.storageClass,
+                dimensions=self.allConnections['outputData'].dimensions + tuple(config.calibrationDimensions)
+            )
+            self.dimensions.update(config.calibrationDimensions)
```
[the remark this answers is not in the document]
czwa: That is what I want now.  I can't currently think up a need to change existing dimensions, so only adding additional constraints seems reasonable.

python/lsst/cp/pipe/cpFunctions.py
PaulPrice: Sorry, that comment was more about a lack of orthogonality in the commits than the code.

pipelines/cpBias.yaml
czwa: Agree on CT; switched off.  This creates slight differences between gen2 and gen3 bias results (gen2 is getting this enabled somehow).

Ordering reordered.

python/lsst/cp/pipe/cpSkyTask.py
czwa: I would rather not hold up the rest of the ticket for the sky calibration (which I've never been completely sure about anyway).  Is there documentation on the algorithm that I can link the ticket covering this issue (DM-22534).

PaulPrice: If the sky construction isn't to be part of this ticket, then you should put that in the ticket description, and remove this code completely. The algorithm is described fairly broadly [here](https://community.lsst.org/t/sky-subtraction/2415).

The flat construction has the same problem.

python/lsst/cp/pipe/cpFunctions.py
```diff
@@ -0,0 +1,560 @@
...
+            Instrument this calibration is for.
+        beginDate : `str`
+            A hyphen delineated date string for the beginning of the valid date range.
+        endDate : `str`
+            A hyphen delineated date string for the end of the valid date range.
+        """
+        if name is None:
+            name = self.calibrationLabel
+        if instrument is None:
+            instrument = self.instrument
```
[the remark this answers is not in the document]
czwa: I've added a check that should prevent None entries from getting to the butler (where I think it would fail anyway based on my tests).

czwa:
I will rebase and squash the commits before merging.

python/lsst/cp/pipe/cpFunctions.py
```diff
@@ -0,0 +1,682 @@
...
+    already exist in the butler.
+
+    Parameters
+    ----------
+    butler : `lsst.daf.butler.Butler`
+        Butler repository to use.
+    inputCollection : `str`
+        Data collection to pull calibrations from.
+    outputCollection : `str`
+        Data collection to store final calibrations.
```
PaulPrice: Add `**kwargs` as "Arguments to be forwarded to `Task.__init__`"

python/lsst/cp/pipe/cpFunctions.py
```diff
@@ -0,0 +1,682 @@
...
+    """
+    _DefaultName = 'BlessCalibration'
+    ConfigClass = pexConfig.Config
+
+    def __init__(self, *, butler=None, inputCollection=None, outputCollection=None,
+                 **kwargs):
+        super().__init__(**kwargs)
+        if butler is not None:
```
PaulPrice: Can this `Task` function at all if `butler is None`? If you're allowing for that possibility, you need to trap this case in the methods that rely on `self.butler` and `self.registry`.

Similarly for `inputCollection` and `outputCollection`.

python/lsst/cp/pipe/cpFunctions.py
PaulPrice: `Task` was built to iterate over multiple datasets, so storing state in the `Task` was forbidden as it allows the potential for state from one dataset to interfere with another. In this case, you're not really using `Task` in that manner, so that's not really a concern. However, you are making the state kind of a global variable, which always scares me. I'd rather see you return a `BlessInputs` object that gets passed to the next method. That way, there are no hidden inputs and outputs.

PaulPrice: Also, by storing state in the `Task` instance, you're establishing an expected sequence of operations, but that sequence isn't enforced (or documented) anywhere, and the only hint to the user that they've got the sequence wrong is going to be an obscure `AttributeError` or similar.

python/lsst/cp/pipe/cpFunctions.py
```diff
@@ -0,0 +1,682 @@
...
+        return data
+
+    def registerDatasetType(self, data, datasetTypeName, dataId):
+        storageClass = 'ExposureF'
+        if datasetTypeName in ('bias', 'dark'):
+            storageClass = 'ImageF'
+        elif datasetTypeName in ('flat', ):
+            storageClass = 'MaskedImageF'
+
+        dimensionArray = list(dataId.keys()) + ["calibration_label"]
```
PaulPrice: If want this to be a `set` eventually, why start it as a `list`?

python/lsst/cp/pipe/cpFunctions.py
```diff
@@ -0,0 +1,682 @@
...
+        Parameters
+        ----------
+        name : `str`
+            A unique string for the calibration_label key.
+        instrument : `str`
+            Instrument this calibration is for.
+        beginDate : `str`
+            An ISO 8601 date string for the beginning of the valid date range.
+        endDate : `str`
+            An ISO 8601 date string for the end of the valid date range.
```
PaulPrice: Note for each that they are optional.

python/lsst/cp/pipe/cpFunctions.py
```diff
@@ -0,0 +1,682 @@
...
+        ------
+        RuntimeError :
+            Raised if the instrument or calibration_label name are not set.
+        """
+        if name is None:
+            name = self.calibrationLabel
+        if instrument is None:
+            instrument = self.instrument
+        if name is None or instrument is None:
+            raise RuntimeError("Instrument and calibration_label name not set.")
```
PaulPrice: Condition says `or`, but error says `and`.

python/lsst/cp/pipe/cpFunctions.py
```diff
@@ -0,0 +1,682 @@
...
+
+        try:
+            existingValues = self.registry.queryDimensions(['calibration_label'],
+                                                           instrument=self.instrument,
+                                                           calibration_label=name)
+            existingValues = [a for a in existingValues]
+            print(f"Found {len(existingValues)} Entries for {self.calibrationLabel}")
```
PaulPrice: You need to tell the user here that this means you're not going to do anything. In particular, they may be expecting to change the `beginDate` and `endDate` (how do I do that if this won't let me do it?).
Why use `print` instead of `self.log`?

python/lsst/cp/pipe/cpFunctions.py
```diff
@@ -0,0 +1,682 @@
...
+        except LookupError:
+            self.butler.registry.insertDimensionData(
+                "calibration_label",
+                {
+                    "name": name,
+                    "instrument": instrument,
+                    "datetime_begin": datetime.datetime.fromisoformat(beginDate),
+                    "datetime_end": datetime.datetime.fromisoformat(endDate),
+                }
+            )
```
PaulPrice: I think the above belongs in its own file.

python/lsst/cp/pipe/cpFunctions.py
```diff
@@ -0,0 +1,682 @@
...
+        Parameters
+        ----------
+        exposure : `lsst.afw.image.Exposure`
+            Exposure to scale.
+        scale : `float` or `list` [`float`], optional
+            Constant scale to divide the exposure by.
+        """
+        if scale is not None:
+            mi = exposure.getMaskedImage()
+            if isinstance(scale, list):
```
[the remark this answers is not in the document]
timj: You have to be a little careful with strings:
```python
>>> import collections
>>> isinstance("xyz", collections.Iterable)
__main__:1: DeprecationWarning: Using or importing the ABCs from 'collections' instead of from 'collections.abc' is deprecated since Python 3.3,and in 3.9 it will stop working
True
```
(noting that it's `collections.abc.Iterable` these days)

python/lsst/cp/pipe/cpFlatNormTask.py
```diff
@@ -0,0 +1,166 @@
...
+    )
+    camera = cT.PrerequisiteInput(
+        name="camera",
+        doc="Input camera to use for gain lookup.",
+        storageClass="Camera",
+        dimensions=("instrument", "calibration_label"),
+    )
+
+    outputScales = cT.Output(
+        name="cpFlatProposal",
```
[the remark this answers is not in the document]
czwa: +1 Changed to 'cpFlatNorm_scales'.

python/lsst/cp/pipe/cpFlatNormTask.py
```diff
@@ -0,0 +1,166 @@
...
+    )
+
+
+class CpFlatNormalizationTaskConfig(pipeBase.PipelineTaskConfig,
+                                    pipelineConnections=CpFlatNormalizationConnections):
+    polyOrder = pexConfig.Field(
+        dtype=int,
+        doc="Polynomial order in each dimension to use to estimate illumination.",
+        default=7,
+    )
```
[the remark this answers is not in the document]
czwa: Removed.

jdswinbank:
With apologies for the slightly tedious request — could we please change the terminology here to avoid the religious overtones of “bless”? Much appreciated if we could use something more neutral instead (approve, sanction, something like that).

pipelines/cpBias.yaml
```diff
@@ -0,0 +1,30 @@
+description: cp_pipe BIAS calibration construction
+tasks:
+  isr:
+    class: lsst.ip.isr.isrTask.IsrTask
+    config:
+      connections.ccdExposure: 'raw'
+      connections.outputExposure: 'cpBiasProc'
+      doBias: False
+      doVariance: True
+      doLinearize: True
```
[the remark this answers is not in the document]
czwa: +1
I think this came up from matching gen2 results, but I agree.

pipelines/cpBias.yaml
```diff
@@ -0,0 +1,30 @@
...
+      doFlat: False
+      doApplyGains: False
+      doFringe: False
+  cpCombine:
+    class: lsst.cp.pipe.cpFunctions.CalibCombineTask
+    config:
+      connections.inputExps: 'cpBiasProc'
+      connections.outputData: 'biasProposal'
+      calibrationType: 'bias'
+      exposureScaling: "ExposureTime"
```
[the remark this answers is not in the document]
czwa: This was relying on default settings.  I've fixed this, made the scaling "None", and confirmed that this applies a scale factor of 1.0.

pipelines/cpDark.yaml
```diff
@@ -0,0 +1,36 @@
+description: cp_pipe DARK calibration construction
+tasks:
+  isr:
+    class: lsst.ip.isr.isrTask.IsrTask
+    config:
+      connections.ccdExposure: 'raw'
+      connections.outputExposure: 'cpDarkISR'
```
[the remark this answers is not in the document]
czwa: I don't know that there's any strong guidance on dataset names in gen3, but I've taken the suggestion here and below.

python/lsst/cp/pipe/cpDarkTask.py
```diff
@@ -0,0 +1,128 @@
...
+            CR rejected, ISR processed Dark Frame."
+        """
+        psf = measAlg.SingleGaussianPsf(self.config.psfSize,
+                                        self.config.psfSize,
+                                        self.config.psfFwhm/(2*math.sqrt(2*math.log(2))))
+        inputExp.setPsf(psf)
+        scaleExp = inputExp.clone()
+        mi = scaleExp.getMaskedImage()
+
+        # Darktime scaling necessary for repair.run() to ID CRs correctly.
```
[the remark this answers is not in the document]
czwa: Added comment to DM-23680, for further investigation.  I agree that it shouldn't be necessary, but removing this removed zero CRs during testing.

python/lsst/cp/pipe/cpFlatNormTask.py
```diff
@@ -0,0 +1,166 @@
...
+    doUnApplyGains = pexConfig.Field(
+        dtype=bool,
+        doc="Divide by gain to restore that component to the flat.",
+        default=True,
+    )
```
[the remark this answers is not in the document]
czwa: Removed.

python/lsst/cp/pipe/cpFlatNormTask.py
```diff
@@ -0,0 +1,166 @@
...
+        Parameters
+        ----------
+        inputExps : `list` [`lsst.afw.image.Exposure`]
+            Unnormalized merged FLATs to rescale.
+        inputDims : `list` [`dict`]
+            Dictionary of input data dimensions/values.
+
+        Returns
+        -------
+        outputScales : `dict` [`dict` [`float`]]
```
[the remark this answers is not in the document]
czwa: I have updated the docstring to note the types of the keys, and to add the additional forgotten key on amplifier.

python/lsst/cp/pipe/cpFlatNormTask.py
```diff
@@ -0,0 +1,166 @@
...
+        # are caused by differences in lamp brightness and vignetting
+        # (ignoring sky/screen illumination issues we can't fix).
+        # Find the incident flux per exposure by averaging the
+        # amp-level measurement divided by the gain.
+        visitLevel = dict()
+        count = dict()
+
+        for inMetadata, inDimensions in zip(inputExps, inputDims):
+            detectorId = inDimensions.get('detector', None)
+            visitId = inDimensions.get('visit', None)
```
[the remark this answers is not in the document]
czwa: Changed to raise.

python/lsst/cp/pipe/cpFunctions.py
```diff
@@ -0,0 +1,682 @@
...
+    """Task to combine calib exposures."""
+    ConfigClass = CalibCombineConfig
+    _DefaultName = 'cpCombine'
+
+    def __init__(self, **kwargs):
+        super().__init__(**kwargs)
+        self.makeSubtask("stats")
+
+    def runQuantum(self, butlerQC, inputRefs, outputRefs):
+        inputs = butlerQC.get(inputRefs)
```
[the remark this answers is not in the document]
czwa: DM-22521 will implement partial image reads to avoid memory issues.  The short summary is that the inputs will be loaded as "deferred" objects, and the combination will iterate over subregions, loading only the active subregion from each deferred image.  I think this has been tested with regular image coadds.

python/lsst/cp/pipe/cpFlatNormTask.py
```diff
@@ -74,93 +165,207 @@ class CpFlatNormalizationTask(pipeBase.PipelineTask,
...
+            nAmp = len(camera[0])
+            bgMatrix = np.zeros((len(detSet) * nAmp, len(expSet)))
+            bgCounts = np.ones((len(detSet) * nAmp, len(expSet)))
+
+        for inMetadata, inDimensions in zip(inputMDs, inputDims):
+            try:
+                exposureId = inDimensions['exposure']
+                detectorId = inDimensions['detector']
+            except Exception as e:
+                raise KeyError("Cannot find expected dimensions in %s %s" % (inDimensions, e))
```
PaulPrice, line 224: `raise KeyError("...") from e`

python/lsst/cp/pipe/cpFlatNormTask.py
```diff
@@ -74,93 +165,207 @@ class CpFlatNormalizationTask(pipeBase.PipelineTask,
...
+        expSet = sorted(set([d['exposure'] for d in inputDims]))
+        detSet = sorted(set([d['detector'] for d in inputDims]))
+
+        if self.config.level == 'DETECTOR':
+            bgMatrix = np.zeros((len(detSet), len(expSet)))
+            bgCounts = np.ones((len(detSet), len(expSet)))
+        elif self.config.level == 'AMP':
+            nAmp = len(camera[0])
+            bgMatrix = np.zeros((len(detSet) * nAmp, len(expSet)))
+            bgCounts = np.ones((len(detSet) * nAmp, len(expSet)))
```
PaulPrice, line 217: There are lots of `len(...)` in the above block. Suggest pulling them out into their own variables, `numDet` and `numExp`.

python/lsst/cp/pipe/cpFlatNormTask.py
```diff
@@ -74,93 +165,207 @@ class CpFlatNormalizationTask(pipeBase.PipelineTask,
...
+                raise KeyError("Cannot find expected dimensions in %s %s" % (inDimensions, e))
+
+            if self.config.level == 'DETECTOR':
+                detId = detSet.index(detectorId)
+                expId = expSet.index(exposureId)
+                try:
+                    value = inMetadata.get('DETECTOR_MEDIAN')
+                    count = inMetadata.get('DETECTOR_N')
+                except Exception as e:
+                    raise KeyError("Cannot read expected metadata string. %s" (e, ))
```
PaulPrice, line 233: `raise ... from e`.

python/lsst/cp/pipe/cpFlatNormTask.py
```diff
@@ -74,93 +165,207 @@ class CpFlatNormalizationTask(pipeBase.PipelineTask,
...
+
+        for inMetadata, inDimensions in zip(inputMDs, inputDims):
+            try:
+                exposureId = inDimensions['exposure']
+                detectorId = inDimensions['detector']
+            except Exception as e:
+                raise KeyError("Cannot find expected dimensions in %s %s" % (inDimensions, e))
+
+            if self.config.level == 'DETECTOR':
+                detId = detSet.index(detectorId)
```
PaulPrice, line 227: Is this efficient? I think it would be better to get this from a lookup table mapping `detectorId` to `detId`. Ditto for the same pattern used in multiple places below.

python/lsst/cp/pipe/cpFlatNormTask.py
```diff
@@ -74,93 +165,207 @@ class CpFlatNormalizationTask(pipeBase.PipelineTask,
...
+
+                detId = detSet.index(detectorId) * nAmp
+                expId = expSet.index(exposureId)
+
+                for ampIdx, amp in enumerate(detector):
+                    try:
+                        value = inMetadata.get(f'AMP_MEDIAN_{ampIdx}')
+                        count = inMetadata.get(f'AMP_N_{ampIdx}')
+                    except Exception as e:
+                        raise KeyError("cannot read expected metadata string. %s" (e, ))
```
PaulPrice, line 254: `raise .. from e`

python/lsst/cp/pipe/cpFlatNormTask.py
```diff
@@ -74,93 +165,207 @@ class CpFlatNormalizationTask(pipeBase.PipelineTask,
...
         return pipeBase.Struct(
             outputScales=outputScales,
         )
+
+    def measureScales(self, bgMatrix, bgCounts, iterations=10):
+        """Convert backgrounds to exposure and detector components.
+
+        Parameters
+        ----------
+        bgMatrix : `np.ndarray`
```
PaulPrice, line 291: Specify shape for each array, so we can see what's expected.

python/lsst/cp/pipe/cpFlatNormTask.py
```diff
@@ -74,93 +165,207 @@ class CpFlatNormalizationTask(pipeBase.PipelineTask,
...
+            Input backgrounds indexed by exposure (axis=0) and
+            detector (axis=1).
+        bgCounts : `np.ndarray`
+            Input pixel counts used to in measuring bgMatrix, indexed
+            identically.
+        iterations : `int`
+            Number of iterations to use in decomposition.
```
PaulPrice, line 298: Surely this is a configuration parameter?

python/lsst/cp/pipe/cpFlatNormTask.py
```diff
@@ -74,93 +165,207 @@ class CpFlatNormalizationTask(pipeBase.PipelineTask,
...
+
+        Returns
+        -------
+        vectorE : `np.array`
+            Output E vector of exposure level scalings.
+        vectorG : `np.array`
+            Output G vector of detector level scalings.
+        bgModel : `np.ndarray`
+            Expected model bgMatrix values, calculated from E and G.
```
PaulPrice, line 307: Return these inside a `Struct`.

python/lsst/cp/pipe/cpFlatNormTask.py
```diff
@@ -74,93 +165,207 @@ class CpFlatNormalizationTask(pipeBase.PipelineTask,
...
+
+    def measureScales(self, bgMatrix, bgCounts, iterations=10):
+        """Convert backgrounds to exposure and detector components.
+
+        Parameters
+        ----------
+        bgMatrix : `np.ndarray`
+            Input backgrounds indexed by exposure (axis=0) and
+            detector (axis=1).
+        bgCounts : `np.ndarray`
```
PaulPrice, line 294: Make optional?

python/lsst/cp/pipe/cpFlatNormTask.py
```diff
@@ -74,93 +165,207 @@ class CpFlatNormalizationTask(pipeBase.PipelineTask,
...
+                logG[bad] = logG[~bad].mean()
+
+            logE = np.array([np.average(logMeas[:, iexp] - logG,
+                                        weights=bgCounts[:, iexp]) for iexp in range(numExps)])
+            fluxLevel = np.average(np.exp(logG), weights=np.sum(bgCounts, axis=1))
+
+            logG -= np.log(fluxLevel)
+            self.log.info(f"ITER {iter}: Flux: {fluxLevel}")
+            self.log.info(f"Exps: {np.exp(logE)}")
+            self.log.info(f"{np.mean(logG)}")
```
PaulPrice, line 363: I think these should be `debug` level.

python/lsst/cp/pipe/cpFlatNormTask.py
```diff
@@ -74,93 +165,207 @@ class CpFlatNormalizationTask(pipeBase.PipelineTask,
...
+
+        logE = np.array([np.average(logMeas[:, iexp] - logG,
+                                    weights=bgCounts[:, iexp]) for iexp in range(numExps)])
+        bgModel = np.zeros((numChips, numExps))
+        for ichip in range(numChips):
+            for iexp in range(numExps):
+                bgModel[ichip][iexp] = np.exp(logE[iexp] - logG[ichip])
```
PaulPrice, line 370: Iterations are inefficient in python. Can't you write something like `bgModel = np.exp(logE[:, np.newaxis] - logG[np.newaxis, :])` (untested!).

python/lsst/cp/pipe/cpFlatNormTask.py
```diff
@@ -74,93 +165,207 @@ class CpFlatNormalizationTask(pipeBase.PipelineTask,
...
+        different exposures.  For a perfect lamp of unit total
+        intensity, this is simply the sum along the rows of B
+        (np.sum(B, axis=1)).
+
+        The algorithm here, from pipe_drivers/constructCalibs.py and
+        from there from Eugene Magnier/PanSTARRS [1]_, attempts to
+        iteratively solve this decomposition from initial "perfect" E
+        and G vectors.  The operation is performed in log space to
+        reduce the multiply and divides to linear additions and
+        subtractions.
```
PaulPrice, line 332: I think it's important to mention that the average of the G vector is constrained to unity by the algorithm (i.e., we're solving for the _relative_ gains, not absolute values).

python/lsst/cp/pipe/cpCombine.py
```diff
@@ -172,10 +172,10 @@ class CalibCombineConfig(pipeBase.PipelineTaskConfig,
     scalingLevel = pexConfig.ChoiceField(
         dtype=str,
         allowed={
-            "Amplifier": "Scale by amplifier.",
-            "Detector": "Scale by detector.",
+            "DETECTOR": "Scale by detector.",
+            "AMP": "Scale by amplifier.",
```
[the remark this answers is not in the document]
czwa: I tried to match what already exists elsewhere in the stack: https://github.com/lsst/ip_isr/blob/master/python/lsst/ip/isr/isrTask.py#L522

python/lsst/cp/pipe/cpFlatNormTask.py
czwa: I had to switch the axes, but I think that does do the right thing.

PR #21, opened 2019-12-05
Rewrite MeasurePhotonTransferCurveTask datatypes

mfisherlevine:


python/lsst/cp/pipe/ptc.py
```diff
@@ -76,7 +78,7 @@ class MeasurePhotonTransferCurveTaskConfig(pexConfig.Config):
     )
     makePlots = pexConfig.Field(
         dtype=bool,
-        doc="Plot the PTC curves?.",
+        doc="Plot the PTC curves?",
```
plazas: This is a trivial comment, but I thought that docstrings needed to be full sentences that ended with periods. Why did you delete the period?

python/lsst/cp/pipe/ptc.py
```diff
@@ -108,6 +110,12 @@ class MeasurePhotonTransferCurveTaskConfig(pexConfig.Config):
         doc="Maximum value to of mean signal (in ADU) to consider.",
         default=9e6,
     )
+    initialNonLinearityExclusionThreshold = pexConfig.Field(
+        dtype=float,
+        doc="Initially exclude data points that are more than a factor of this from being linear from"
```
plazas: Linear in the var vs mean curve, correct? Maybe a clarification is in order since the code also calculates linearity residuals from "mean vs time".

python/lsst/cp/pipe/ptc.py
```diff
@@ -108,6 +110,12 @@ class MeasurePhotonTransferCurveTaskConfig(pexConfig.Config):
...
+        " the iterative fit.",
+        default=0.25,
```
plazas: Why was the default value chosen? Trial and error with the data set that you used to test the code?

python/lsst/cp/pipe/ptc.py
```diff
@@ -432,85 +485,141 @@ def funcAstier(self, pars, x):
...
+    def _getInitialGoodPoints(means, variances, maxDeviation):
+        """Return a boolean array to mask bad points.
+
+        A linear function has a constant ratio, so find the median
+        value of the ratios, and exclude the points that deviate
+        from that by more than a factor of maxDeviation.
+
+        Too high and points that are so bad that fit will fail will be included
+        Too low and the non-linear points will be excluded, biasing the NL fit."""
+        ratios = [b/a for (a, b) in zip(means, variances)]
```
plazas: Should you guard against division by zero? (with an error message or something)

python/lsst/cp/pipe/ptc.py
```diff
@@ -432,85 +485,141 @@ def funcAstier(self, pars, x):
...
+        medianRatio = np.median(ratios)
+        ratioDeviations = [r/medianRatio for r in ratios]
+        goodPoints = [abs(1-r) < maxDeviation for r in ratioDeviations]
+        return np.array(goodPoints)
+
+    def fitPtcAndNl(self, dataset, ptcFitType):
+        """Fit the photon transfer curve and calculate linearity and residuals.
+
+        Fit the photon transfer curve with either a polynomial of order
+        specified in the task config, or using the Astier approximation.
```
plazas: "...the Astier approximation (Eq. 16 of 1905.08677)".  Their paper has a full model (Eq. 20) which has an approximation (Eq. 15). This one, in turn, has another approximation (Eq. 16), which is the one we are using for now.

python/lsst/cp/pipe/ptc.py
```diff
@@ -432,85 +485,141 @@ def funcAstier(self, pars, x):
...
+    def _initialParsForPolynomial(order):
+        assert(order >= 2)
+        pars = np.zeros(order, dtype=np.float)
+        pars[0] = 10
+        pars[1] = 1
+        pars[2:] = 0.0001
+        return pars
+
+    @staticmethod
+    def _boundsForPolynomial(initialPars):
```
plazas: This function, as it is defined, seems to always return bounds that are independent on the argument "initialPars", namely [0, -inf, -inf, ..., -inf] and [inf, inf, ... , inf]. Something similar happens for the function below ("_boundsForAstier"). Is this intentional or maybe I am misinterpreting things?

python/lsst/cp/pipe/ptc.py
```diff
@@ -432,85 +485,141 @@ def funcAstier(self, pars, x):
...
-            index0 = ((meanVecOriginal > self.config.minMeanSignal) &
-                      (meanVecOriginal <= self.config.maxMeanSignal))
-            #  Before bootstrap fit, do an iterative fit to get rid of outliers in PTC
+
+        for ampName in dataset.ampNames:
+            timeVecOriginal = np.array(dataset.rawExpTimes[ampName])
+            meanVecOriginal = np.array(dataset.rawMeans[ampName])
+            varVecOriginal = np.array(dataset.rawVars[ampName])
+
+            mask = ((meanVecOriginal > self.config.minMeanSignal) &
```
plazas: Should it be ">=" self.config.minMeanSignal too? In either case, we should update the docstring for these two parameters ("self.config.minMeanSignal " and "self.config.maxMeanSignal") to say explicitly if the value of the parameter is included in the cut or not (sorry I did not do this originally).

python/lsst/cp/pipe/ptc.py
timj: A question mark takes the role of a period. There should never be a period after a question mark.

mfisherlevine: Thanks for the English explanation, that is indeed the reason. There is loads of prior art where boolean config options which contain an imperative verb (`do*`, `make*` _etc_) have docstrings ending in a question mark, so will leave this as-is.

python/lsst/cp/pipe/ptc.py
mfisherlevine: Well, the mask is applied to both if you read the code; this initial cut is indeed just to the PTC code, but points that are thrown out of the PTC fit are also excluded from the non-linearity fit. I've updated the docstring to make that as clear as possible though.

python/lsst/cp/pipe/ptc.py
mfisherlevine: The docstring for the function applying this config parameter explains it more fully, and I think that's the right place to have longer explanations. If we're guarding against negative we should also enforce the range zero to one, but I think if people are doing things that wrong we've got bigger problems 🙂

mfisherlevine: Or rather, and less flippantly, config options are there to be changed by people who know what they're doing, and with good reason, and most places in the code don't bother checking too carefully whether someone's done something really stupid there.

mfisherlevine: OK, fine, in writing that I convinced myself that's not ideal, so I'll change this to a `pexConfig.RangeField`!

python/lsst/cp/pipe/ptc.py
mfisherlevine: Hmmm... that would require the variance to be exactly zero. For that to be the the case, all the data must be bitwise identical, as this is the empirical variance, not the mean of a  (potentially non-existent) variance plane. I think if that's the case, this is a good a place as any for the user to find out that their data is extremely messed up. I think that adding a  check to values _before_ this function isn't a bad idea, but that should be work to add sanity checks to this whole task, and probably belongs on a new tickets (or be added in general when tasks are written).

python/lsst/cp/pipe/ptc.py
mfisherlevine: I was just improving the docstring that you originally wrote here! 😄 I think that info is already in the main class docstring though, and it's clear here that we're talking about the same function. Maybe that one could be improved too though, with this information. If you think so, could you suggest exactly how you'd like the main class docstring to read, and I'll update that instead?

python/lsst/cp/pipe/ptc.py
mfisherlevine: It takes `initialPars` as an argument because it needs to know how many there are, in order to return the right number of bounds (see it used in the list comps). The values are all plus/minus `inf` for the Astier fit, because they were previously unbounded, and I was just keeping them the same, whilst also adding both a place-holder method for adding bounds, should we want them in the future, whilst also not hard-coding the number of bounds into the code, especially as this is different for polynomial vs. Astier.

python/lsst/cp/pipe/ptc.py
mfisherlevine: I don't know, I didn't change the logic in this block from your original code. The only reason this line appears in this PR is because the name of the variable it is assigned to changed from `index0` to `mask`.

mfisherlevine: I've changed the logic to be inclusive for both bounds now, and updated the docstrings to  say so though.

python/lsst/cp/pipe/ptc.py
plazas: OK, since it is already in the main class docstring, then it is fine. The paper and the equation are also already cited there.

python/lsst/cp/pipe/ptc.py
mfisherlevine: Oh, I forgot, I also update the docstring on the config param to read:
```        doc="Initially exclude data points with a variance that are more than a factor of this from being"
            " linear, from the PTC fit. Note that these points will also be excluded from the non-linearity"
            " fit. This is done before the iterative outlier rejection, to allow an accurate determination"
            " of the sigmas for said iterative fit.",```

PR #23, opened 2020-01-14
Fix PTC fit type

mfisherlevine:


python/lsst/cp/pipe/makeBrighterFatterKernel.py
```diff
@@ -529,6 +529,8 @@ def runDataRef(self, dataRef, visitPairs):
             ptcConfig = MeasurePhotonTransferCurveTaskConfig()
             ptcConfig.isrForbiddenSteps = []
             ptcConfig.doFitBootstrap = True
+            ptcConfig.ptcFitType = 'POLYNOMIAL'  # default Astier doesn't work for gain correction
```
czwa: I thought POLYNOMIAL was the default ptcFitType?

mfisherlevine: Aaaargh, you're right, and therefore this doesn't fix the problem! 😩 Something is overriding that, need to hunt. Sorry for the premature review... 😔

PR #25, opened 2020-02-13
Update defect finding task to add metadata

mfisherlevine:
Update the defect finding task to add all metadata necessary for ingest task to use the newly generated defects.

### DIFF
--- a/python/lsst/cp/pipe/defects.py
+++ b/python/lsst/cp/pipe/defects.py
@@ -36,6 +36,7 @@ import lsst.afw.detection as afwDetection
 import lsst.afw.display as afwDisplay
 from lsst.afw import cameraGeom
 from lsst.geom import Box2I, Point2I
+import lsst.daf.base as dafBase
 
 from lsst.ip.isr import IsrTask
 from .utils import NonexistentDatasetTaskDataIdContainer, SingleVisitListTaskRunner, countMaskedPixels, \
@@ -159,6 +160,19 @@ class FindDefectsTaskConfig(pexConfig.Config):
         doc=("Ensure that all visits are from the same run? Raises if this is not the case, or"
              "if the run key isn't found."),
         default=False,  # false because most obs_packages don't have runs. obs_lsst/ts8 overrides this.
+    )
+    ignoreFilters = pexConfig.Field(
+        dtype=bool,
+        doc=("Set the filters used in the CALIB_ID to NONE regardless of the filters on the input"
+             " images. Allows mixing of filters in the input flats. Set to False if you think"
+             " your defects might be chromatic and want to have registry support for varying"
+             " defects with respect to filter."),
+        default=True,
+    )
+    nullFilterName = pexConfig.Field(
+        dtype=str,
+        doc=("The name of the null filter if ignoreFilters is True. Usually something like NONE or EMPTY"),
+        default="NONE",
     )
     combinationMode = pexConfig.ChoiceField(
         doc="Which types of defects to identify",
@@ -295,6 +309,9 @@ class FindDefectsTask(pipeBase.CmdLineTask):
 
         defectLists = {'dark': [], 'flat': []}
 
+        midTime = 0
+        filters = set()
+
         if self.config.mode == 'MASTER':
             if len(visitList) > 1:
                 raise RuntimeError(f"Must only specify one visit when using mode MASTER, got {visitList}")
@@ -302,6 +319,8 @@ class FindDefectsTask(pipeBase.CmdLineTask):
 
             for datasetType in defectLists.keys():
                 exp = dataRef.get(datasetType)
+                midTime += self._getMjd(exp)
+                filters.add(exp.getFilter().getName())
                 defects = self.findHotAndColdPixels(exp, datasetType)
 
                 msg = "Found %s defects containing %s pixels in master %s"
@@ -310,6 +329,7 @@ class FindDefectsTask(pipeBase.CmdLineTask):
                 if self.config.makePlots:
                     self._plot(dataRef, exp, visitList[0], self._getNsigmaForPlot(datasetType),
                                defects, datasetType)
+            midTime /= len(defectLists.keys())
 
         elif self.config.mode == 'VISITS':
             butler = dataRef.getButler()
@@ -328,11 +348,15 @@ class FindDefectsTask(pipeBase.CmdLineTask):
                     exp = self.isrForFlats.runDataRef(dataRef).exposure
                     defects = self.findHotAndColdPixels(exp, imageType)
                     defectLists['flat'].append(defects)
+                    midTime += self._getMjd(exp)
+                    filters.add(exp.getFilter().getName())
 
                 elif imageType == 'dark':
                     exp = self.isrForDarks.runDataRef(dataRef).exposure
                     defects = self.findHotAndColdPixels(exp, imageType)
                     defectLists['dark'].append(defects)
+                    midTime += self._getMjd(exp)
+                    filters.add(exp.getFilter().getName())
 
                 else:
                     raise RuntimeError(f"Failed on imageType {imageType}. Only flats and darks supported")
@@ -342,6 +366,8 @@ class FindDefectsTask(pipeBase.CmdLineTask):
 
                 if self.config.makePlots:
                     self._plot(dataRef, exp, visit, self._getNsigmaForPlot(imageType), defects, imageType)
+
+            midTime /= len(visitList)
 
         msg = "Combining %s defect sets from darks for detector %s"
         self.log.info(msg, len(defectLists['dark']), detNum)
@@ -357,7 +383,7 @@ class FindDefectsTask(pipeBase.CmdLineTask):
         brightDarkPostMerge = [mergedDefectsFromDarks, mergedDefectsFromFlats]
         allDefects = self._postProcessDefectSets(brightDarkPostMerge, exp.getDimensions(), mode='OR')
 
-        self._writeData(dataRef, allDefects)
+        self._writeData(dataRef, allDefects, midTime, filters)
 
         self.log.info("Finished finding defects in detector %s" % detNum)
         return pipeBase.Struct(defects=allDefects, exitStatus=0)
@@ -375,7 +401,7 @@ class FindDefectsTask(pipeBase.CmdLineTask):
             nPix += d.getBBox().getArea()
         return nPix
 
-    def _writeData(self, dataRef, defects):
+    def _writeData(self, dataRef, defects, midTime, filters):
         """Write the data out to the defect file.
 
         Parameters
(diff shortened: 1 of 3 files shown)


python/lsst/cp/pipe/defects.py
```diff
@@ -367,27 +393,91 @@ def _writeData(self, dataRef, defects):
...
+        now = dafBase.DateTime.now().toPython()
+        mdOriginal = defects.getMetadata()
+        mdSupplemental = {"INSTRUME": instrumentName,
+                          "DETECTOR": dataRef.dataId['detector'],
+                          "CALIBDATE": date,
+                          "CALIB_ID": CALIB_ID,
+                          "CALIB_CREATION_DATE": now.date().isoformat(),
+                          "CALIB_CREATION_TIME": now.time().isoformat()}
+
+        for k, v in mdSupplemental.items():
```
timj: These two lines can be rewritten as:
```python
mdOriginal.update(mdSupplemental)
```

python/lsst/cp/pipe/defects.py
```diff
@@ -367,27 +393,91 @@ def _writeData(self, dataRef, defects):
...
+            mdOriginal.add(k, v)
+        defects.setMetadata(mdOriginal)
```
timj: I don't think this is needed because you used `getMetadata` above.

python/lsst/cp/pipe/defects.py
```diff
@@ -142,6 +143,19 @@ class FindDefectsTaskConfig(pexConfig.Config):
              "if the run key isn't found."),
         default=False,  # false because most obs_packages don't have runs. obs_lsst/ts8 overrides this.
     )
+    ignoreFilters = pexConfig.Field(
+        dtype=bool,
+        doc=("Set the filters used in the CALIB_ID to NONE regardless of the filters on the input"
+             " images. Allows mixing of filters in the input flats. Set to True if you think"
```
czwa: If the default is True, then this docstring should note that the user should "Set to False if you think your defects are not chromatic" etc.  My actual suggestion is to make the default False, because I would prefer not to think about chromatic defects.

python/lsst/cp/pipe/defects.py
```diff
@@ -277,13 +291,18 @@ def runDataRef(self, dataRef, visitList):
...
         defectLists = {'dark': [], 'flat': []}
 
+        midTime = 0
+        filters = set()
+
         if self.config.mode == 'MASTER':
             if len(visitList) > 1:
                 raise RuntimeError(f"Must only specify one visit when using mode MASTER, got {visitList}")
-            dataRef.dataId['visit'] = visitList[0]
+            dataRef.dataId['expId'] = visitList[0]
```
czwa: Should visitList become exposureList, to reduce possible confusion?

czwa: Requisite comment that processing should happen in a Task.run() method, and not the Task.runDataRef() method.

python/lsst/cp/pipe/defects.py
```diff
@@ -367,27 +393,91 @@ def _writeData(self, dataRef, defects):
         defects : `lsst.meas.algorithms.Defect`
             The defects to be written.
         """
-        filename = dataRef.getUri(write=True)  # does not guarantee that full path exists
-        dirname = os.path.dirname(filename)
-        if not os.path.exists(dirname):
-            os.makedirs(dirname)
+        date = str(dafBase.DateTime(midTime, dafBase.DateTime.MJD).toPython())
```
timj: does:
```python
date = dafBase.DateTime(midTime, dafBase.DateTime.MJD).toPython().isoformat()
```
work so you don't need the next line?

python/lsst/cp/pipe/defects.py
```diff
@@ -367,27 +393,91 @@ def _writeData(self, dataRef, defects):
...
+        date = date.replace(" ", "T")  # original has a space in the middle
```
czwa: Can this be done as `date = dafBase.DateTime(midTime, dafBase.DateTime.MJD).toPython().isoformat()`, so the formatting options are done on the datetime object instead of messing with the string?

python/lsst/cp/pipe/defects.py
```diff
@@ -170,11 +184,11 @@ class FindDefectsTaskConfig(pexConfig.Config):
     writeAs = pexConfig.ChoiceField(
         doc="Write the output file as ASCII or FITS table",
         dtype=str,
-        default="FITS",
+        default="ECSV",
         allowed={
-            "ASCII": "Write the output as an ASCII file",
+            "ECSV": "Write the output as an ECSV file",
```
timj: The api is `writeText` since I deliberately avoided making the underlying text format something people needed to care about.

python/lsst/cp/pipe/defects.py
```diff
@@ -367,27 +393,91 @@ def _writeData(self, dataRef, defects):
...
+        instrumentName = self._getInstrumentName(dataRef)
+        detNum = self._getDetectorNumber(dataRef)
+        if not self.config.ignoreFilters:
+            filt = self._filterSetToFilterString(filters)
+        else:
+            filt = self.config.nullFilterName
+
+        CALIB_ID = f"detectorName={detName} detector={detNum} calibDate={date} ccd={detNum} filter={filt}"
+        try:
+            raftName = detName.split("_")[0]
```
timj: Do all instruments include the raft name in detName but also need a separate raftName?

python/lsst/cp/pipe/defects.py
```diff
@@ -367,27 +393,91 @@ def _writeData(self, dataRef, defects):
...
+                          "CALIBDATE": date,
+                          "CALIB_ID": CALIB_ID,
+                          "CALIB_CREATION_DATE": now.date().isoformat(),
+                          "CALIB_CREATION_TIME": now.time().isoformat()}
+
+        for k, v in mdSupplemental.items():
+            mdOriginal.add(k, v)
+        defects.setMetadata(mdOriginal)
+
+        templateFilename = dataRef.getUri(write=True)  # does not guarantee that full path exists
```
timj: This all seems wrong if butler is involved. Why can't butler.put work directly? Why do you have to ask butler for a template at all?

python/lsst/cp/pipe/defects.py
```diff
@@ -367,27 +393,91 @@ def _writeData(self, dataRef, defects):
...
+        baseDirName = os.path.dirname(templateFilename)
+        # ingest curated calibs demands detectorName is lowercase
+        dirName = os.path.join(baseDirName, instrumentName, "defects", detName.lower())
```
timj: @SimonKrughoff I'm not sure I follow why this is needed. @mfisherlevine are you trying to write the output file directly to obs_lsst_data?

python/lsst/cp/pipe/defects.py
mfisherlevine: Yeah, and the rest! There are currently 75 other instances of the string `visit` in that file alone. I was concentrating on the code for now, because I was in a hurry, but definitely do agree that ideally all instances would be changed...

How about we ticket doing that once Andrés' addition of tests for defects merge, so that 1) we have better test coverage when renaming variables, and 2) I don't give him a rebase nightmare to deal with by changing 75 instances on him?

mfisherlevine: Whilst appropriate, 1) this change is not relevant, and 2) it's hard for `cp_pipe` stuff operating on tons of visits, as you know. Will ignore this comment for now unless you shout 🙂

python/lsst/cp/pipe/defects.py
mfisherlevine: No (hence the `try`), plus we're actually using `detName` totally wrong at the moment, but I can only try to fix so many things at once 😔

python/lsst/cp/pipe/defects.py
mfisherlevine: Come chat with me and KT as per Slack to resolve this 🙂

python/lsst/cp/pipe/defects.py
mfisherlevine: In short - no. In long, see above request for a chat 🙂

python/lsst/cp/pipe/defects.py
mfisherlevine: Sounds good. What do you propose to call this in order to let people choose between `.writeText()` and `.writeFits()`? "TEXT"? Or leave it as `"ASCII"`?

python/lsst/cp/pipe/defects.py
mfisherlevine: I think so - Chris suggested the same above. Will check and change if so.

python/lsst/cp/pipe/defects.py
mfisherlevine: Ah, thanks!

python/lsst/cp/pipe/defects.py
mfisherlevine: That is indeed much nicer 👍

python/lsst/cp/pipe/defects.py
timj: TEXT or leave ASCII is fine with me.

python/lsst/cp/pipe/defects.py
czwa: Something I thought of but forgot to ask: does this need to modify detName if a raftName is found, or is something like `detName='R00_S00', raftName='R00'` expected?

python/lsst/cp/pipe/defects.py
mfisherlevine: I agree with the first part, and for the default to be not supporting chromatic defects, but I therefore think you've for a sign flip. It now reads:
`    ignoreFilters = pexConfig.Field(
        dtype=bool,
        doc=("Set the filters used in the CALIB_ID to NONE regardless of the filters on the input"
             " images. Allows mixing of filters in the input flats. Set to False if you think"
             " your defects might be chromatic and want to have registry support for varying"
             " defects with respect to filter."),
        default=True,`

mfisherlevine: I think that is correct now. Shout if I'm mistaken.

python/lsst/cp/pipe/defects.py
mfisherlevine: Yup - done. Much nicer, thanks.

python/lsst/cp/pipe/defects.py
mfisherlevine: Cool, just dropped that whole commit 👍

python/lsst/cp/pipe/defects.py
mfisherlevine: It's not expected, as such, but that is indeed exactly the "using `detName` totally wrong" that I refer to above. `detName` should only _ever_ be `S12`-like, but that's not how it is right now 😔

python/lsst/cp/pipe/defects.py
mfisherlevine: Discussion OOB has resolved this, I think. I am putting in a TODO to sort this, and the comment below, out in Gen3 and leaving as is for now.

PR #26, opened 2020-02-21
DM-23195: DM-21221 broke cp_pipe due to lack of tests

plazas:


python/lsst/cp/pipe/ptc.py
```diff
@@ -705,6 +705,11 @@ def fitPtcAndNonLinearity(self, dataset, tableArray, ptcFitType):
             'ASTIERAPPROXIMATION' to the PTC
         tableArray : `np.array`
             Look-up table array with size rows=nAmps and columns=ADU values
+
+        Returns
+        -------
+        dataset : `lsst.cp.pipe.ptc.PhotonTransferCurveDataset`
```
jdswinbank: This `dataset` is the same object as the `dataset` parameter, but has been modified (right?). Let's document that — rather than just repeating the same words, let's be explicit about what this method changes in that object.

jdswinbank:
There are actually three separate changes here, and I think it would be helpful to split them into three separate commits:

- Change from `visit` to `expId`. The commit message for this should explain why it was necessary (e.g. by referring to the ticket which changed definitions elsewhere).
- Correct the method call from `fitPtcAndNl` to `fitPtcAndNonLinearity`.
- Adjust the definition of `fitPtcAndNonLinearity` to return `dataset`.

jdswinbank:
Also, remember the character limit on the first line of commit messages: see our [guidelines](https://developer.lsst.io/work/flow.html#git-commit-message-best-practices).

python/lsst/cp/pipe/ptc.py
```diff
@@ -354,7 +354,7 @@ def runDataRef(self, dataRef, visitPairs):
         lookupTableArray = np.zeros((numberAmps, numberAduValues), dtype=np.float32)
 
         # Fit PTC and (non)linearity of signal vs time curve, produce linearizer
-        self.fitPtcAndNonLinearity(dataset, lookupTableArray, ptcFitType=self.config.ptcFitType)
+        dataset = self.fitPtcAndNonLinearity(dataset, lookupTableArray, ptcFitType=self.config.ptcFitType)
```
jdswinbank: It might be helpful to clarify the comment before this line. Given the comment “Fit..., produce linearizer”, I would expect this to return something called `linearizer`, not `dataset`.

python/lsst/cp/pipe/ptc.py
plazas: I have expanded the docstring with clarifications.

python/lsst/cp/pipe/ptc.py
plazas: I have added a couple of extra comments in there.

plazas:
Thank you for your comments. I have addressed them.

PR #27, opened 2020-03-16
DM-23681: Check, in the unit tests, that fitPtcAndNonLinearity returns what it is supposed to return.

plazas:


tests/test_ptc.py
```diff
@@ -95,50 +99,94 @@ def setUp(self):
...
-        for ampName in self.ampNames:
-            self.assertAlmostEqual(self.gain, localDataset.gain[ampName])
-            self.assertAlmostEqual(np.sqrt(self.noiseSq)*self.gain, localDataset.noise[ampName])
-            # Linearity residual should be zero
-            self.assertTrue(localDataset.nonLinearityResiduals[ampName].all() == 0)
+    def test_ptcFitQuadAndCubic(self):
+        for order in [2, 3]:
+            self.ptcFitQuadAndCubic(order)
 
     def test_ptcFitAstier(self):
```
jdswinbank: Almost all of the changes in this function are a duplicate of those in `ptcFitQuadAndCubic`. Could you take this opportunity to combine them into one? It'd be a lot easier to see what was going on, and to make sure they are updated in sync in future!

PR #33, opened 2020-04-22
DM-24532: findDefects.py is taking a unusually long time to run (w_2020_16)

plazas:


python/lsst/cp/pipe/defects.py
```diff
@@ -731,12 +731,10 @@ def _markBlocksInBadColumn(self, x, y, multipleX, defects):
                 assert len(limits)%2 == 0, 'limits is even by design, but check anyways'
                 for i in np.arange(0, len(limits)-1, 2):
                     s = Box2I(minimum=Point2I(x0, limits[i]), maximum=Point2I(x0, limits[i+1]))
-                    if s not in defects:
-                        defects.append(s)
```
mfisherlevine, line 735: Removing both of these checks looks like it could have adverse effects. Why is this something that should be done, and why is it safe to do exactly?

plazas: For now, removing these speeds things up.  When you and I discussed this, we were concerned about duplicates in the list. I have checked with Tim Jenness that they can be removed for now as a temporary measure (if we end up with duplicate Defect objects in Defects then we will still get the same mask). A new ticket was filed for this: DM-24554

mfisherlevine: It seems like it speeds it up on one place, but presumably creates a fantastically large list somewhere else (if it wasn't really big, it wouldn't be that slow in the first place, right?).

Why is this temporary and not-obviously-safe change necessary, rather than the coming ticker to fix it properly?

timj: If someone wants to write the Defects.normalize method (there is a ticket for that) it would be only a few lines to implement (find the extent of the largest bounding box, make mask, generate new defects from mask) -- then you can run `normalize` when the loop finishes.

plazas:
This ticket was declared as "won't fix" (see https://jira.lsstcorp.org/browse/DM-24532)

PR #35, opened 2020-05-05
DM-24556: Add normalize method to Defects

plazas:


tests/test_defects.py
```diff
@@ -151,6 +151,40 @@ def check_maskBlocks(self, inputDefects, expectedDefects):
         for boxMeas in boxesMeasured:
             self.assertIn(boxMeas, unionInputExpectedBoxes)
 
+    def test_normalize_defects(self):
+        """A test for the lsst.meas.algorithms.Defect.normalize() method.
```
mfisherlevine, line 155: Why does this live in `cp_pipe`? Shouldn't the tests for `lsst.meas.algorithms.Defect` live in `lsst.meas.algorithms`?

tests/test_defects.py
```diff
@@ -151,6 +151,40 @@ def check_maskBlocks(self, inputDefects, expectedDefects):
...
+        # Second series of 1-pixel contiguos defects
+        defects.append(Box2I(corner=Point2I(20, 11), dimensions=Extent2I(1, 1)))
+        defects.append(Box2I(corner=Point2I(20, 12), dimensions=Extent2I(1, 1)))
+        defects.append(Box2I(corner=Point2I(20, 13), dimensions=Extent2I(1, 1)))
+        defects.append(Box2I(corner=Point2I(20, 14), dimensions=Extent2I(1, 1)))
+        defects.append(Box2I(corner=Point2I(20, 15), dimensions=Extent2I(1, 1)))
+
+        # the normalize() method should replace the 10 defects above by just
+        # two defect boxes with adjusted dimensions
+        newDefects = defects.normalize(self.flatExp.getBBox())
```
mfisherlevine, line 174: Yeah, this line shows that my comment on the other PR asking about if this should be a class method is definitely justified. I think, looking at this, it should actually not be a class method, but one that normalises the set of defects itself, but I'm certainly open to the alternative, but I don't think it can be like this.

tests/test_defects.py
```diff
@@ -151,6 +151,40 @@ def check_maskBlocks(self, inputDefects, expectedDefects):
...
+        for defect in newDefects:
+            boxesMeasured.append(defect.getBBox())
+
+        # The normalizing function should have created the following two boxes out
+        # of the individual 1-pixel defects from above
+        expectedDefects = [Box2I(corner=Point2I(15, 1), dimensions=Extent2I(1, 5)),
+                           Box2I(corner=Point2I(20, 11), dimensions=Extent2I(1, 5))]
+
+        for d in expectedDefects:
+            self.assertIn(d, boxesMeasured)
```
mfisherlevine, line 186: This all looks good, but could you check it's these and only these instead of just ensuring these are in there?

tests/test_defects.py
plazas: I'll move it to somewhere in "meas_algorithms/tests/".

tests/test_defects.py
plazas: I'll make it a method that modifies the defects object in place.

plazas: After the comment by Tim in the PR of `meas_algorithms` (that he is fine with the method returning an object), I'll leave the method to return a new defects object.

plazas:
I'm moving these tests to "tests/test_interp.py " in "meas_algorithms". Since this is the only file modified in "cp_pipe", I'm closing this PR.

PR #37, opened 2020-05-12
DM-24756: Make sigma clipping configurable in PTC task

plazas:


python/lsst/cp/pipe/ptc.py
```diff
@@ -145,6 +145,11 @@ class MeasurePhotonTransferCurveTaskConfig(pexConfig.Config):
         doc="Sigma cut for outlier rejection in PTC.",
         default=5.0,
     )
+    sigmaCutAfwMathStatsControl = pexConfig.Field(
```
czwa, line 148: This should probably be called somethingNumSigmaClip, so it's clear what the clipping is being applied to (the "something") and so it matches the name of the property set in the statsControl object.

python/lsst/cp/pipe/ptc.py
```diff
@@ -145,6 +145,11 @@ class MeasurePhotonTransferCurveTaskConfig(pexConfig.Config):
...
+        dtype=float,
+        doc="Sigma cut for afwMatch.StatisticsControl()",
```
czwa, line 150: Typo I think: Match vs Math.

python/lsst/cp/pipe/ptc.py
```diff
@@ -145,6 +145,11 @@ class MeasurePhotonTransferCurveTaskConfig(pexConfig.Config):
...
+        default=3.0,
```
mfisherlevine, line 151: Can we change this to 5.5 please?

RobertLuptonTheGood: On `None` and use the default if not set

python/lsst/cp/pipe/ptc.py
mfisherlevine: `nSigmaClipPtc` would, I think, be in keeping with our namings elsewhere.

python/lsst/cp/pipe/ptc.py
plazas: I changed the name of the parameter to `nSigmaClipPtc`

plazas: ...but I did fix the typo in the docstring, thanks!

PR #39, opened 2020-06-09
DM-20070: Enhance PTC task to use Pierre^2's code

plazas:


python/lsst/cp/pipe/ptc.py
```diff
@@ -367,11 +434,269 @@ def runDataRef(self, dataRef, visitPairs):
...
+            if self.config.makePlots:
+                self.plotCovariancesAstier(dataRef, covFits, covFitsNoB, tupleCovariancesWithTags,
+                                           log=self.log)
+
+            datasetCovAstier = PhotonTransferCurveCovAstierDataset()
+            datasetCovAstier.covariancesTuple = tupleCovariancesWithTags
+            datasetCovAstier.covariancesFits = covFits
+            datasetCovAstier.covariancesFitsWithNoB = covFitsNoB
+
+            dataset = self.getOutputPtcDataCovAstier(covFits)
```
czwa, line 456: Can this generate the thing we want without having to reassign the output values to the PTC dataset?

python/lsst/cp/pipe/ptc.py
```diff
@@ -367,11 +434,269 @@ def runDataRef(self, dataRef, visitPairs):
...
+            datasetCovAstier.var = dataset.var
+            datasetCovAstier.varModel = dataset.varModel
+            datasetCovAstier.meanSignalAdu = dataset.signalAdu
+            datasetCovAstier.varAdu = dataset.varAdu
+            datasetCovAstier.varModelAdu = dataset.varModelAdu
+            datasetCovAstier.a00 = dataset.a00
+            datasetCovAstier.b00 = dataset.b00
+
+            self.log.info(f"Writing Astier+19 covariances data")
+            dataRef.put(datasetCovAstier, datasetType="photonTransferCurveCovAstierDataset")
```
czwa, line 470: Can this method be transformed into a runDataRef()->run() chain, where runDataRef loads all necessary data, passes it to run, which generates the outputs, returns them in a Struct to runDataRef, which writes the outputs?

python/lsst/cp/pipe/ptc.py
```diff
@@ -367,11 +434,269 @@ def runDataRef(self, dataRef, visitPairs):
...
+            self.log.info("Writing linearizers: \n "
+                          "lookup table (linear component of polynomial fit), \n "
+                          "polynomial (coefficients for a polynomial correction), \n "
+                          "and squared linearizer (quadratic coefficient from polynomial)")
+
+            detName = detector.getName()
+            now = datetime.datetime.utcnow()
+            calibDate = now.strftime("%Y-%m-%d")
+
+            for linType, dataType in [("LOOKUPTABLE", 'linearizeLut'),
```
czwa, line 494: Linearizers cannot be generated from the Astier+19 measurement of the covariance?  I thought it was included in the "ptc pipeline" diagram.

python/lsst/cp/pipe/ptc.py
```diff
@@ -367,11 +434,269 @@ def runDataRef(self, dataRef, visitPairs):
...
+        tupleRecords = []
+        allTags = []
+
+        self.log.info('Measuring Astier covariances using %s visits for detector %s' % (visitPairs,
+                                                                                        detector.getId()))
+
+        for (v1, v2) in visitPairs:
+            # Perform ISR on each exposure
+            dataRef.dataId['expId'] = v1
+            exp1 = self.isr.runDataRef(dataRef).exposure
```
czwa, line 606: Is there a benefit for running ISR here, instead of pulling the postISRCCD data from disk after they've been processed with a particular configuration?  I assume the ISR processing doesn't change, so repeating it from scratch every time a different PTC calculation is done seems like it takes up a lot of the time.

python/lsst/cp/pipe/ptc.py
```diff
@@ -367,11 +434,269 @@ def runDataRef(self, dataRef, visitPairs):
...
+            dataRef.dataId['expId'] = v2
+            exp2 = self.isr.runDataRef(dataRef).exposure
+            del dataRef.dataId['expId']
+
+            checkExpLengthEqual(exp1, exp2, v1, v2, raiseWithMessage=True)
+            expTime = exp1.getInfo().getVisitInfo().getExposureTime()
+
+            tupleRows = []
+            for ampNumber, amp in enumerate(detector):
```
czwa, line 615: This method duplicates most of the functionality of computeStandardPtcAndNonLinearity, with only the measurement method changing.  If the output dataset can be unified, then this code can be moved up into the run() method, with the configuration defining with measurement code is run.

python/lsst/cp/pipe/ptc.py
```diff
@@ -367,11 +434,269 @@ def runDataRef(self, dataRef, visitPairs):
...
+                # covs: (i, j, var, cov, npix)
+                mu1, mu2, covs = self.getCovariancesAstier(exp1, exp2, region=amp.getBBox())
+                tupleRows += [(mu1, mu2) + cov + (ampNumber, expTime, amp.getName()) for cov in covs]
+                tags = ['mu1', 'mu2', 'i', 'j', 'var', 'cov', 'npix', 'ext', 'expTime', 'ampName']
+            allTags += tags
+            tupleRecords += tupleRows
+        covariancesWithTags = np.core.records.fromrecords(tupleRecords, names=allTags)
+        return covariancesWithTags
+
+    def getCovariancesAstier(self, exposure1, exposure2, region=None):
```
czwa, line 625: This method also shares a lot of code with measureMeanVarPair.  Can they be unified into one method?
(It seems like that would invalidate my previous comment, moving the config option check into that unified method, and out of the run() method.)

tests/test_ptc.py
```diff
@@ -95,6 +97,229 @@ def setUp(self):
...
+        dataset = task.getOutputPtcDataCovAstier(covFits)
+
+        # Chek the gain and that the ratio of the variance caclulated via cov Astier (FFT) and
+        # that calculated with the standard PTC is close to 1.
+        for amp in self.ampNames:
+            self.assertAlmostEqual(dataset.gain[amp][0], 0.75, places=2)
+            for v1, v2 in zip(varStandard[amp], dataset.varAdu[amp][0]):
+                self.assertAlmostEqual(v1/v2, 1.0, places=1)
+
+    def getCovariancesAstierDirect(self, exposure1, exposure2, config, region=None):
```
czwa, line 175: Should this be moved to ptc.py?  This duplicates the measureMeanVarPair() code as well.

tests/test_ptc.py
```diff
@@ -95,6 +97,229 @@ def setUp(self):
...
+        w12 = w1*w2
+        wDiff = np.where(diffIm.getMask().getArray() == 0, 1, 0)
+        w = w12*wDiff
+
+        maxRangeCov = config.maximumRangeCovariancesAstier
+        covs = self.computeCovDirect(diffIm.getImage().getArray(), w, maxRangeCov)
+
+        return mu1, mu2, covs
+
+    def computeCovDirect(self, diffImage, weightImage, maxRange):
```
czwa, line 217: Same with this.  If this is a way to calculate things, it should be in the ptc task as an optional way to calculate things.

tests/test_ptc.py
```diff
@@ -95,6 +97,229 @@ def setUp(self):
...
+                    nPix = nPix1 + nPix2
+                else:
+                    cov, nPix = self.covDirectValue(diffImage, weightImage, dx, dy)
+                if (dx == 0 and dy == 0):
+                    var = cov
+                outList.append((dx, dy, var, cov, nPix))
+
+        return outList
+
+    def covDirectValue(self, diffImage, weightImage, dx, dy):
```
czwa, line 267: Same  comment.

python/lsst/cp/pipe/astierCovPtcFit.py
```diff
@@ -0,0 +1,610 @@
...
+        a2 = a2[np.newaxis, xc:xc + range, yc:yc + range]
+        a3 = a3[np.newaxis, xc:xc + range, yc:yc + range]
+        ac = ac[np.newaxis, xc:xc + range, yc:yc + range]
+        c1 = c[np.newaxis, ::]
+        if mu is None:
+            mu = self.mu
+        # assumes that mu is 1d
+        bigMu = mu[:, np.newaxis, np.newaxis]*gain
+        # c(=a*b in Astier+19) also has a contribution to the last term, that is absent for now.
+        covModel = (bigMu/(gain*gain)*(a1*bigMu+2./3.*(bigMu*bigMu)*(a2+c1) +
```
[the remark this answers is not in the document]
plazas: I separated them, but maybe it did not complain because it was inside parentheses?

python/lsst/cp/pipe/astierCovPtcFit.py
```diff
@@ -0,0 +1,610 @@
...
+        return mu, covariance, model, weights
+
+    def __call__(self, params):
+        self.setParamValues(params)
+        chi2 = self.chi2()
+
+        return chi2
+
+
+class WeightedRes:
```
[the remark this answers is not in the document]
plazas: I think that you are right. It is called by "cov_fit_fft" in the original code, but I don't think it is used: https://github.com/PierreAstier/bfptc/blob/03deeac9eb1c5f8ca97eba9cf782654a2e4024e8/py/ptcfit.py#L437

python/lsst/cp/pipe/ptc.py
```diff
@@ -94,6 +97,16 @@ class MeasurePhotonTransferCurveTaskConfig(pexConfig.Config):
             "ASTIERAPPROXIMATION": "Approximation in Astier+19 (Eq. 16)."
         }
     )
+    doCovariancesAstier = pexConfig.Field(
```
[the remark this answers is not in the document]
plazas: The option here is between calculating the full covariances from flats and fitting the full (Eq. 20) model in Astier+19, and calculating what ptc.py has been calculating so far: just Cov00 (the variance; to produce what is traditionally called the PTC, which is var vs mean), fitting a polynomial or the approximation to Eq. 20 (which is Eq. 16 in Astier+19)  to the var vs mean curve, and producing linearizers from the “mean_signal” vs “exp_time” curve.  It is true that the full calculation of variances (doCovariancesAstier=True) includes as a special case Cov00=Var, so we could just use this full calculation of variances to produce the standard PTC (and then the linearizers, given the exposure times) and then eventually just use this code and get rid of most of the code that is currently in ptc.py (but for the moment I did not want to do create a major disruption in the way that we calculate the PTC, which seems to work fine, and that’s why decided to put the whole extra code in this ticket to be used as an option).

python/lsst/cp/pipe/ptc.py
plazas: Yes. That document/diagram (pipeline_astier_shape_of_ptc.pdf in DM-20070 in Jira) was made by Pierre Astier, to try to explain his whole code and paper. It turns out his original code does, at least, 3 major things:
1) Corrects for non-linearity using photodiode data
2) Corrects for CTI in a way which I don’t remember right now
3) Does the actual covariances calculations and fits those to the full model in the paper.
I decided to have this ticket focus on 3), and leave 1) and 2) to other parts of the ISR chain (the linearity correction and the CTI correction).

python/lsst/cp/pipe/astierCovFitParameters.py
```diff
@@ -0,0 +1,282 @@
...
+
+- `FitParameters` : a class to manage large parameter vectors.  It
+allows to easily fix/release specific parameters or entire subgroups,
+and remap the remaining free parameters into a contiguous vector.
+"""
+
+import numpy as np
+
+
+class Structure(object):
```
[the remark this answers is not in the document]
plazas: I’m not sure if changing "class Structure(object)" for a "pipeBase.Struct" will preserve the methods and properties of "StructArray(np.ndarray)" (based on "Structure") and "FitParameters" (based on "StructArray") in "astierCovFitParameters.py". In the end, "FitParameters" is only used in "self.params" for the fits in "astierCovPtcFit.py". I chose to leave the whole astierCovFitParameters.py to avoid breaking things.

python/lsst/cp/pipe/ptc.py
plazas: I moved the ISR part out of "ptc.py", which now operates on "postISR" images (created by "runIsr.py").

python/lsst/cp/pipe/astierCovPtcPlots.py
```diff
@@ -0,0 +1,536 @@
...
+    cb1 = plt.colorbar(im1)
+    cb1.ax.tick_params(labelsize='x-large')
+    ax1.set_title(r'$b \times 10^6$', fontsize='x-large')
+    ax1.xaxis.set_ticks_position('bottom')
+    plt.tight_layout()
+    pdfPages.savefig(fig)
+
+
+def plotPtcAndResiduals(covFits, ampName, pdfPages):
+    """Figure 7 in Astier+19"""
```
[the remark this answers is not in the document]
plazas: plotCovariances produces the residuals shown in the lower parts in Fig. 7, as well as C00 (var) vs signal (i.e., the PTC), so it's like another version of the exact Fig. 7 (produced by plotPtcAndResiduals)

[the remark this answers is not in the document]
plazas: I deleted "plotPtcAndResiduals", as the same information is plotted by "plotCovariances" (just in a different way).

python/lsst/cp/pipe/ptc.py
```diff
@@ -367,11 +434,269 @@ def runDataRef(self, dataRef, visitPairs):
...
+            # Calculate the full covariances as in Astier+19. The variances for the PTC will be
+            # cov[0,0].
+
+            tupleCovariancesWithTags = self.computeCovariancesAstier(dataRef, visitPairs, detector)
+            # use the np.recarray to obtain the CovFit objects
+            covFits, covFitsNoB = fitData(tupleCovariancesWithTags, maxMu=self.config.maxMeanSignal,
+                                          maxMuElectrons=self.config.maxMeanSignal,
+                                          r=self.config.maximumRangeCovariancesAstier)
+
+            if self.config.makePlots:
```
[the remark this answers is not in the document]
plazas: Created a new task defined in "plotPtc.py"

python/lsst/cp/pipe/ptc.py
plazas: I tried to clean this up a bit (e.g., disentangling the non-linearity function, trying to do data handling and I/O in dataRef, etc.).

python/lsst/cp/pipe/ptc.py
```diff
@@ -369,147 +362,218 @@ def runDataRef(self, dataRef, visitPairs):
...
+            numberAmps = len(amps)
+            numberAduValues = self.config.maxAduForLookupTableLinearizer
+            lookupTableArray = np.zeros((numberAmps, numberAduValues), dtype=np.float32)
+
+            # Fit (non)linearity of signal vs time curve.
+            # Fill up PhotonTransferCurveDataset object.
+            # Fill up array for LUT linearizer (tableArray).
+            # Produce coefficients for Polynomial ans Squared linearizers.
+            # Build linearizer objects.
+            datasetPtc, linsArray = self.fitNonLinearityAndBuildLinearizers(datasetPtc, detector,
```
czwa, line 419: This shouldn't modify datasetPtc.  All of the linearity information should be contained in the linearity object.  If something needs to be added to the linearity class definition to support that, we should add that.  Saving that in the PTC is not obvious to me.

python/lsst/cp/pipe/ptc.py
```diff
@@ -369,147 +362,218 @@ def runDataRef(self, dataRef, visitPairs):
...
+        """
+
+        # Fit NonLinearity
+        dataset = self.fitNonLinearity(dataset, tableArray=tableArray)
+
+        # Produce linearizer
         now = datetime.datetime.utcnow()
         calibDate = now.strftime("%Y-%m-%d")
 
+        linArray = []
```
czwa, line 478: Why not make the linearity type to fit/create a config option?  Since this code no longer needs to run everything from scratch, there's less of an issue of needing to do all the fits at once.  This simplifies linearity creation, making it more similar to the PTC generation/writes.

python/lsst/cp/pipe/ptc.py
```diff
@@ -369,147 +362,218 @@ def runDataRef(self, dataRef, visitPairs):
...
+            Dictionary of CovFit objects, with amp names as keys.
+
+        Returns
+        -------
+        dataset : `lsst.cp.pipe.ptc.PhotonTransferCurveDataset`
+            This is the same dataset as the input paramter, however, it has been modified
+            to include extra information such as the mask 1D array, gains, reoudout noise, measured signal,
+            measured variance, modeled variance, a, and b coefficient matrices (see Astier+19) per amplifier.
+            See the class `PhotonTransferCurveDatase`.
+            """
```
czwa, line 552: Odd indent that I thought would cause problems, but maybe doesn't?

python/lsst/cp/pipe/ptc.py
```diff
@@ -369,147 +362,218 @@ def runDataRef(self, dataRef, visitPairs):
...
-                dataset.rawExpTimes[ampName].append(expTime)
-                dataset.rawMeans[ampName].append(mu)
-                dataset.rawVars[ampName].append(varDiff)
-                dataset.inputVisitPairs[ampName].append((v1, v2))
+        return pipeBase.Struct(exitStatus=0)
 
-        numberAmps = len(detector.getAmplifiers())
-        numberAduValues = self.config.maxAduForLookupTableLinearizer
-        lookupTableArray = np.zeros((numberAmps, numberAduValues), dtype=np.float32)
+    def fitNonLinearityAndBuildLinearizers(self, dataset, detector, tableArray=None, log=None):
```
czwa, line 442: Can this be moved down to where fitNonLinearity is? Grouping the PTC methods separate from the Linearizer methods makes it much easier to trace through the code without big jumps.  It's possible your editor makes it easy to find related methods.

python/lsst/cp/pipe/ptc.py
```diff
@@ -783,12 +928,12 @@ def calculateLinearityResidualAndLinearizers(self, exposureTimeVector, meanSigna
         dataset.polynomialLinearizerCoefficients : `list` of `float`
             Coefficients for LinearizePolynomial, where corrImage = uncorrImage + sum_i c_i uncorrImage^(2 +
             i).
-            c_(j-2) = -k_j/(k_1^j) with units (DN^(1-j)). The units of k_j are DN/t^j, and they are fit from
-            meanSignalVector = k0 + k1*exposureTimeVector + k2*exposureTimeVector^2 +...
-                             + kn*exposureTimeVector^n, with n = "polynomialFitDegreeNonLinearity".
-            k_0 and k_1 and degenerate with bias level and gain, and are not used by the non-linearity
-            correction. Therefore, j = 2...n in the above expression (see `LinearizePolynomial` class in
-            `linearize.py`.)
+            c_(j-2) = -k_j/(k_1^j) with units DN^(1-j) (c.f., Eq. 37 of 2003.05978). The units of k_j are
```
czwa, line 931: It might be easier to put the full explanation in a docstring Notes, and keep the Returns list a bit simpler.
As mentioned above, putting everything necessary into the linearizer, and making the linearity type configurable will simplify this as well.

python/lsst/cp/pipe/ptc.py
```diff
@@ -563,138 +643,203 @@ def measureMeanVarPair(self, exposure1, exposure2, region=None):
...
-
-    def funcPolynomial(self, pars, x):
-        """Polynomial function definition"""
-        return poly.polyval(x, [*pars])
-
-    def funcAstier(self, pars, x):
-        """Single brighter-fatter parameter model for PTC; Equation 16 of Astier+19"""
-        a00, gain, noise = pars
-        return 0.5/(a00*gain*gain)*(np.exp(2*a00*x*gain)-1) + noise/(gain*gain)
+        linearizer.validate()
```
czwa, line 828: Docstrings?

python/lsst/cp/pipe/plotPtc.py
```diff
@@ -0,0 +1,851 @@
...
+        self.plotNormalizedCovariances(covFits, covFitsNoB, 0, 1, pdfPages, log=log)
+        self.plotNormalizedCovariances(covFits, covFitsNoB, 1, 0, pdfPages, log=log)
+        self.plot_a_b(covFits, pdfPages)
+        self.ab_vs_dist(covFits, pdfPages, brange=4)
+        self.plotAcoeffsSum(covFits, pdfPages)
+        self.plotRelativeBiasACoeffs(covFits, covFitsNoB, maxMuElectrons, pdfPages)
+
+        return
+
+    def plotCovariances(self, covFits, pdfPages):
```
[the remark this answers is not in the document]
plazas: I'll make them static: I don't think they need to modify the state of the class.

python/lsst/cp/pipe/ptc.py
plazas: I have implemented this, but we always fit for the polynomial regardless of the type (that's why the code was persisting everything, since two of the linearizers, as they are implemented now, are by-products of tat polynomial fit).  I think the linearizer squared is redundant with the polynomial. On the other hand, I think is good to have the options split for the reasons you list.

python/lsst/cp/pipe/plotPtc.py
```diff
@@ -533,7 +513,42 @@ def ab_vs_dist(self, covFits, pdfPages, brange=4):
...
         return
 
-    def plotRelativeBiasACoeffs(self, covFits, covFitsNoB, mu_el, pdfPages, maxr=None):
+    @staticmethod
+    def plotAcoeffsSum(self, covFits, pdfPages):
+        """Fig. 14. of Astier+19
+
+        Cumulative sum of a_ij as a function of maximum separation. This plot displays the average over
+        channels.
+        """
```
czwa, line 522: I think these need Parameters docstrings as well.

python/lsst/cp/pipe/plotPtc.py
```diff
@@ -533,7 +513,42 @@ def ab_vs_dist(self, covFits, pdfPages, brange=4):
...
+        ax.set_ylabel(r'$[\sum_{|i|<n\  &\  |j|<n} a_{ij}] / |a_{00}|$', fontsize='x-large')
+        ax.set_xlabel('n', fontsize='x-large')
+        ax.tick_params(axis='both', labelsize='x-large')
+        plt.tight_layout()
+        pdfPages.savefig(fig)
+
+        return
+
+    @staticmethod
+    def plotRelativeBiasACoeffs(self, covFits, covFitsNoB, maxMu, pdfPages, maxr=None):
```
czwa, line 551: Same here: Parameters docstrings.

PR #42, opened 2020-07-14
DM-25934: MeasurePhotonTransferCurveTask appears to ignore the defect mask

plazas:


python/lsst/cp/pipe/ptc.py
```diff
@@ -598,12 +603,21 @@ def measureMeanVarCov(self, exposure1, exposure2, region=None, covAstierRealSpac
         im1Area = afwMath.binImage(im1Area, self.config.binSize)
         im2Area = afwMath.binImage(im2Area, self.config.binSize)
 
-        statsCtrl = afwMath.StatisticsControl()
-        statsCtrl.setNumSigmaClip(self.config.nSigmaClipPtc)
-        statsCtrl.setNumIter(self.config.nIterSigmaClipPtc)
+        im1MaskVal = exposure1.getMask().getPlaneBitMask(self.config.maskNameList)
+        im1StatsCtrl = afwMath.StatisticsControl(self.config.nSigmaClipPtc,
```
czwa, line 607: I think setting things with the statsCtrl.setXYZ() methods is clearer, but this is perfectly fine as the config options state what they are.

PR #44, opened 2020-07-20
DM-25984: Investigate why the afw means of flat images are NANs for several amps of BOT data after DM-25934

plazas:


python/lsst/cp/pipe/ptc.py
```diff
@@ -139,7 +139,7 @@ class MeasurePhotonTransferCurveTaskConfig(pexConfig.Config):
     maskNameList = pexConfig.ListField(
         dtype=str,
         doc="Mask list to exclude from statistics calculations.",
-        default=['DETECTED', 'BAD', 'NO_DATA'],
+        default=['SUSPECT', 'BAD', 'NO_DATA'],
```
mfisherlevine, line 142: Seems like adding SUSPECT is 👍 but why are we removing DETECTED? Maybe it should never have been there in the first place? Seems weird that detection would even have been run, so I don't disagree with this change, just curious what's happening here.

python/lsst/cp/pipe/ptc.py
```diff
@@ -618,6 +628,8 @@ def measureMeanVarCov(self, exposure1, exposure2, region=None, covAstierRealSpac
         #  Clipped mean of images; then average of mean.
         mu1 = afwMath.makeStatistics(im1Area, afwMath.MEANCLIP, im1StatsCtrl).getValue()
         mu2 = afwMath.makeStatistics(im2Area, afwMath.MEANCLIP, im2StatsCtrl).getValue()
+        if np.isnan(mu1) or np.isnan(mu2):
+            return np.nan, np.nan, np.nan
```
mfisherlevine, line 632: This function claims to return `float, float, list`, so either change the doc or maybe make this return an empty list if that works for other places as a sentinel value.

python/lsst/cp/pipe/ptc.py
```diff
@@ -404,12 +404,18 @@ def runDataRef(self, dataRef, visitPairs):
...
 
             tupleRows = []
+            nAmpsNan = 0
             for ampNumber, amp in enumerate(detector):
                 ampName = amp.getName()
                 # covAstier: (i, j, var (cov[0,0]), cov, npix)
                 doRealSpace = self.config.covAstierRealSpace
                 muDiff, varDiff, covAstier = self.measureMeanVarCov(exp1, exp2, region=amp.getBBox(),
                                                                     covAstierRealSpace=doRealSpace)
+                if np.isnan(muDiff) and np.isnan(varDiff) and np.isnan(covAstier):
```
mfisherlevine, line 414: Are you OK with some but not all of these being NaN? Or do you want `or`s here?

python/lsst/cp/pipe/ptc.py
plazas: I also think that perhaps this should not have been there in the first place (it does not seem to me that DETECTED pixels should be in the list of pixels to mask), so I considered removing it.

python/lsst/cp/pipe/ptc.py
plazas: I'll put `or`s to report any NaN in the covariances too.

RobertLuptonTheGood:
Where are these NaNs coming from?  There should be none in the input frames (because we interpolate them all), so it's because there are labelled as defects.  We addressed this problem on PFS; please check with @PaulPrice

craiglagegit:
It happened in saturated flats.  Apparently all of the pixels in the amp are saturated so with no good pixels MEANCLIP returned a NaN.
________________________________

tests/test_ptc.py
```diff
@@ -281,6 +281,44 @@ def test_meanVarMeasurement(self):
...
+        temp *= expectedMu1
+        diffIm = im1.clone()
+        diffIm *= expectedMu2
+        diffIm -= temp
+        diffIm /= expectedMu
+
+        expectedVar = 0.5*np.nanvar(diffIm.image.array)
+
+        self.assertLess(np.sqrt(expectedVar) - np.sqrt(varDiff), 1)
+        self.assertLess(expectedMu - mu, 1)
```
mfisherlevine, line 309: Could you just put a comment in the code for why these conditions are being tested for? Clearly it works, but would be nice to for the next person not to have to think it all through.

tests/test_ptc.py
```diff
@@ -281,6 +281,44 @@ def test_meanVarMeasurement(self):
...
+    def test_meanVarMeasurementAllNan(self):
+        task = self.defaultTask
+        self.flatExp1.image.array[:, :] = np.nan
+        self.flatExp2.image.array[:, :] = np.nan
+
+        mu, varDiff, covDiff = task.measureMeanVarCov(self.flatExp1, self.flatExp2)
+
+        self.assertTrue(mu is np.nan)
+        self.assertTrue(varDiff is np.nan)
+        self.assertTrue(covDiff is np.nan)
```
mfisherlevine, line 320: I think these should all be changed to `self.assertTrue(np.isnan(mu))` etc because if those functions above returned `float('nan')` instead then this would fail, and that would be annoying and hard to debug.

tests/test_ptc.py
```diff
@@ -281,6 +281,44 @@ def test_meanVarMeasurement(self):
         self.assertLess(self.flatWidth - np.sqrt(varDiff), 1)
         self.assertLess(self.flatMean - mu, 1)
 
+    def test_meanVarMeasurementSomeNan(self):
+        task = self.defaultTask
+        self.flatExp1.image.array[20:30] = np.nan
+        self.flatExp2.image.array[20:30] = np.nan
```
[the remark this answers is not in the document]
plazas: Changed it to [20:30, :], thanks for the suggestion.

PR #45, opened 2020-07-22
DM-26067: Fix bug in ptc.py that calls np.isnan on array (after DM-25984)

plazas:


python/lsst/cp/pipe/ptc.py
```diff
@@ -411,7 +411,7 @@ def runDataRef(self, dataRef, visitPairs):
                 doRealSpace = self.config.covAstierRealSpace
                 muDiff, varDiff, covAstier = self.measureMeanVarCov(exp1, exp2, region=amp.getBBox(),
                                                                     covAstierRealSpace=doRealSpace)
-                if np.isnan(muDiff) or np.isnan(varDiff) or np.isnan(covAstier):
```
czwa, line 414: Do we not care if the covariance is populated?  If this catches a failure, can we return `None` instead of the matrix, so we can check that case?

PR #46, opened 2020-08-12
DM-23302: Add ci_cpp package to allow calibration product generation to be tested

czwa:
Update cpCertify for upstream API changes.
Fix errors detected by ci_cpp code.


python/lsst/cp/pipe/cpCertify.py
```diff
@@ -137,11 +137,6 @@ def convertStorageClass(self, data, datasetTypeName):
         data : `lsst.afw.image.Image` or `lsst.afw.image.MaskedImage`
             Converted image data to register.
         """
-        if datasetTypeName in ('bias', 'dark'):
-            data = data.getImage()
-        # elif datasetTypeName in ('flat', ):
-        #     data = data.getMaskedImage()
-
```
plazas, line 144: You deleted this, and now the function just returns "data", which is the input. So, why do we need this function?

python/lsst/cp/pipe/cpCertify.py
```diff
@@ -155,10 +150,7 @@ def registerDatasetType(self, datasetTypeName, dataId):
             Data ID providing the list of dimensions for the new
             datasetType.
         """
-        storageClassMap = {'bias': 'ImageF',
-                           'dark': 'ImageF',
-                           'flat': 'ExposureF',
-                           }
+        storageClassMap = {'crosstalk': 'CrosstalkCalib'}
```
plazas, line 153: Why substitute bias, dark, and flat for just crosstalk? I guess the storage classes for the former are defined somewhere else?

python/lsst/cp/pipe/cpDarkTask.py
```diff
@@ -37,14 +37,14 @@ class CpDarkConnections(pipeBase.PipelineTaskConnections,
     inputExp = cT.Input(
         name="cpDarkISR",
         doc="Input pre-processed exposures to combine.",
-        storageClass="ExposureF",
```
plazas, line 40: Whys is this change necessary? (and below; from "ExposureF" to "Exposure". )

timj: You should only use ExposureF if you absolutely require 32-bit float exposures. If the next task doesn't care what type of Exposure it gets then it's better to be generic.

python/lsst/cp/pipe/cpCertify.py
czwa: Inertia.  Deleted.

python/lsst/cp/pipe/cpCertify.py
czwa: All other gen3 producible calibration types should be `ExposureF`, so that's now the default.  I anticipate that as other gen2 is converted, the `storageClassMap` will need to be extended.

bin.src/certifyCalibration.py
```diff
@@ -67,11 +67,11 @@
...
+                                 inputCollection=args.inputCollection,
+                                 outputCollection=args.outputCollection)
 
-    crozier.findInputs(args.datasetTypeName)
+    certify.findInputs(args.datasetTypeName)
     if not args.skipCL:
-        crozier.addCalibrationLabel(beginDate=args.beginDate, endDate=args.endDate)
-    crozier.registerCalibrations(args.datasetTypeName)
+        certify.addCalibrationLabel(beginDate=args.beginDate, endDate=args.endDate)
+    certify.registerCalibrations(args.datasetTypeName)
```
mfisherlevine, line 77: `certifier` as a noun perhaps? Don't much care really as it's in a bin script though.

mfisherlevine (approved):

bin.src/certifyCalibration.py
```diff
@@ -67,11 +67,11 @@
     butler = Butler(args.root, run=args.inputCollection)
```
timj, line 67: In the butler root above can you say "URI to butler" or something because it's not always a path.

timj:
As a general comment, we are trying to move away from lots of code in bin.src python scripts. They should follow the new approach where the bin.src just has two lines to call `main` in a `script` module. See how commands work in obs_lsst. This then lets you automatically include sphinx documentation for the command line tooling.

PR #47, opened 2020-08-25
DM-26453: Add sqrt(var) as weight to EXPAPPROXIMATION and POLYNOMIAL fit residual in ptc.py

plazas:


python/lsst/cp/pipe/utils.py
```diff
@@ -374,15 +386,19 @@ def fitBootstrap(initialParams, dataX, dataY, function, confidenceSigma=1.):
     reducedChiSqBootstrap : `float`
         Reduced chi squared.
```
mfisherlevine, line 387: Maybe amend this too so it's in keeping with the other one about the optional weighting? Or make the other like this, I think that's actually fine too.

python/lsst/cp/pipe/utils.py
```diff
@@ -391,7 +407,7 @@ def errFunc(p, x, y):
         randomDelta = np.random.normal(0., sigmaErrTotal, len(dataY))
         randomDataY = dataY + randomDelta
         randomFit, _ = leastsq(errFunc, initialParams,
-                               args=(dataX, randomDataY), full_output=0)
+                               args=(dataX, randomDataY, 1./np.sqrt(randomDataY)), full_output=0)
```
mfisherlevine, line 410: Seems like `randomDataY` can be negative here - what happens then? I think you end up with a `RuntimeWarning` and a bunch of `nan`s in the return array, and then I'm not sure what happens downstream...

mfisherlevine (commented):
Very minor comments.

PR #48, opened 2020-08-26
 DM-24703: Make linearity a subclass of lsst.ip.isr.IsrCalib

czwa:
Move linearity calibration construction out of `ptc.py`, and make the new task gen3 compliant.

### DIFF
--- a/python/lsst/cp/pipe/linearity.py
+++ b/python/lsst/cp/pipe/linearity.py
@@ -133,6 +133,20 @@ class LinearitySolveTask(pipeBase.PipelineTask, pipeBase.CmdLineTask):
                 Final linearizer calibration.
             ``outputProvenance`` : `lsst.ip.isr.IsrProvenance`
                 Provenance data for the new calibration.
+
+        Notes
+        -----
+        This task currently fits only polynomial-defined corrections,
+        where the correction coefficients are defined such that:
+            corrImage = uncorrImage + sum_i c_i uncorrImage^(2 + i)
+        These `c_i` are defined in terms of the direct polynomial fit:
+            meanVector ~ P(x=timeVector) = sum_j k_j x^j
+        such that c_(j-2) = -k_j/(k_1^j) in units of DN^(1-j) (c.f.,
+        Eq. 37 of 2003.05978). The `config.polynomialOrder` defines
+        the maximum order of x^j to fit.  As k_0 and k_1 are
+        degenerate with bias level and gain, they are not included in
+        the non-linearity correction.
+
         """
         if camera:
             detector = camera[inputDims['detector']]
@@ -206,7 +220,23 @@ class LinearitySolveTask(pipeBase.PipelineTask, pipeBase.CmdLineTask):
         )
 
     def debugFit(self, stepname, timeVector, meanVector, linearizer, ampName):
-        """
+        """Debug method for linearity fitting.
+
+        Parameters
+        ----------
+        stepname : `str`
+            A label to use to check if we care to debug at a given
+            line of code.
+        timeVector : `numpy.array`
+            The values to use as the independent variable in the
+            linearity fit.
+        meanVector : `numpy.array`
+            The values to use as the dependent variable in the
+            linearity fit.
+        linearizer : `lsst.ip.isr.Linearizer`
+            The linearity correction to compare.
+        ampName : `str`
+            Amplifier name to lookup linearity correction values.
         """
         frame = getDebugFrame(self._display, stepname)
         if frame:
--- a/python/lsst/cp/pipe/linearity.py
+++ b/python/lsst/cp/pipe/linearity.py
@@ -1,0 +1,235 @@
+# This file is part of cp_pipe.
+#
+# Developed for the LSST Data Management System.
+# This product includes software developed by the LSST Project
+# (https://www.lsst.org).
+# See the COPYRIGHT file at the top-level directory of this distribution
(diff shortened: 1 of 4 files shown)


python/lsst/cp/pipe/linearity.py
```diff
@@ -0,0 +1,235 @@
...
+    """
+    linearityType = pexConfig.ChoiceField(
+        dtype=str,
+        doc="Type of linearizer to construct.",
+        default="Polynomial",
+        allowed={
+            "LookupTable": "Create a lookup table solution.",
+            "Polynomial": "Create an arbitrary polynomial solution.",
+            "Squared": "Create a single order squared solution.",
+            "None": "Create a dummy solution.",
```
plazas, line 73: Craig has recently been obtaining results that indicate that we might also need to implement a spline fit option. Could this be something added to this ticket, or perhaps something for another ticket?

timj: Another ticket please.

python/lsst/cp/pipe/linearity.py
```diff
@@ -0,0 +1,235 @@
...
+        linearizer.updateMetadata(setDate=True)
+        provenance = IsrProvenance(calibType='linearizer')
+
+        return pipeBase.Struct(
+            outputLinearizer=linearizer,
+            outputProvenance=provenance,
+        )
+
+    def debugFit(self, stepname, timeVector, meanVector, linearizer, ampName):
+        """
```
plazas, line 209: No docstring here?  I guess un in L197 you are calling this function to produce linearity plots for debugging while the code is running, right? (like in the crosstalk code)

python/lsst/cp/pipe/ptc.py
```diff
@@ -85,27 +84,17 @@ class MeasurePhotonTransferCurveTaskConfig(pexConfig.Config):
         doc="Degree of polynomial to fit the PTC, when 'ptcFitType'=POLYNOMIAL.",
         default=3,
     )
+    linearity = pexConfig.ConfigurableField(
```
plazas, line 87: I see that this is used below when calling this new linearity subtask, but I don't quite understand why this is a configuration field. (why is it here at this point?). I'm probably missing something.

python/lsst/cp/pipe/ptc.py
```diff
@@ -171,32 +160,27 @@ class MeasurePhotonTransferCurveTaskConfig(pexConfig.Config):
...
+#     """
+#     # Normalized coefficients for polynomial NL correction
+#     polynomialLinearizerCoefficients: list
+#     # Normalized coefficient for quadratic polynomial NL correction (c0)
+#     quadraticPolynomialLinearizerCoefficient: float
+#     # LUT array row for the amplifier at hand
+#     linearizerTableRow: list
+#     meanSignalVsTimePolyFitPars: list
+#     meanSignalVsTimePolyFitParsErr: list
+#     meanSignalVsTimePolyFitReducedChiSq: float
```
plazas, line 183: I see that this dataclass is fully commented. If it is not being used anymore, then it should be completely removed.

python/lsst/cp/pipe/ptc.py
```diff
@@ -792,282 +761,6 @@ def covDirectValue(self, diffImage, weightImage, dx, dy):
...
-        datase members:
-
-        dataset.polynomialLinearizerCoefficients : `list` of `float`
-            Coefficients for LinearizePolynomial, where corrImage = uncorrImage + sum_i c_i uncorrImage^(2 +
-            i).
-            c_(j-2) = -k_j/(k_1^j) with units DN^(1-j) (c.f., Eq. 37 of 2003.05978). The units of k_j are
-            DN/t^j, and they are fit from meanSignalVector = k0 + k1*exposureTimeVector +
-            k2*exposureTimeVector^2 + ... + kn*exposureTimeVector^n, with
-            n = "polynomialFitDegreeNonLinearity". k_0 and k_1 and degenerate with bias level and gain,
-            and are not used by the non-linearity correction. Therefore, j = 2...n in the above expression
```
plazas, line 916: It seems that this explanation and reference got lost, but I think it helped the user understand a bit better how the polynomial correction was constructed. Could this (or a summary) be placed back somewhere? Or maybe we could still document this in the comments of the Jira ticket, at least.

python/lsst/cp/pipe/linearity.py
czwa: I've added DM-26545 for this.

python/lsst/cp/pipe/linearity.py
czwa: That's the plan.  Since this can be run independent of the previous steps (ISR, PTC), the results can be checked through the debug code.  I've added the missing docstring.

python/lsst/cp/pipe/ptc.py
czwa: By making this a `ConfigurableField`, the target task can be instantiated as a subtask (L334).  This allows the PTC code to call the subtask's run method directly (L448).  This moves the linearity configuration to `config.linearity.polynomialOrder`and similar fields, instead of being directly part of the `MeasurePhotonTransferCurveTaskConfig`.

python/lsst/cp/pipe/ptc.py
czwa: Added and updated this discussion in the Notes section of the run method docstring.

PR #49, opened 2020-09-03
DM-26595: Have cp_pipe/ptc.py continue if there is a bad flat pair

plazas:


python/lsst/cp/pipe/ptc.py
```diff
@@ -404,10 +404,15 @@ def runDataRef(self, dataRef, visitPairs):
...
-            exp1 = dataRef.get("postISRCCD", immediate=True)
-            dataRef.dataId['expId'] = v2
-            exp2 = dataRef.get("postISRCCD", immediate=True)
+            try:
+                dataRef.dataId['expId'] = v1
+                exp1 = dataRef.get("postISRCCD", immediate=True)
+                dataRef.dataId['expId'] = v2
+                exp2 = dataRef.get("postISRCCD", immediate=True)
+            except RuntimeError:
+                self.log.warn(f"postISR exposure for either visit {v1} or visit {v2} could not be retreived. "
```
czwa, line 413: This should probably not use "visit" anymore, as these should all be expId values now.  There are a lot of them already in place, but we should at least not add any more.

python/lsst/cp/pipe/ptc.py
```diff
@@ -404,10 +404,15 @@ def runDataRef(self, dataRef, visitPairs):
...
+                              "Ignoring flat pair.")
+                continue
```
czwa, line 415: Does this pair show up as masked on ~L303?  I'm trying to think if there are downstream issues due to dropping this pair.

czwa (approved):
It looks like it should solve the problem.

python/lsst/cp/pipe/ptc.py
plazas: The skipped pair does not show up in the return of {{getVisitsUsed}} (as this function only returns those pairs that were used), so that means that it does show up as masked. Also, as you point out in Jira, the example in there runs.

PR #50, opened 2020-09-04
DM-26616: Gain slope is inverted in EXPAPPROXIMATION case in plotPtc.py

plazas:


mfisherlevine (approved):
Code looks fine, and I'll trust you that it's improving the plots 🙂

PR #52, opened 2020-09-14
DM-26697: Deprecate needing to pre-pair exposures in PTC task, and calculate an appropriate pair based on header data.

plazas:


python/lsst/cp/pipe/ptc.py
```diff
@@ -255,9 +254,9 @@ def __setattr__(self, attribute, value):
             self.__dict__[attribute] = value
 
     def getVisitsUsed(self, ampName):
```
czwa, line 256: We're not really using visits anywhere, right?  Can we drop it everywhere in favor of 'exposure', or should that be a separate ticket?  We'll want to do it at some point.

python/lsst/cp/pipe/ptc.py
```diff
@@ -446,12 +413,42 @@ def runDataRef(self, dataRef, visitPairs):
 
             butler.put(linearizer, datasetType='Linearizer', dataId={'detector': detNum,
                        'detectorName': detName, 'calibDate': calibDate})
-
-        self.log.info(f"Writing PTC data to {dataRef.getUri(write=True)}")
+        rerunDir = list(dataRef.getButler().storage.repositoryCfgs.keys())[0]
```
czwa, line 416: This seems like a lot of work for something that the butler should be handling.  I also don't think that this will work in gen3.

python/lsst/cp/pipe/ptc.py
```diff
@@ -446,12 +413,42 @@ def runDataRef(self, dataRef, visitPairs):
...
+        for dataRef in dataRefList:
+            try:
+                tempFlat = dataRef.get("postISRCCD")
+            except RuntimeError:
+                self.log.warn(f"postISR exposure could not be retrieved. Ignoring flat.")
+                continue
+            expTime = tempFlat.getInfo().getVisitInfo().getExposureTime()
+            listAtExpTime = sortedPairs.setdefault(expTime, [])
+            if len(listAtExpTime) < 2:
+                listAtExpTime.append(tempFlat)
```
czwa, line 444: Don't we want to sort on DATE-OBS or SEQNUM or something to prevent multiple flat ramps being paired incorrectly?  This also looks like it will toss any extra exposures, even when they could be used as well.

czwa (changes requested):
I'm concerned with the `makePairs` method, as it seems like it only will accept two exposures at a given exposure time, and that it may not pair the shortest time interval pair, so there's less guarantee that the illumination will be stable.

python/lsst/cp/pipe/ptc.py
```diff
@@ -446,12 +413,42 @@ def runDataRef(self, dataRef, visitPairs):
...
         dataRef.put(datasetPtc, datasetType="photonTransferCurveDataset")
 
         return pipeBase.Struct(exitStatus=0)
 
+    def makePairs(self, dataRefList):
+        """Produce a list of flat pairs indexed by exposure time.
+
+        Parameters
+        ----------
+        dataReflist : `list` [`lsst.daf.peristence.ButlerDataRef`]
```
timj, line 427: Is there a plan for making this gen3 compatible?  (also typo in the class name).

python/lsst/cp/pipe/ptc.py
```diff
@@ -446,12 +413,42 @@ def runDataRef(self, dataRef, visitPairs):
...
+            Data references for exposures for detectors to process.
+
+        Return
+        ------
+        sortedPairs : `dict`
```
timj, line 432: I think you need to say dict of what? Keys are exposure time in seconds and values are list of `Exposure`? But only the first two exposures that have that same exposure time?

python/lsst/cp/pipe/ptc.py
czwa: This is step one of that conversion.  https://jira.lsstcorp.org/browse/DM-21786 is the umbrella ticket for this.

python/lsst/cp/pipe/ptc.py
plazas: Do we need this line then?  {{dataRef.getUri(write=True)}} fails. This line is printing the location of the dataset pickle file. Is there another way of getting this, if we decide that we need to print this information?

czwa: I don't think we need to say where we're writing things.  A simple "Writing PTC data." will note that it's done with the calculation.

python/lsst/cp/pipe/ptc.py
plazas: I thought we were assuming for this task that the user would pass flat fields from a single PTC ramp. Are you suggesting that we should check that the DATE-OBS is the same for all exposures? Or do you have something else in mind? What if the flat ramp starts one night and continues to the next day, resulting in the images having different dates?  
The way we calculate the PTC now is by taking the difference between two flats at the same expTime to reduce PRNU. If for some reason there are more than two flats for the same expTime, we still need to pick two (the first two, in this code). Do you have something in mind for using possible extra exposures at the same expTime?

python/lsst/cp/pipe/ptc.py
plazas: Thanks for pointing out the typo. Regarding gen3 compatibility, I'm not sure about what you are referring to. Are you talking about the "makePairs" function in particular? Or the PTC task in general? If the latter, yes, this ticket is part of a larger ticket (DM-21786) to convert the PTC task to gen3.

python/lsst/cp/pipe/ptc.py
plazas: Yes, I agree, I'll add more information about the dictionary: the keys and values are as you describe. The input exposures are not necessarily sorted by ascending time in the command line, so this function organizes them in a dictionary. I can change the name sorted to something else (e.g.,  just "pairs").

python/lsst/cp/pipe/ptc.py
timj: I was just noting that this is brand new code that will have to be rewritten on a November timeline.

python/lsst/cp/pipe/ptc.py
timj: I definitely would not use `sorted` anywhere in the function if nothing is doing any sorting.

python/lsst/cp/pipe/ptc.py
czwa: I thought it would be general for any number of inputs.  The ramps that I'd looked at (from LATISS, though, so maybe different from how they're taking them on BOT) had four exposures at each exposure time.  More entries should provide better fits, as we're less dependent on any single pair.

plazas: I do remember the LATIISS ramps with more than 2 exposures per exposure time, but I think that the PTC, as it's implemented now, uses the difference of only two. I'll add a message as you suggest.

python/lsst/cp/pipe/ptc.py
plazas: The idea is that after DM-21786 the PTC task will be fully converted to gen3.

python/lsst/cp/pipe/ptc.py
czwa: I think dropping exposures is a mistake.  This also isn't deterministic: for a set of equal exposure time exposures `1^2^3^4`, `1^3^2^4`, etc. will all give different outputs.  Even if you want to keep only one pair for each exposure time, you need to do some sort operation to ensure that order transposes all give the same result.

timj: VisitInfo does include the date of observation if that helps.

plazas: The PTC task uses only two flats at a given exposure time. Should we require the user to make sure that this is the case? We could also look at the date of observation, as Tim suggests, but what if more than 2 flats have the same date of observation?

timj: How do you have two flats with the same date of observation and same detector? Isn't that impossible for a single instrument?

czwa: I think this is the "date = %Y-%m-%d %H:%M:%S" vs "date = %Y-%m-%d" difference.  Sorting along the first (which I think is what `getVisitInfo().getDate()` returns) should sort this issue.

plazas: I don't know, is it? If that is the case, how are you suggesting to use the date of observation to end up with only two flats at the same exposure time? (in case of Nflats with same expTime > 2).

plazas: I just saw Chris's comment (after I posted mine, replying to Tim). Thanks!

timj: Yes, VisitInfo stores the full date, not just the day.

czwa: This looks right to me.  If you can double check that it's deterministic (something like the example above where exposures are specified out of order but still return the in-order sorting), then I think this is done.

czwa (approved):

PR #55, opened 2020-09-30
DM-26725: Update the PTC dataset to be a proper IsrCalib

plazas:


python/lsst/cp/pipe/ptc.py
```diff
@@ -406,19 +304,20 @@ def runDataRef(self, dataRefList):
                 doRealSpace = self.config.covAstierRealSpace
                 muDiff, varDiff, covAstier = self.measureMeanVarCov(exp1, exp2, region=amp.getBBox(),
                                                                     covAstierRealSpace=doRealSpace)
+                datasetPtc.rawExpTimes[ampName].append(expTime)
+                datasetPtc.rawMeans[ampName].append(muDiff)
+                datasetPtc.rawVars[ampName].append(varDiff)
+                datasetPtc.inputExpIdPairs[ampName] = expIds
```
czwa, line 310: I don't think it's necessarily wrong, but doesn't this reset `datasetPtc.inputExpIdPairs[ampName]` on each loop through the `expPairs`?

python/lsst/cp/pipe/ptc.py
```diff
@@ -437,28 +336,30 @@ def runDataRef(self, dataRefList):
             # Fill up PhotonTransferCurveDataset object.
             datasetPtc = self.fitPtc(datasetPtc, self.config.ptcFitType)
 
+        detName = detector.getName()
+        now = datetime.datetime.utcnow()
+        calibDate = now.strftime("%Y-%m-%d")
+        butler = dataRef.getButler()
+        datasetPtc.updateMetadata(setDate=True, detectorName=detName, detectorId=detector.getId(),
```
czwa, line 343: `updateMetadata` now supports passing the camera and detector directly, which simplifies this a bit.

python/lsst/cp/pipe/ptc.py
```diff
@@ -543,48 +444,98 @@ def fitCovariancesAstier(self, dataset, covariancesWithTagsArray):
...
+                    wc, varMask) = fit.getFitData(0, 0, divideByMu=False, returnMasked=True)
+                gain = fit.getGain()
+                dataset.expIdMask[amp] = varMask
+                dataset.gain[amp] = gain
+                dataset.gainErr[amp] = fit.getGainErr()
+                dataset.noise[amp] = np.sqrt(fit.getRon())
+                dataset.noiseErr[amp] = fit.getRonErr()
+
+                padLength = lenInputTimes - len(varVecFinal)
+                dataset.finalVars[amp] = np.pad(varVecFinal/(gain**2), (0, padLength), 'constant',
```
czwa, line 505: I see, these are padded during creation.

python/lsst/cp/pipe/ptc.py
```diff
@@ -543,48 +444,98 @@ def fitCovariancesAstier(self, dataset, covariancesWithTagsArray):
...
+                                                constant_values=-9999.0)
+                dataset.finalModelVars[amp] = np.pad(varVecModel/(gain**2), (0, padLength), 'constant',
+                                                     constant_values=-9999.0)
+                dataset.finalMeans[amp] = np.pad(meanVecFinal/gain, (0, padLength), 'constant',
+                                                 constant_values=-9999.0)
+            else:
+                # Bad amp
+                # Entries need to have proper dimensions so read/write with astropy.Table works.
+                matrixSide = self.config.maximumRangeCovariancesAstier
+                nanMatrix = np.empty((matrixSide, matrixSide))
```
czwa, line 515: `np.full((matrixSide, matrixSide), np.nan)` might be clearer.

czwa (approved):
I think it's just typos and me possibly being confused about things.

python/lsst/cp/pipe/ptc.py
plazas: I took it out of the expPairs loop to set it once.

tests/test_ptc.py
```diff
@@ -88,6 +90,62 @@ def setUp(self):
...
+            localDataset.covariancesModelNoB[ampName] = np.full((nSignalPoints, nSideCovMatrix,
+                                                                nSideCovMatrix), np.nan)
+            localDataset.covariancesSqrtWeightsNoB[ampName] = np.full((nSignalPoints, nSideCovMatrix,
+                                                                      nSideCovMatrix), np.nan)
+            localDataset.aMatrixNoB[ampName] = np.full((nSideCovMatrix, nSideCovMatrix), np.nan)
+            localDataset.finalVars[ampName] = self.noiseSq + self.c1*self.flux*np.arange(nSignalPoints)
+            localDataset.finalModelVars[ampName] = np.repeat(100.0, nSignalPoints)
+            localDataset.finalMeans[ampName] = self.flux*np.arange(nSignalPoints)
+
+        # Round trip (write-read) with FITS
```
czwa, line 139: This should test equivalence of the local dataset and the one read back from disk.  `tempfile` will also handle the file IO so you don't need to.

tests/test_ptc.py
```diff
@@ -88,6 +90,62 @@ def setUp(self):
             self.dataset.rawExpTimes[ampName] = timeVec
             self.dataset.rawMeans[ampName] = muVec
 
+    def test_ptcDataset(self):
```
czwa, line 93: Can this be moved to `ip_isr`?  This tests the integrity of the calibration definition, not the measurement code.

czwa (approved):
If the round-trip versions are equivalent, then I think this is done (although please consider moving that test to ip_isr).

tests/test_ptc.py
timj: Yes, you want to write it to fits and yaml and then read both back in and ensure they agree with what you had originally (and yes this test should be with the calibration class code in ip_isr.

PR #58, opened 2020-10-13
DM-27159: measurePhotonTransferCurve.py fails on LSSTCam

plazas:


mfisherlevine (commented):
Looks good, but see OOB comments about logic on L405/407 re warnings etc.

PR #60, opened 2020-10-16
DM-27185: ptc.py fails with ptcFitType=FULLCOVARIANCE

plazas:


python/lsst/cp/pipe/utils.py
```diff
@@ -282,6 +282,51 @@ def makeDataRefList(self, namespace):
             self.refList += refList
 
 
+def irlsFit(initialParams, dataX, dataY, function, weightsY=None):
```
czwa, line 285: The two commits adding this should be squashed together so the indentations are correct on the first ticket.

czwa (approved):

PR #61, opened 2020-10-23
DM-26944: Move certifyCalibrations.py content to cpCertify.py

czwa:
Move certifyCalibration code to cpCertify.


python/lsst/cp/pipe/cpCertify.py
```diff
@@ -69,19 +75,78 @@ def __init__(self, *, registry, inputCollection, outputCollection, lastRunOnly=T
...
+    parser : `argparse.ArgumentParser`
+        The constructed argument parser.
+    """
+
+    parser = argparse.ArgumentParser(description="""
+    Certify a cp_pipe generated calibration as valid for a given date
+    range, and register them with the butler.
+    """)
+
+    parser.add_argument("root", help="Path to butler to use")
```
timj, line 117: URI to butler to use. It's not a path.

python/lsst/cp/pipe/cpCertify.py
```diff
@@ -69,19 +75,78 @@ def __init__(self, *, registry, inputCollection, outputCollection, lastRunOnly=T
...
+    parser.add_argument("inputCollection", help="Input collection to pull from.")
+    parser.add_argument("outputCollection", help="Output collection to add to.")
+    parser.add_argument("datasetTypeName", help="Dataset type to bless.")
+
+    parser.add_argument("-b", "--beginDate",
+                        help="Start date for using the calibration")
```
timj, line 123: What is the format of this date string? ISO8601? TAI vs UTC? Can it be MJD?  If you don't specify it what happens?

python/lsst/cp/pipe/cpCertify.py
```diff
@@ -69,19 +75,78 @@ def __init__(self, *, registry, inputCollection, outputCollection, lastRunOnly=T
...
+
+    parser = argparse.ArgumentParser(description="""
+    Certify a cp_pipe generated calibration as valid for a given date
+    range, and register them with the butler.
+    """)
+
+    parser.add_argument("root", help="Path to butler to use")
+    parser.add_argument("inputCollection", help="Input collection to pull from.")
+    parser.add_argument("outputCollection", help="Output collection to add to.")
+    parser.add_argument("datasetTypeName", help="Dataset type to bless.")
```
timj, line 120: certify not bless

python/lsst/cp/pipe/cpCertify.py
```diff
@@ -69,19 +75,78 @@ def __init__(self, *, registry, inputCollection, outputCollection, lastRunOnly=T
...
+    """Construct the argument parser for ``certifyCalibrations.py``.
+
+    Returns
+    -------
+    parser : `argparse.ArgumentParser`
+        The constructed argument parser.
+    """
+
+    parser = argparse.ArgumentParser(description="""
+    Certify a cp_pipe generated calibration as valid for a given date
```
timj, line 113: It doesn't seem like this command really requires it to be a cp_pipe generated calibration does it? Any calibration would work given that you only need a dataset type and input collection.

python/lsst/cp/pipe/cpCertify.py
```diff
@@ -69,19 +75,78 @@ def __init__(self, *, registry, inputCollection, outputCollection, lastRunOnly=T
...
+
         refs = set(self.registry.queryDatasets(datasetTypeName, collections=[self.inputCollection]))
         if not refs:
             raise RuntimeError(f"No inputs found for dataset {datasetTypeName} in {self.inputCollection}.")
         self.registry.registerCollection(self.outputCollection, type=CollectionType.CALIBRATION)
         self.registry.certify(self.outputCollection, refs, timespan)
+
+
+def build_parser():
+    """Construct the argument parser for ``certifyCalibrations.py``.
```
timj, line 104: We don't require a `.py` any longer.

python/lsst/cp/pipe/cpCertify.py
```diff
@@ -69,19 +75,78 @@ def __init__(self, *, registry, inputCollection, outputCollection, lastRunOnly=T
         self.inputCollection = inputCollection
         self.outputCollection = outputCollection
 
-    def run(self, datasetTypeName, timespan):
+    def run(self, datasetTypeName, beginDate=None, endDate=None):
```
timj, line 78: If you prefer run to take a Timespan it still can. You don't need to change the `run` method here -- it's allows for the `main()` below to convert the arguments to a time span.  It just depends on what you feel is best for this API.

timj (changes requested):
I don't think the parser and main functions should be here. I think we should follow convention and put them in a `lsst.cp.pipe.scripts` module.

I really don't understand why this isn't `butler certify-calibrations` -- there is no question that it's entirely a butler support routine. What's the motivation for not following the other butler commands?

I'd go further and say that I don't understand why it's even in cp_pipe.  It seems like completely generic butler code that doesn't need to know anything about cp_pipe. It's three methods on registry. It doesn't care about anything other than the existence of registry.

timj (approved):
Looks fine with me. Let me know when the daf_butler change needs reviewing.

PR #62, opened 2020-11-02
DM-25424: Convert Defect to IsrCalib

czwa:
Update Defect code to use the IsrCalib, and to work in gen3 as well as gen2.

### DIFF
--- a/pipelines/findDefects.yaml
+++ b/pipelines/findDefects.yaml
@@ -1,0 +1,33 @@
+description: cp_pipe CROSSTALK calibration construction.
+tasks:
+  isr:
+    class: lsst.ip.isr.isrTask.IsrTask
+    config:
+      connections.ccdExposure: 'raw'
+      connections.outputExposure: 'cpCrosstalkProc'
+      doWrite: True
+      doOverscan: True
+      doAssembleCcd: True
+      doBias: True
+      doVariance: False
+      doLinearize: False
+      doCrosstalk: False
+      doBrighterFatter: False
+      doDark: False
+      doStrayLight: False
+      doFlat: False
+      doFringe: False
+      doApplyGains: False
+      doDefect: False
+      doSaturationInterpolation: False
+      growSaturationFootprintSize: 0
+  measureDefects:
+    class: lsst.cp.pipe.defects.MeasureDefectsTask
+    config:
+      connections.inputExp: 'cpDefectsProc'
+      connections.outputRatios: 'cpPartialDefects'
+  mergeDefects:
+    class: lsst.cp.pipe.defects.MergeDefectsTask
+    config:
+      connections.inputRatios: 'cpPartialDefects'
+      connections.outputCrosstalk: 'crosstalk'
--- a/pipelines/findDefects.yaml
+++ b/pipelines/findDefects.yaml
@@ -1,4 +1,4 @@
-description: cp_pipe CROSSTALK calibration construction.
+description: cp_pipe DEFECT calibration construction.
 tasks:
   isr:
     class: lsst.ip.isr.isrTask.IsrTask
--- a/pipelines/findDefects.yaml
+++ b/pipelines/findDefects.yaml
@@ -4,7 +4,7 @@ tasks:
     class: lsst.ip.isr.isrTask.IsrTask
     config:
       connections.ccdExposure: 'raw'
-      connections.outputExposure: 'cpCrosstalkProc'
+      connections.outputExposure: 'cpDefectsProc'
       doWrite: True
       doOverscan: True
       doAssembleCcd: True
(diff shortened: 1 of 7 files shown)


pipelines/findDefects.yaml
```diff
@@ -0,0 +1,33 @@
+description: cp_pipe CROSSTALK calibration construction.
```
plazas, line 1: Change "CROSSTALK"

tests/test_defects.py
```diff
@@ -587,7 +533,7 @@ def test_pixelCounting(self):
 
         # Test the result of _nPixFromDefects()
         # via two different ways of calculating area.
-        self.assertEqual(defectArea, task._nPixFromDefects(defects))
+        # self.assertEqual(defectArea, task._nPixFromDefects(defects))
```
plazas, line 536: If we don't need this anymore (why?), then it should go, no?

tests/test_defects.py
```diff
@@ -617,23 +563,7 @@ def test_getNumGoodPixels(self):
...
-        self.assertEqual(countMaskedPixels(mi, 'EDGE'), 0)
-        self.defaultTask._setEdgeBits(mi)
-
-        hEdge = self.defaultConfig.nPixBorderLeftRight
-        vEdge = self.defaultConfig.nPixBorderUpDown
-        xSize, ySize = mi.getDimensions()
-
-        nEdge = xSize*vEdge*2 + ySize*hEdge*2 - hEdge*vEdge*4
-
-        self.assertEqual(countMaskedPixels(mi, 'EDGE'), nEdge)
```
plazas, line 636: Is this test not relevant anymore?

plazas (approved):

python/lsst/cp/pipe/defects.py
```diff
@@ -119,7 +85,7 @@ class FindDefectsTaskConfig(pexConfig.Config):
         dtype=float,
         doc=("Number of sigma below mean for dark pixel detection. The default value was found to be",
              " appropriate for some LSST sensors in DM-17490."),
-        default=5.0,
+        default=-5.0,
```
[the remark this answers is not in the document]
czwa: This is an algorithm change.  By having these sigma values retain their polarity, the code in `findHotAndColdPixels` doesn't need external knowledge to determine which way to apply the threshold.  The new code simply iterates over a list of sigma values, creates the threshold based on the sign and absolute value of the sigma value, and returns the merged set for that list of values.

I've added a block before the call to `findHotAndColdPixels` that limits the measurement on DARK frames to only search for bright pixels, while leaving both bright and dark thresholds for all other exposure types.  I originally included both for DARK frames, under the assumption that no dark pixels could be found, but this makes the current code closer to the previous version.

python/lsst/cp/pipe/defects.py
```diff
@@ -148,472 +114,104 @@ class FindDefectsTaskConfig(pexConfig.Config):
...
-    The task retrieves the corresponding dark and flat exposures for the
-    supplied visit. If a flat is available the task will (be able to) look
-    for both bright and dark defects. If only a dark is found then only bright
-    defects will be sought.
+    def validate(self):
+        super().validate()
+        if self.nSigmaBright < 0.0:
+            raise ValueError("nSigmaBright must be above 0.0.")
+        if self.nSigmaDark > 0.0:
+            raise ValueError("nSigmaDark must be below 0.0.")
```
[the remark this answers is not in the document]
czwa: As above.

python/lsst/cp/pipe/defects.py
```diff
@@ -148,472 +114,104 @@ class FindDefectsTaskConfig(pexConfig.Config):
...
             The exposure in which to find defects.
-        imageType : `str`
-            The image type, either 'dark' or 'flat'.
-        setMask : `bool`
-            If true, update exp with hot and cold pixels.
-            hot: DETECTED
-            cold: DETECTED_NEGATIVE
+        nSigma : `list [ `float` ]
+            Detection threshold to use.  Positive for DETECTED pixels,
+            negative for DETECTED_NEGATIVE pixels.
```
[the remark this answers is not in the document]
czwa: It probably is only useful in the case that the length is less than or equal to 2.  As noted above, I've added a case for DARK inputs frames.  I agree that longer lists will be redundant, but this pulls logic out of the `findHotAndColdPixels` method and puts it in `run`, so that method only needs to handle the thresholds given.  This seemed clearer to me, as from the `run` method, it's obvious what the `find` method will do.

tests/test_defects.py
czwa: Commented to avoid an error, now fixed and restored.

tests/test_defects.py
czwa: I think this was a more extreme case of commented to avoid errors during testing leading to deletion.  Restored as well.

PR #63, opened 2020-11-06
DM-27437: Have maxMeanSignal (and minMeanSignal) be a list per amp, instead of a single number in the PTC task

plazas:


python/lsst/cp/pipe/ptc.py
```diff
@@ -101,14 +101,16 @@ class MeasurePhotonTransferCurveTaskConfig(pexConfig.Config):
         default=1,
     )
     minMeanSignal = pexConfig.Field(
```
czwa, line 103: Can this use `pexConfig.DictField` instead?  This won't have a clean command line interface (I think you need to use a config file to set `DictField` entries), but it would put the type checking into the config object instead of forcing some later parsing.  The lack of a command line interface will also be mitigated by the switch to gen3.
An example of the config field: https://github.com/lsst/pipe_tasks/blob/master/python/lsst/pipe/tasks/ingest.py#L58
An example of setting the values: https://github.com/lsst/obs_lsst/blob/master/config/ucd/ingestCalibs.py#L14

python/lsst/cp/pipe/ptc.py
```diff
@@ -922,14 +948,11 @@ def errFunc(p, x, y):
             varVecOriginal = np.array(dataset.rawVars[ampName])
             varVecOriginal = self._makeZeroSafe(varVecOriginal)
 
-            mask = ((meanVecOriginal >= self.config.minMeanSignal) &
```
czwa, line 925: This mask is no longer needed because of the filtering on line 346?  Was this mask unnecessary in the previous version of the code?  I'm a bit confused why this isn't necessary (likely because I haven't read through the call stack again).

czwa (changes requested):

python/lsst/cp/pipe/ptc.py
plazas: Thanks, I'll try this.

python/lsst/cp/pipe/ptc.py
plazas: Right. Before, this was at this point (the EXPAPPROX and POLYNOMIAL path of the code). Than I moved it to 346, so now it applies to both paths of the code (FULLCOV being the other one). so, leaving it here would be redundant.

czwa (approved):
Took me a moment to find the default setting for min/maxMeanSigmaDict, but it looks fine now.

PR #64, opened 2020-11-06
DM-27438: Report in plots the latest point used in the PTC fit as the PTC turn off

plazas:


czwa (approved):
It might be good to move the text location to a different location to avoid overlaps, but looking over the descriptions of how to do that (https://matplotlib.org/tutorials/text/annotations.html) made it feel like it's probably more work than it's worth.

PR #67, opened 2020-12-01
DM-27783: Add improvements to getInitialGoodPoints in ptc.py

plazas:


python/lsst/cp/pipe/ptc.py
```diff
@@ -854,6 +870,14 @@ def _getInitialGoodPoints(means, variances, maxDeviationPositive, maxDeviationNe
             Maximum deviation from being constant for the variance/mean
             ratio, in the negative direction.
 
+        minMeanRatioTest : `float`
+            Minimum signal value (in ADU) after which to start examining
+            the ratios var/mean.
+
+       minVarPivotSearch : `float`
```
czwa, line 877: See below.

python/lsst/cp/pipe/ptc.py
```diff
@@ -889,14 +914,17 @@ def _getInitialGoodPoints(means, variances, maxDeviationPositive, maxDeviationNe
...
-        if len(pivot) == 0:
-            return goodPoints
-        else:
-            pivot = np.min(pivot)
-            goodPoints[pivot:len(goodPoints)] = False
-            return goodPoints
+        if len(pivot) > 0:
+            # For small values, sometimes the variance decreases slightly
+            # Only look when var > self.config.minVarPivotSearch
+            pivot = [p for p in pivot if variances[p] > minVarPivotSearch]
```
czwa, line 922: Just confirming that applying the minimum in the variance space is correct.  Cutting on the mean seems better to me, as this is a saturation effect, and we want to not cut for saturation before we think we're at that point.  This also feels odd because we're looking for a turnover but only if the turnover is above that minimum value.  I'm fine either way, so feel free to ignore this.

python/lsst/cp/pipe/ptc.py
```diff
@@ -1133,3 +1122,35 @@ def errFunc(p, x, y):
...
+        """Fill the dataset with NaNs if there are not enough good points.
+
+        Parameters
+        ----------
+        dataset : `lsst.ip.isr.ptcDataset.PhotonTransferCurveDataset`
+            The dataset containing the means, variances and exposure times.
+
+        ptcFitType : `str`
+            Fit a 'POLYNOMIAL' (degree: 'polynomialFitDegree') or
+            'EXPAPPROXIMATION' (Eq. 16 of Astier+19) to the PTC.
```
czwa, line 1136: Can this not happen with FULLCOVARIANCE?

tests/test_ptc.py
```diff
@@ -314,19 +314,19 @@ def test_meanVarMeasurementAllNan(self):
     def test_getInitialGoodPoints(self):
         xs = [1, 2, 3, 4, 5, 6]
         ys = [2*x for x in xs]
-        points = self.defaultTask._getInitialGoodPoints(xs, ys, 0.1, 0.25)
+        points = self.defaultTask._getInitialGoodPoints(xs, ys, 0.1, 0.25, 0., 0.)
```
czwa, line 317: Can these explicitly set the control parameters? `_getInitialGoodPoints(xs, ys, maxDeviationPositive=0.2, maxDeviationNegative=0.25, minMeanRatioTest=0.0, minVarPivotSearch=0.0)`?  This is easier to read and understand than a long list of floats.

czwa (approved):
Minor issues and questions.

python/lsst/cp/pipe/ptc.py
plazas: This helper function is used by `fitPtc`, which is used only by `POLYNOMIAL` or `EXPAPPROXIMATION`. It was written just to facilitate the reading of `fitPtc`.

python/lsst/cp/pipe/ptc.py
plazas: "This also feels odd because we're looking for a turnover but only if the turnover is above that minimum value"---> Right, this is what this `minVarPivotSearch` cut is meant to help with. 
In addition, at this point, the mean vector has been sorted, and the variance vector too using the indices of the mean vector. The code at this point assumes that the variance, after being sorted by the index of the sorted mean, increases monotonically.

PR #68, opened 2020-12-03
DM-27854: CpFlatNormalizationTask improperly indexes detectors

czwa:
Use detector indices instead of detector-scale indices.

### DIFF
--- a/python/lsst/cp/pipe/cpFlatNormTask.py
+++ b/python/lsst/cp/pipe/cpFlatNormTask.py
@@ -251,8 +251,8 @@ class CpFlatNormalizationTask(pipeBase.PipelineTask,
                 raise KeyError("Cannot find expected dimensions in %s" % (inDimensions, )) from e
 
             if self.config.level == 'DETECTOR':
-                detId = detMap[detectorId]
-                expId = expMap[exposureId]
+                detIdx = detMap[detectorId]
+                expIdx = expMap[exposureId]
                 try:
                     value = inMetadata.get('DETECTOR_MEDIAN')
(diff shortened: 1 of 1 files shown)


python/lsst/cp/pipe/cpFlatNormTask.py
```diff
@@ -294,15 +294,17 @@ def run(self, inputMDs, inputDims, camera):
 
         outputScales = defaultdict(lambda: defaultdict(lambda: defaultdict(lambda: defaultdict(float))))
 
+        # Note that the enumerated "detId"/"expId" here index the
+        # "detScales" and "expScales" arrays.
```
mrawls, line 298: This comment does enhance clarity, but I'd prefer not reusing `detId` and `expId` here since they were defined previously (lines 254-255 or lines 273-274). Perhaps `for detectorIdx, detector in enumerate(expSet)` and `for exposureIdx, exposure in enumerate(expSet)`?

mrawls (approved):
The change looks OK and should fix the problem.

Please consider renaming `detSet` to, e.g., `detectorsWithData` and `expSet` correspondingly to, e.g., `exposuresWithData`. Throughout, I do think it would read more clearly if you wrote out `exposure` and `detector` in the variable names, including `detectorScales` and `exposureScales`, etc., but I'm not going to insist on it everywhere.

python/lsst/cp/pipe/cpFlatNormTask.py
czwa: I've changed every variable in the `run` method that is a matrix index to have the `Idx` suffix, to avoid confusion with dataId identifiers.

PR #70, opened 2021-01-04
DM-28184: gen3 crosstalk generation does not set DET_NAME correctly.

czwa:
Fix bad metadata that wasn't fully tested.

### DIFF
--- a/python/lsst/cp/pipe/measureCrosstalk.py
+++ b/python/lsst/cp/pipe/measureCrosstalk.py
@@ -829,7 +829,14 @@ class MeasureCrosstalkTask(pipeBase.CmdLineTask):
             result = self.extract.run(exposure)
             ratios.append(result.outputRatios)
 
-        finalResults = self.solver.run(ratios, camera=camera)
+        for detIter, detector in enumerate(camera):
+            if detector.getName() == activeChip:
+                detectorId = detIter
+        outputDims = {'instrument': camera.getName(),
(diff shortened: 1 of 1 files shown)


python/lsst/cp/pipe/measureCrosstalk.py
```diff
@@ -827,7 +829,14 @@ def runDataRef(self, dataRefList):
...
 
-        finalResults = self.solver.run(ratios, camera=camera)
+        for detIter, detector in enumerate(camera):
+            if detector.getName() == activeChip:
+                detectorId = detIter
+        outputDims = {'instrument': camera.getName(),
+                      'detector': detectorId,
+                      }
+
+        finalResults = self.solver.run(ratios, camera=camera, outputDims=outputDims)
```
plazas, line 839: Why were the output dimensions not needed before, but now they are being passed?  What difference does it make?

plazas (approved):
Looks good. Maybe you can comment/clarify in the Jira ticket what was actually wrong?

python/lsst/cp/pipe/measureCrosstalk.py
czwa: I didn't think they were needed, because I had thought that the `calibChip` was the name, and so could be used to set the `_detectorName` attribute.  Constructing an `outputDims` for gen2 also reduces possible differences.

PR #73, opened 2021-01-27
DM-28457: In cpCombine.py, the config option "combine" should be a string, not the integer stats value

czwa:
Set the CalibCombineTask statistics properties by name.

### DIFF
--- a/python/lsst/cp/pipe/cpCombine.py
+++ b/python/lsst/cp/pipe/cpCombine.py
@@ -38,9 +38,9 @@ class CalibStatsConfig(pexConfig.Config):
     """Parameters controlling the measurement of background statistics.
     """
     stat = pexConfig.Field(
-        dtype=int,
-        default=int(afwMath.MEANCLIP),
-        doc="Statistic to use to estimate background (from lsst.afw.math)",
+        dtype=str,
+        default='MEANCLIP',
+        doc="Statistic name to use to estimate background (from lsst.afw.math)",
     )
     clip = pexConfig.Field(
         dtype=float,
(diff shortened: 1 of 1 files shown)


python/lsst/cp/pipe/cpCombine.py
```diff
@@ -201,9 +201,9 @@ class CalibCombineConfig(pipeBase.PipelineTaskConfig,
         doc="Mask planes to respect",
     )
     combine = pexConfig.Field(
-        dtype=int,
-        default=int(afwMath.MEANCLIP),
-        doc="Statistic to use for combination (from lsst.afw.math)",
+        dtype=str,
+        default='MEANCLIP',
+        doc="Statistic name to use for combination (from lsst.afw.math)",
```
plazas, line 206: I guess the only option for the user to know the possible options is to look at `Statistics.h`, no?

python/lsst/cp/pipe/ptc/cpSolvePtcTask.py
```diff
@@ -524,7 +524,7 @@ def fitPtc(self, dataset):
         if dataset.ptcFitType:
             ptcFitType = dataset.ptcFitType
         else:
-            raise RuntimeError(f"ptcFitType is None of empty in PTC dataset.")
+            raise RuntimeError("ptcFitType is None of empty in PTC dataset.")
```
plazas, line 527: Thanks for fixing this type. Can you fix the other typo ("None or empty" instead of "None of empty") before you merge?

plazas (approved):

python/lsst/cp/pipe/cpCombine.py
czwa: I think if this gets changed more than just to compare against eotest, then it probably should be made into a ChoiceField.

python/lsst/cp/pipe/ptc/cpSolvePtcTask.py
czwa: I think this was resolved on another ticket.  I fixed this on my branch to avoid a linter error, and it looks like someone else pushed first.

PR #74, opened 2021-01-29
DM-28531: imsim data needs different ptc pairing.

czwa:
Add exposureId based pairing for PTC.

This allows imsim data to be processed, as it has multiple flux levels
at the same exposure time.

### DIFF
--- a/python/lsst/cp/pipe/ptc/cpExtractPtcTask.py
+++ b/python/lsst/cp/pipe/ptc/cpExtractPtcTask.py
@@ -65,7 +65,7 @@ class PhotonTransferCurveExtractConfig(pipeBase.PipelineTaskConfig,
     """
     matchByExposureId = pexConfig.Field(
         dtype=bool,
-        doc="Should exposures by matched by ID rather than exp_time?",
+        doc="Should exposures by matched by ID rather than exposure time?",
         default=False,
     )
     maximumRangeCovariancesAstier = pexConfig.Field(
--- a/python/lsst/cp/pipe/ptc/cpExtractPtcTask.py
+++ b/python/lsst/cp/pipe/ptc/cpExtractPtcTask.py
@@ -24,7 +24,7 @@ import numpy as np
 import lsst.afw.math as afwMath
 import lsst.pex.config as pexConfig
 import lsst.pipe.base as pipeBase
-from lsst.cp.pipe.utils import arrangeFlatsByExpTime
+from lsst.cp.pipe.utils import arrangeFlatsByExpTime, arrangeFlatsByExpId
 
 import lsst.pipe.base.connectionTypes as cT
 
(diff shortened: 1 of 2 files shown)


python/lsst/cp/pipe/utils.py
```diff
@@ -540,6 +540,37 @@ def arrangeFlatsByExpTime(exposureList):
...
+    """Arrange exposures by exposure ID.
+
+    Parameters
+    ----------
+    exposureList : `list`[`lsst.afw.image.exposure.exposure.ExposureF`]
+        Input list of exposures.
+
+    Returns
+    ------
+    flatsAtExpTime : `dict` [`float`,
```
plazas, line 553: This seems copied from the other function. Please update, and provide an explanation on the pairs are grouped.  You are sorting by `expId`, and then group the exposures in pairs, correct? For each pair, do the exposures have the same flux? How do we know? Is it guaranteed that this is the case, just by sorting by exposure ID?

python/lsst/cp/pipe/utils.py
```diff
@@ -540,6 +540,37 @@ def arrangeFlatsByExpTime(exposureList):
...
+                      `list`[`lsst.afw.image.exposure.exposure.ExposureF`]]
+        Dictionary that groups flat-field exposures that have the same
+        exposure time (seconds).
+    """
+    flatsAtExpId = {}
+    sortedExposures = sorted(exposureList, key=lambda exp: exp.getInfo().getVisitInfo().getExposureId())
+
+    for jPair, exp in enumerate(sortedExposures):
```
plazas, line 561: Could you please a comment explaining a bit better what's happening here? I think I understand, but it would be easier to read.

python/lsst/cp/pipe/utils.py
```diff
@@ -540,6 +540,37 @@ def arrangeFlatsByExpTime(exposureList):
...
+        if (jPair + 1) % 2:
+            kPair = jPair // 2
+            listAtExpId = flatsAtExpId.setdefault(kPair, [])
+            listAtExpId.append(exp)
+            try:
+                listAtExpId.append(sortedExposures[jPair + 1])
```
plazas, line 567: Is this line here for when the number of exposures is not even?  If not, what happens when this is the case? or when there are more than two exposures with the same flux? What if there are 3 exposures with the same flux, and then another one with a different flux, so in the end you end up with a pair at the same flux (the first two), but then a pair with fluxes that don't match (exposures 3 and 4 in my example)?  Maybe you ready took this into account.

python/lsst/cp/pipe/ptc/cpExtractPtcTask.py
```diff
@@ -63,6 +63,11 @@ class PhotonTransferCurveExtractConfig(pipeBase.PipelineTaskConfig,
                                        pipelineConnections=PhotonTransferCurveExtractConnections):
     """Configuration for the measurement of covariances from flats.
     """
+    matchByExposureId = pexConfig.Field(
+        dtype=bool,
+        doc="Should exposures by matched by ID rather than exp_time?",
```
plazas, line 68: Should exp_time be spelled out?

python/lsst/cp/pipe/utils.py
czwa: I put it into a Notes section, as I couldn't figure out how to split it into comment sized chunks.

python/lsst/cp/pipe/utils.py
czwa: I've moved the first append into the try block.  This will resolve the case of an odd number of input exposures (as the second append will trigger the rollback of the first exposure as well, preventing partially populated pairs).  The case of bad inputs is not handled.  My solution here was mostly a way to make the best of the imSim situation, as I didn't think that simply sorting by mean would work.  Pathological cases like `flux=[1000, 1001, 1002, 1010]` where there might be better choices convinced me that using the simplest solution would be best.

plazas (approved):

PR #76, opened 2021-03-23
DM-29271: PTC task: Refactoring/simplification of code ported from Pierre Astier's repository. 

plazas:


python/lsst/cp/pipe/ptc/cpExtractPtcTask.py
```diff
@@ -449,18 +425,88 @@ def measureMeanVarCov(self, exposure1, exposure2, region=None, covAstierRealSpac
         wDiff = np.where(diffIm.getMask().getArray() == 0, 1, 0)
         w = w12*wDiff
 
-        if np.sum(w) < self.config.minNumberGoodPixelsForFft:
+        if np.sum(w) < self.config.minNumberGoodPixelsForFastFourierTransform:
```
czwa, line 428: Does this mean the real space/direct covariance calculation also can't be constructed?  If so, then could this be `minNumberPixelsForCovariance`?  Otherwise, it seems like this should warn, and then attempt to use the direct calculation.

python/lsst/cp/pipe/ptc/cpExtractPtcTask.py
```diff
@@ -449,18 +425,88 @@ def measureMeanVarCov(self, exposure1, exposure2, region=None, covAstierRealSpac
...
+            tempSize = np.array(np.log(s)/np.log(2.)).astype(int)
+            fftSize = np.array(2**(tempSize+1)).astype(int)
+            fftShape = (fftSize[0], fftSize[1])
+
+            c = CovFastFourierTransform(diffIm.image.array, w, fftShape, maxRangeCov)
+            covDiffAstier = c.reportCovFastFourierTransform(maxRangeCov)
 
         return mu, varDiff, covDiffAstier
+
+    def fillPtcDatasetAmp(self, ptcDataset, ampName, rawExpTime=[np.nan], rawMean=[np.nan], rawVar=[np.nan],
```
czwa, line 451: This method makes me think there maybe should be a method (or set of methods) on the PTC object to assist in construction.  This sets everything to single-element lists for each amplifier property, correct?  I think my idea is an `initAmp` method that accepts lists like this is doing, but that then checks that all the lengths for supplied values match.  This could then be used here as well as in `CpSolvePtcTask` when the dataset is filled with the collated results, and also even in the `PtcDataset.__init__`.  Maybe on the next pass at refactoring.

python/lsst/cp/pipe/ptc/cpExtractPtcTask.py
```diff
@@ -449,18 +425,88 @@ def measureMeanVarCov(self, exposure1, exposure2, region=None, covAstierRealSpac
...
+        ptcDataset.bMatrix[ampName] = np.full_like(covArray[0], np.nan)
+        ptcDataset.aMatrixNoB[ampName] = np.full_like(covArray[0], np.nan)
+        ptcDataset.ptcFitPars[ampName] = [np.nan]
+        ptcDataset.ptcFitParsError[ampName] = [np.nan]
+        ptcDataset.ptcFitChiSq[ampName] = np.nan
+        ptcDataset.finalVars[ampName] = [np.nan]
+        ptcDataset.finalModelVars[ampName] = [np.nan]
+        ptcDataset.finalMeans[ampName] = [np.nan]
+
+        return
```
czwa, line 512: You don't have to return if there's nothing expecting the `None` return value.

python/lsst/cp/pipe/ptc/astierCovPtcUtils.py
```diff
@@ -233,36 +227,7 @@ def covDirectValue(diffImage, weightImage, dx, dy):
...
-    """
-    def __init__(self):
-        self.r = 8
-        self.subtractDistantValue = False
-        self.start = 5
-        self.offsetDegree = 1
-
-
-def parseData(dataset, params):
+def parseData(dataset):
```
czwa, line 230: This function just creates the `CovFit` objects that are later fit, correct?  I'm unhappy with it returning a dict but calling the variable name a list, but that's a side point.  I'm not entirely sure I see the point of generating a dictionary of `CovFit` objects, then iterating over the items in that dict instead of just doing a normal loop over amplifiers stored in the dataset, instantiating a `CovFit` object, running the fit, storing the useful fit values in the dataset.  Are the `CovFit` objects used downstream in a way that I'm just not thinking about correctly?

czwa (commented):
I think I have too many questions to approve directly.  This definitely helps squash some of the unused code/odd single-call functions, but I'm still a bit confused on whether more could be done.

python/lsst/cp/pipe/ptc/cpExtractPtcTask.py
plazas: Right (to the first option). This `w` weight matrix is also an input of the function that calculates the covariance in real space, so I'll change the name to your suggestion.

python/lsst/cp/pipe/ptc/astierCovPtcUtils.py
plazas: Yes, the dictionary with the `CovFit` objects is then used in `cpSolvePtcTask` by `getOutputPtcDataCovAstier` when getting the output `dataset`.

python/lsst/cp/pipe/ptc/cpExtractPtcTask.py
plazas: I have placed the amp-filling function as part of the ptcDataset code in`ip_isr`.

czwa (approved):
I think this answers all of my concerns, thanks!

PR #78, opened 2021-04-01
DM-29272: PTC task: Validate that the variance calculation provided by awf and the C_00 entry of the covariance matrix produced via FFT using Astier's code is the same

plazas:


arunkannawadi (approved):
It looks good overall. I just have some minor suggestions which I commented inline.

PR #79, opened 2021-04-05
DM-24762: Add option to PTC task to correct for sigma clipping bias

plazas:


python/lsst/cp/pipe/ptc/cpExtractPtcTask.py
```diff
@@ -272,6 +273,11 @@ def run(self, inputExp, inputDims):
                 # in {maxLag, maxLag}^2]
                 muDiff, varDiff, covAstier = self.measureMeanVarCov(exp1, exp2, region=region,
                                                                     covAstierRealSpace=doRealSpace)
+                # Correction factor for sigma clipping. Function returns 1/sqrt(varFactor),
+                # so it needs to be squared. varDiff is calculated via afwMath.VARIANCECLIP.
+                varFactor = sigmaClipCorrection(self.config.nSigmaClipPtc)**2
```
mfisherlevine, line 278: I just want to check that L305 where you square things (as it was before) isn't re-squaring (_i.e._ because you're squaring the result here already).

Basically this line and the next are as they were before (I think) but it looks like L305 and L306 might be expecting it not to be squared, so just check that's all consistent.

Definitely not saying this is wrong, but just check this is really what you want.

plazas: We actually have two ways of getting the variance of the difference image: via `afwMath.VARIANCECLIP` and via the `[0,0]` element of the covariance calculation (Pierre’s code; L305, and L306).  So that’s why the squaring was done in two different places

PR #80, opened 2021-04-06
DM-29569: PTC maxMeanSignal no longer works 

plazas:


python/lsst/cp/pipe/ptc/cpSolvePtcTask.py
```diff
@@ -231,6 +231,10 @@ def run(self, inputCovariances, camera=None, inputExpList=None):
                     datasetPtc.rawVars[ampName].append(partialPtcDataset.rawVars[ampName][0])
                 else:
                     datasetPtc.rawVars[ampName].append(partialPtcDataset.rawVars[ampName])
+                if type(partialPtcDataset.expIdMask[ampName]) is list:
```
czwa, line 234: Are there any cases where these are not lists with a single element each?

python/lsst/cp/pipe/ptc/cpSolvePtcTask.py
```diff
@@ -607,7 +611,12 @@ def errFunc(p, x, y):
                 assert (len(mask) == len(timeVecOriginal) == len(meanVecOriginal) == len(varVecOriginal))
             if not (mask.any() and newMask.any()):
                 continue
-            dataset.expIdMask[ampName] = mask  # store the final mask
+            dataset.expIdMask[ampName] = np.array(dataset.expIdMask[ampName])
```
czwa, line 614: I'm sure the answer is "yes", but these are mask==1 is good, correct?

czwa (approved):

python/lsst/cp/pipe/ptc/cpSolvePtcTask.py
plazas: In theory, no. After running {{cpExtract}}, these should be lists with single elements.

python/lsst/cp/pipe/ptc/cpSolvePtcTask.py
plazas: More precisely, they have `True` (correct) and `False`

PR #82, opened 2021-04-06
DM-27650: Allow irlsFit function in cp_pipe/utils.py to use a general iterative weight function.

czwa:
Add alternate weight methods from reference to irlsFit.

### DIFF
--- a/python/lsst/cp/pipe/utils.py
+++ b/python/lsst/cp/pipe/utils.py
@@ -359,6 +359,11 @@ def irlsFit(initialParams, dataX, dataY, function, weightsY=None, weightType='Ca
     weightsY : `list` [`float`]
         Final weights used for each point.
 
+    Raises
+    ------
+    RuntimeError :
+        Raised if an unknown weightType string is passed.
+
     """
     if not weightsY:
         weightsY = np.ones_like(dataX)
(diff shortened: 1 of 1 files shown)


python/lsst/cp/pipe/utils.py
```diff
@@ -296,7 +296,7 @@ def makeDataRefList(self, namespace):
             self.refList += refList
 
 
-def irlsFit(initialParams, dataX, dataY, function, weightsY=None):
+def irlsFit(initialParams, dataX, dataY, function, weightsY=None, weightType='Cauchy'):
     """Iteratively reweighted least squares fit.
 
     This uses the `lsst.cp.pipe.utils.fitLeastSq`, but applies
```
plazas, line 302: Do you need to update the docstring a bit? (e.g., "applies Cauchy by default, but there are more options, see below" or something like that).

python/lsst/cp/pipe/utils.py
```diff
@@ -315,6 +315,8 @@ def irlsFit(initialParams, dataX, dataY, function, weightsY=None):
         Function to fit.
     weightsY : `numpy.array` [`float`]
         Weights to apply to the data.
+    weightType : `str`, optional
```
plazas, line 318: Should you say here (or above) that the list of options for weights is explained in the body of the function? or do you formally need to exhaustively list all the options here in the docstring?

plazas (approved):
It looks good. The only question is, why did you decide all of a sudden that we needed to have all these options for the weights? Have you found recently situations in which Cauchy is not enough, or do we anticipate in the future to have to use each one of these weights in different situations, depending on how optimally they behave?

python/lsst/cp/pipe/utils.py
czwa: I've listed all the options here, because it seemed reasonable.  I've also added a raise if an unknown option is passed.

PR #85, opened 2021-04-21
DM-29699: CalibCombineConnections changes its quantum dimensions at construction

czwa:
Use separate classes for the cpCombine step when it does and does not use the 'physical_filter' dimension.

The new CalibCombineByFilter Connections, Config, and Task now inherit
from their respective CalibCombine counterparts, but simply add the
new dimension.  This avoids the previous work around to have the
dimensions edited on construction.

Update the pipeline definitions for this as well.

### DIFF
--- a/pipelines/cpFlat.yaml
+++ b/pipelines/cpFlat.yaml
@@ -28,13 +28,12 @@ tasks:
       connections.inputMDs: 'flatStats'
       connections.outputScales: 'cpFlatNormScales'
   cpFlatCombine:
-    class: lsst.cp.pipe.cpCombine.CalibCombineTask
+    class: lsst.cp.pipe.cpCombine.CalibCombineByFilterTask
     config:
       connections.inputExps: 'cpFlatProc'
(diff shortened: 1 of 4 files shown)


mrawls (commented):
This seems like a good simplification. Two questions. (1) It's fine to have `CalibCombineByFilterTask` be the default, but what would the vanilla `CalibCombineTask` ever be used for? (2) What would happen if someone did try to run a pipeline that still explicitly gave `calibrationDimensions`, and can a useful error be raised?

czwa:
`CalibCombineTask` is still used for bias and dark creation, as those should have no illumination at all (and so can ignore filters).  If an prior pipeline were run with the new processing code, there'd be a dimension mismatch and an error.  If I'm thinking about it correctly, that error would be raised while constructing the pipeline quantum graph, and so adding a message in the `cp_pipe` code wouldn't help.

mrawls (approved):
Bracing myself for people to ask about qgraph dimension errors in a month or so when their stack versions catch up with this change 😁 Looks great otherwise!

PR #86, opened 2021-04-29
DM-29877: Avoid avoidable logic error in measureCrosstalk

czwa:
Fix assumption that the ID and Name match for all detectors.

### DIFF
--- a/python/lsst/cp/pipe/measureCrosstalk.py
+++ b/python/lsst/cp/pipe/measureCrosstalk.py
@@ -480,7 +480,8 @@ class CrosstalkSolveTask(pipeBase.PipelineTask,
         for ratioDict, fluxDict in zip(inputRatios, inputFluxes):
             for targetChip in ratioDict:
                 if calibChip and targetChip != calibChip and targetChip != calibDetector.getName():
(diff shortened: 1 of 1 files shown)


python/lsst/cp/pipe/measureCrosstalk.py
```diff
@@ -474,7 +479,7 @@ def run(self, inputRatios, inputFluxes=None, camera=None, inputDims=None, output
         combinedFluxes = defaultdict(lambda: defaultdict(list))
         for ratioDict, fluxDict in zip(inputRatios, inputFluxes):
             for targetChip in ratioDict:
-                if calibChip and targetChip != calibChip:
+                if calibChip and targetChip != calibChip and targetChip != calibDetector.getName():
                     raise RuntimeError("Received multiple target chips!")
```
timj, line 483: Can that error message give some feedback as to which of the three checks failed?

timj (approved):

python/lsst/cp/pipe/measureCrosstalk.py
czwa: Updated to be clearer.

PR #87, opened 2021-05-05
DM-29987: Save Detectors with master calibrations

czwa:
Add the detector to the outputs of cpCombine.

### DIFF
--- a/python/lsst/cp/pipe/cpCombine.py
+++ b/python/lsst/cp/pipe/cpCombine.py
@@ -316,6 +316,10 @@ class CalibCombineTask(pipeBase.PipelineTask,
         self.combineHeaders(inputExps, combinedExp,
                             calibType=self.config.calibrationType, scales=expScales)
 
+        # Set the detector
+        inputDetector = inputExps[0].getDetector()
+        combinedExp.setDetector(inputDetector)
+
         # Return
         return pipeBase.Struct(
             outputData=combinedExp,
--- a/python/lsst/cp/pipe/cpCombine.py
+++ b/python/lsst/cp/pipe/cpCombine.py
@@ -254,6 +254,13 @@ class CalibCombineTask(pipeBase.PipelineTask,
         if numExps < self.config.maxVisitsToCalcErrorFromInputVariance:
             stats.setCalcErrorFromInputVariance(True)
 
+        detectorList = [exp.getDetector() for exp in inputExps]
+        if None in detectorList:
+            self.log.warn("Not all input detectors defined.")
+        numDetectors = len(set([det.getId() for det in detectorList if det is not None]))
+        if numDetectors > 1:
+            raise RuntimeError("Input data contains multiple detectors.")
+
         # Create output exposure for combined data.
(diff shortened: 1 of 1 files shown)


timj (approved):

python/lsst/cp/pipe/cpCombine.py
```diff
@@ -281,6 +281,13 @@ def run(self, inputExps, inputScales=None, inputDims=None):
         if numExps < self.config.maxVisitsToCalcErrorFromInputVariance:
             stats.setCalcErrorFromInputVariance(True)
 
+        detectorList = [exp.getDetector() for exp in inputExps]
+        if None in detectorList:
+            self.log.warn("Not all input detectors defined.")
```
timj, line 286: Doesn't this warning imply that `inputExps[0].getDetector()` called below could return `None`? I think this means that you need to at least extract the first not-None from this list so you can use it below rather than picking the first input.

python/lsst/cp/pipe/cpCombine.py
```diff
@@ -343,6 +350,10 @@ def run(self, inputExps, inputScales=None, inputDims=None):
         self.combineHeaders(inputExps, combinedExp,
                             calibType=self.config.calibrationType, scales=expScales)
 
+        # Set the detector
+        inputDetector = inputExps[0].getDetector()
```
timj, line 354: I think this calculation should be moved to the earlier logic where you build the list and work out which one is not-none.

python/lsst/cp/pipe/cpCombine.py
```diff
@@ -281,6 +281,13 @@ def run(self, inputExps, inputScales=None, inputDims=None):
         if numExps < self.config.maxVisitsToCalcErrorFromInputVariance:
             stats.setCalcErrorFromInputVariance(True)
 
+        detectorList = [exp.getDetector() for exp in inputExps]
+        if None in detectorList:
+            self.log.warn("Not all input detectors defined.")
+        numDetectors = len(set([det.getId() for det in detectorList if det is not None]))
+        if numDetectors > 1:
+            raise RuntimeError("Input data contains multiple detectors.")
```
timj, line 289: Presumably it's impossible for us to tell from a detector whether the detector itself is different even if the ID is the same? Do you care if the serial is different?

timj (approved):

python/lsst/cp/pipe/cpCombine.py
czwa: I've reworked this bit of code to pass `None` into the list of detectorIds, and changed the criterion to raise if the set length is not equal to 1.  This allows for a case in which no detectors exist (which is a valid possibility), as well as checking that all inputs have the same detector if it exists.

python/lsst/cp/pipe/cpCombine.py
czwa: I've written this comment arguing both ways (ids vs serials), and have finally just changed the code to check that the sets of both are the same length, and that that length is 1.

PR #89, opened 2021-05-12
DM-30130: Establish a 1-1 correspondence between exposures and input dimensions in cpPtcExtract

plazas:


python/lsst/cp/pipe/ptc/cpExtractPtcTask.py
```diff
@@ -201,30 +201,47 @@ def runQuantum(self, butlerQC, inputRefs, outputRefs):
...
         """
         inputs = butlerQC.get(inputRefs)
+        # Ids of input list of exposures
+        inputs['inputDims'] = [expId.dataId['exposure'] for expId in inputRefs.inputExp]
+
+        # Define a 1-1 correspondence between exp.getInfo().getVisitInfo().getExposureId()
+        # and inputs['inputDims']. This will be used in `run` when saving partial datasets.
+        expIdToInputDim = {}
+        for exp, dim in zip(inputs['inputExp'], inputs['inputDims']):
+            expId = exp.getInfo().getVisitInfo().getExposureId()
```
czwa, line 211: I think we want to get away from using `getInfo().getVisitInfo().getExposureId()`.  Is it possible to just pass the inputDims to the arrange functions, and have them be sorted as well?  This would future-proof this part of the code in case the info/VisitInfo objects change.  This would also remove the need to pass a third mapping object to the run method.

python/lsst/cp/pipe/ptc/cpExtractPtcTask.py
```diff
@@ -343,22 +360,12 @@ def run(self, inputExp, inputDims):
...
             # Use location of exp1 to save PTC dataset from (exp1, exp2) pair.
-            # expId1 and expId2, as returned by getInfo().getVisitInfo().getExposureId(),
-            # and the exposure IDs stured in inoutDims,
-            # may have the zero-padded detector number appended at
-            # the end (in gen3). A temporary fix is to consider expId//1000 and/or
-            # inputDims//1000.
-            # Below, np.where(expId1 == np.array(inputDims)) (and the other analogous
-            # comparisons) returns a tuple with a single-element array, so [0][0]
+            # Below, np.where(expId1 == np.array(inputDims))  returns a tuple
+            # with a single-element array, so [0][0]
```
czwa, line 364: This logic would also be replaced by simply pulling the exposure values at the same time as the exposures in the loop above.

czwa (changes requested):
Let me know if this suggested change makes sense.

python/lsst/cp/pipe/ptc/cpExtractPtcTask.py
timj: The DataId will have the detector number and exposure ID in it already. This is not the same thing as getExposureId because getExposureId is the detector exposure ID not the exposure ID. We are actively discussing changing `getExposureId` so it would be best not to use it here.  I think for grouping you are better off creating a tuple of exposure+detector -- then you are being explicit in how you are doing your grouping using gen3 concepts and not relying on the visitInfo at all.

timj: If I'm not clear, I mean that you use the tuple of the exposure and detector dataId values as the key in your grouping algorithm.

python/lsst/cp/pipe/ptc/cpExtractPtcTask.py
```diff
@@ -201,34 +201,36 @@ def runQuantum(self, butlerQC, inputRefs, outputRefs):
...
+        if self.config.matchByExposureId:
+            inputs['inputExp'] = arrangeFlatsByExpId(inputs['inputExp'], inputs['inputDims'])
+        else:
+            inputs['inputExp'] = arrangeFlatsByExpTime(inputs['inputExp'], inputs['inputDims'])
+
         outputs = self.run(**inputs)
         butlerQC.put(outputs, outputRefs)
 
-    def run(self, inputExp, inputDims):
+    def run(self, inputDims, inputExp):
```
czwa, line 216: This switch seems unnecessary now.

czwa (approved):
This looks like what I was thinking.  This will hopefully avoid id confusion.

PR #91, opened 2021-06-01
DM-30257: Remove eotest dependency from cp_pipe

czwa:
Remove runEotestTask.


mfisherlevine (approved):
The change, as it is, looks totally fine.

The only thing I do wonder is whether we should just leave the file as-is, and either remove the warning altogether, or shove it down to DEBUG or TRACE level, but I guess we want to reduce cruft, and should we ever decide to resurrect this, we can always just pull it out of the git history, so yeah, on balance I think this is good to go.

PR #92, opened 2021-06-04
DM-29338: Incorporate Gen 3 crosstalk in HiTS runs

mrawls:
This PR adds a DECam flat-building pipeline, and necessarily standardizes ISR task names so yaml imports work.


isullivan (approved):
It is unfortunate that "calibration products" and "community pipeline" both get shortened to "cp" for DECam, but that is beyond the scope of this ticket.

PR #94, opened 2021-07-14
DM-31096: Fix bitrot in gen3 linearity solver

czwa:
This largely duplicates the gen3 connections in the brighter-fatter code (which also works off of the PTC dataset).
Additional handlers were added to ensure that the length of the output fields match between solvable and unsolveable amplifiers.  If the lengths do not match, an error is thrown by the FITS writing code.

### DIFF
--- a/pipelines/cpLinearitySolve.yaml
+++ b/pipelines/cpLinearitySolve.yaml
@@ -1,0 +1,7 @@
+description: cp_pipe linearity calibration construction.
+tasks:
+  linearitySolve:
+    class: lsst.cp.pipe.LinearitySolveTask
+    config:
+      connections.inputPtc: ptc
+      connections.outputLinearizer: linearity
--- a/python/lsst/cp/pipe/linearity.py
+++ b/python/lsst/cp/pipe/linearity.py
@@ -31,26 +31,37 @@ from lsstDebug import getDebugFrame
 from lsst.ip.isr import (Linearizer, IsrProvenance)
 
 from .utils import (funcPolynomial, irlsFit)
-
+from ._lookupStaticCalibration import lookupStaticCalibration
 
 __all__ = ["LinearitySolveTask", "LinearitySolveConfig", "MeasureLinearityTask"]
 
 
 class LinearitySolveConnections(pipeBase.PipelineTaskConnections,
-                                dimensions=("instrument", "detector")):
-    inputPtc = cT.Input(
-        name="inputPtc",
-        doc="Input PTC dataset.",
-        storageClass="StructuredDataDict",
-        dimensions=("instrument", "detector"),
-        multiple=False,
+                                dimensions=("instrument", "exposure", "detector")):
+    dummy = cT.Input(
+        name="raw",
+        doc="Dummy exposure.",
+        storageClass='Exposure',
+        dimensions=("instrument", "exposure", "detector"),
+        multiple=True,
+        deferLoad=True,
     )
-    camera = cT.Input(
+    camera = cT.PrerequisiteInput(
         name="camera",
         doc="Camera Geometry definition.",
         storageClass="Camera",
         dimensions=("instrument", ),
+        isCalibration=True,
+        lookupFunction=lookupStaticCalibration,
     )
+    inputPtc = cT.PrerequisiteInput(
+        name="ptc",
+        doc="Input PTC dataset.",
+        storageClass="PhotonTransferCurveDataset",
+        dimensions=("instrument", "detector"),
+        isCalibration=True,
+    )
+
     outputLinearizer = cT.Output(
         name="linearity",
         doc="Output linearity measurements.",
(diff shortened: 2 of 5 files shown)


python/lsst/cp/pipe/linearity.py
```diff
@@ -31,26 +31,37 @@
...
-    camera = cT.Input(
+    camera = cT.PrerequisiteInput(
         name="camera",
         doc="Camera Geometry definition.",
         storageClass="Camera",
         dimensions=("instrument", ),
+        isCalibration=True,
+        lookupFunction=lookupStaticCalibration,
     )
+    inputPtc = cT.PrerequisiteInput(
```
plazas, line 57: Before, this task wasn't using as input the output of the PTC?

python/lsst/cp/pipe/linearity.py
```diff
@@ -106,6 +117,11 @@ class LinearitySolveConfig(pipeBase.PipelineTaskConfig,
         doc="Maximum deviation from linear solution for Poissonian noise.",
         default=5.0,
     )
+    ignorePtcMask = pexConfig.Field(
```
plazas, line 120: Did you find cases in which it was useful to ignore the values masked by the PTC task? Also, maybe explain a bit more in the docstring, if possible, to which values you are referring to (I assume points at a certain mean signal level in ADU).

plazas (approved):
I just requested I few clarifications.

python/lsst/cp/pipe/linearity.py
```diff
@@ -129,12 +140,12 @@ def runQuantum(self, butlerQC, inputRefs, outputRefs):
...
 
         # Use the dimensions to set calib/provenance information.
-        inputs['inputDims'] = [exp.dataId.byName() for exp in inputRefs.inputPtc]
+        inputs['inputDims'] = inputRefs.inputPtc.dataId.byName()
 
         outputs = self.run(**inputs)
         butlerQC.put(outputs, outputRefs)
 
-    def run(self, inputPtc, camera, inputDims):
+    def run(self, inputPtc, dummy, camera, inputDims):
```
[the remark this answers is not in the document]
czwa: Added docstring.  The answer is that the Butler doesn't know how to retrieve the PTC without having something to look it up.  The dummy exposure allows Butler to know which PTC it should retrieve, but isn't needed for the linearity calculation.  This is the same trick that `makeBrighterFatterKernel.py` does (and I've spotted a typo there, and will duplicate the new docstring which is clearer than the old one).

pipelines/cpLinearitySolve.yaml
```diff
@@ -0,0 +1,7 @@
+description: cp_pipe linearity calibration construction.
+tasks:
+  linearitySolve:
+    class: lsst.cp.pipe.LinearitySolveTask
+    config:
+      connections.inputPtc: ptc
```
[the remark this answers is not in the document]
czwa: This has to match the `dataset_type_name` that is stored in the Butler.  It can technically be anything, as long as it's consistent between the PTC solve code and the downstream linearity/bfk code.  This value matches what we have in https://github.com/lsst/cp_pipe/blob/master/pipelines/measurePhotonTransferCurve.yaml#L39 so I'm hesitant to change it.

python/lsst/cp/pipe/linearity.py
```diff
@@ -31,26 +31,37 @@
...
 class LinearitySolveConnections(pipeBase.PipelineTaskConnections,
-                                dimensions=("instrument", "detector")):
-    inputPtc = cT.Input(
-        name="inputPtc",
-        doc="Input PTC dataset.",
-        storageClass="StructuredDataDict",
-        dimensions=("instrument", "detector"),
-        multiple=False,
+                                dimensions=("instrument", "exposure", "detector")):
+    dummy = cT.Input(
```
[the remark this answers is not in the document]
czwa: As before, this is here to enable a database JOIN in butler to work.  As I discovered yesterday while debugging this, not doing this results in "Cannot query dataset type 'ptc' in CALIBRATION collection 'u/x/y/z'" (or some similarly worded error message).

python/lsst/cp/pipe/linearity.py
czwa: It was, but it wasn't a `PrerequisiteInput`.  I also rearranged the connections in the class to match how it was in `makeBrighterFatterKernel.py`.

python/lsst/cp/pipe/linearity.py
czwa: For my test PTC, it flagged a number of points that looked reasonable between ~20k and ~50k ADU.  Even if these points don't match the expectation of the PTC code, they're still valid measurements of the flux at that particular exposure time.  Only once the detector saturates are the values no longer useful for linearity measurement.  With this option and the `maxLinearAdu` option, I was able to use all of the points less than the apparent saturation.

PR #101, opened 2021-09-24
DM-31809: length of mask and means can disagree in BrighterFatterKernelSolveTask

czwa:
Fix mask error on input PTC data.

Switching to using a mask based on the finalMeans removes both the bad
points (which are excluded from finalMeans) and the NaN padded entries
(needed to ensure the FITS table representation is a standard array).


plazas (approved):
Thank you for this fix.  Did you test it for all three types of fits in the PTC code (POLYNOMIAL, EXPAPPROXIMATION, FULLCOVARIANCE)?

PR #102, opened 2021-09-28
DM-31306: Create single SAL script that produces biases, darks, and flats

plazas:


czwa (approved):
We're using lower-case booleans in all pipeline yaml definitions for consistency.  Other than that, as long as a test confirms it does what you want, it looks fine to me.

PR #103, opened 2021-10-07
DM-26428: Bin images after difference image of flat pairs has been created (not before) in PTC task.

plazas:


python/lsst/cp/pipe/ptc/cpSolvePtcTask.py
```diff
@@ -595,8 +600,11 @@ def errFunc(p, x, y):
                 parsIniPtc = [-1e-9, 1.0, 10.]  # a00, gain, noisei^2
                 # lowers and uppers obtained from BOT data studies by
                 # C. Lage (UC Davis, 11/2020).
-                bounds = self._boundsForAstier(parsIniPtc, lowers=[-1e-4, 0.5, -2000],
-                                               uppers=[1e-4, 2.5, 2000])
+                if self.config.binSize > 1:
```
czwa, line 603: Is this saying that binned data does not have any restrictions on the bounds?  Does that explain why the flux cutoffs on the figures attached to the jira ticket are so much lower than on the default un-binned case?

czwa (approved):

PR #104, opened 2021-10-08
DM-31801: Clean up log usage

timj:
* Use python logging.
* Fix a lot of log usage issues.
* Ensure that `lsst.` is used as a logger prefix.

### DIFF
--- a/python/lsst/cp/pipe/ptc/cpExtractPtcTask.py
+++ b/python/lsst/cp/pipe/ptc/cpExtractPtcTask.py
@@ -277,14 +277,14 @@ class PhotonTransferCurveExtractTask(pipeBase.PipelineTask,
 
         if self.config.numEdgeSuspect > 0:
             isrTask = IsrTask()
-            self.log.info(f"Masking {self.config.numEdgeSuspect} pixels from the edges "
-                          "of all exposures as SUSPECT.")
+            self.log.info("Masking %d pixels from the edges of all exposures as SUSPECT.",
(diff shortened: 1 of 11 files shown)


python/lsst/cp/pipe/ptc/cpExtractPtcTask.py
```diff
@@ -277,24 +277,23 @@ def run(self, inputExp, inputDims):
 
         if self.config.numEdgeSuspect > 0:
             isrTask = IsrTask()
-            self.log.info(f"Masking {self.config.numEdgeSuspect} pixels from the edges "
-                          "of all exposures as SUSPECT.")
+            self.log.info("Masking %d pixels from the edges of all exposures as SUSPECT.",
```
plazas, line 280: I thought it was preferred to use f-strings. Isn't it the case?

plazas (approved):
Looks good. I only had a clarifying question regarding switching to not using f-strings in the log messages.

python/lsst/cp/pipe/ptc/cpExtractPtcTask.py
timj: You aren't allowed to use f-strings in log messages. Log messages are designed to deliberately defer the string formatting until it is known that the message has to be output. If you have warning log messages only then this info message will not appear and the message will never be formatted. With an f-string it always gets formatted.

czwa (approved):

PR #108, opened 2021-11-02
DM-13943: Deprecate VisitInfo.getExposureId()

kfindeisen:
This PR removes uses of `VisitInfo.getExposureId()`, which was deprecated in lsst/afw#614.

### DIFF
--- a/python/lsst/cp/pipe/ptc/measurePtcGen2Task.py
+++ b/python/lsst/cp/pipe/ptc/measurePtcGen2Task.py
@@ -124,7 +124,7 @@ class MeasurePhotonTransferCurveTask(pipeBase.CmdLineTask):
                 self.log.warn("postISR exposure could not be retrieved. Ignoring flat.")
                 continue
             expList.append(tempFlat)
-        expIds = [exp.getInfo().getVisitInfo().getExposureId() for exp in expList]
+        expIds = [exp.info.getVisitInfo().id for exp in expList]
 
         # Create dictionary of exposures, keyed by exposure time
         expDict = arrangeFlatsByExpTime(expList)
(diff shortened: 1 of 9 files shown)


python/lsst/cp/pipe/ptc/measurePtcGen2Task.py
```diff
@@ -124,7 +124,7 @@ def runDataRef(self, dataRefList):
                 self.log.warn("postISR exposure could not be retrieved. Ignoring flat.")
                 continue
             expList.append(tempFlat)
-        expIds = [exp.getInfo().getVisitInfo().getExposureId() for exp in expList]
+        expIds = [exp.info.getVisitInfo().id for exp in expList]
```
spenczar, line 127: Should this be `exp.info.id for exp in expList`?

python/lsst/cp/pipe/utils.py
```diff
@@ -688,7 +688,7 @@ def arrangeFlatsByExpId(exposureList, exposureIdList):
     """
     flatsAtExpId = {}
     # sortedExposures = sorted(exposureList, key=lambda exp:
-    # exp.getInfo().getVisitInfo().getExposureId())
+    # exp.info.getVisitInfo().id)
```
spenczar, line 691: What's going on here?

spenczar (changes requested):

python/lsst/cp/pipe/ptc/measurePtcGen2Task.py
kfindeisen: Oops, I could swear I included an explanation in the commit message (I'll go back and add one).

@czwa told me that these two identifiers were supposed to be the visit ID and not the per-detector ID, so basically the use of `getExposureId()` here was a bug.

python/lsst/cp/pipe/utils.py
czwa: The comment was written too long, and was split when comment line length linting was enabled (part of DM-30245).  It should probably be removed, as it's no longer useful.

spenczar (approved):

PR #111, opened 2021-11-17
DM-32621: ci_cpp_gen3 fails with TypeError

czwa:
Fix type errors in log messages.


plazas (approved):
Looks good, thanks for catching this.

PR #112, opened 2021-11-17
DM-31896: Clarify PTC code

plazas:


python/lsst/cp/pipe/utils.py
```diff
@@ -710,6 +711,415 @@ def arrangeFlatsByExpId(exposureList, exposureIdList):
...
+    cov = np.ndarray((len(muVals), np.max(i)+1, np.max(j)+1))
+    var = np.zeros_like(cov)
+    cov[ind, i, j] = c
+    var[ind, i, j] = v**2/n
+    var[:, 0, 0] *= 2  # var(v) = 2*v**2/N
+
+    return cov, var, muVals
+
+
+def symmetrize(inputArray):
```
czwa, line 1097: This duplicates/is duplicated by BrighterFatterKernelSolveTask._tileArray().  I've added DM-33353 to unify these methods.

python/lsst/cp/pipe/utils.py
```diff
@@ -710,6 +711,415 @@ def arrangeFlatsByExpId(exposureList, exposureIdList):
...
+        covariance = covariance[maskFromWeights]
+
+    if divideByMu:
+        covariance /= mu
+        covarianceModel /= mu
+        weights *= mu
+    return mu, covariance, covarianceModel, weights, maskFromWeights
+
+
+def makeCovArray(inputTuple, maxRangeFromTuple=8):
```
czwa, line 1003: Is this ever called outside of PhotonTransferCurveExtractTask?  I have the same question with `computeCovDirect` above.

python/lsst/cp/pipe/utils.py
```diff
@@ -710,6 +711,415 @@ def arrangeFlatsByExpId(exposureList, exposureIdList):
...
+        for dy in range(maxRange+1):
+            for dx in range(maxRange+1):
+                cov, npix = self.cov(dx, dy)
+                if (dx == 0 and dy == 0):
+                    var = cov
+                tupleVec.append((dx, dy, var, cov, npix))
+        return tupleVec
+
+
+def computeCovDirect(diffImage, weightImage, maxRange):
```
czwa, line 810: Is there a reason to keep both the direct and FFT convolutions?  If they give the same result, we should just implement one.  I see we use the FFT by default, so maybe keeping that one?  However, we tend to use lags of 8, which suggests the direct method should be faster (it's also simpler).

python/lsst/cp/pipe/ptc/cpSolvePtcTask.py
```diff
@@ -273,101 +299,345 @@ def fitCovariancesAstier(self, dataset):
...
+                          'fullModelNoB': {'a': [], 'c': [], 'noise': [], 'gain': [], 'paramsErr': []}}
+            for key in functionsDict:
+                params, paramsErr, _ = fitLeastSq(pInit, muAtAmp,
+                                                  covAtAmp.flatten(), functionsDict[key],
+                                                  weightsY=covSqrtWeightsAtAmp.flatten())
+                a = params[:lenParams].reshape((matrixSide, matrixSide))
+                c = params[lenParams:2*lenParams].reshape((matrixSide, matrixSide))
+                noise = params[2*lenParams:3*lenParams].reshape((matrixSide, matrixSide))
+                gain = params[-1]
+                if key == 'fullModel':
```
czwa, line 419: Can't this just be `fitResults[key]['a'] = a`, and avoid this if/else?

czwa (commented):
I'm hoping we can still simplify some more by removing the fft/direct choice, and keeping functions that are only called by one task as task methods.

python/lsst/cp/pipe/utils.py
plazas: I kept the FFT version.

python/lsst/cp/pipe/utils.py
plazas: I think it's only called by the extract class, so I moved it there.

python/lsst/cp/pipe/ptc/cpExtractPtcTask.py
```diff
@@ -316,64 +322,81 @@ def run(self, inputExp, inputDims):
...
                                         self.config.maximumRangeCovariancesAstier), np.nan)
                     covSqrtWeights = np.full_like(covArray, np.nan)
 
+                # Mask data point if it is outside of the
+                # specified mean signal range.
                 if (muDiff <= minMeanSignalDict[ampName]) or (muDiff >= maxMeanSignalDict[ampName]):
                     expIdMask = False
 
                 if covAstier is not None:
+                    # Turn the tuples with teh measured information
```
czwa, line 364: typo

python/lsst/cp/pipe/ptc/cpExtractPtcTask.py
```diff
@@ -502,20 +616,17 @@ def measureMeanVarCov(self, exposure1, exposure2, region=None, covAstierRealSpac
...
-            # Calculate the sizes of FFT dimensions.
-            s = shapeDiff + maxRangeCov
-            tempSize = np.array(np.log(s)/np.log(2.)).astype(int)
-            fftSize = np.array(2**(tempSize+1)).astype(int)
-            fftShape = (fftSize[0], fftSize[1])
-
-            c = CovFastFourierTransform(diffIm.image.array, w, fftShape, maxRangeCov)
-            covDiffAstier = c.reportCovFastFourierTransform(maxRangeCov)
+
+        # Calculate covariances via FFT (default).
```
czwa, line 620: Not the default anymore, as we only have the one option.

czwa (approved):
Looks good.  A few minor typos are all I could see.

PR #113, opened 2021-12-01
DM-32670: Running ISR and PTC task on BOT data on 189 CCDs fails at NCSA

plazas:


python/lsst/cp/pipe/utils.py
```diff
@@ -631,25 +631,27 @@ def arrangeFlatsByExpTime(exposureList, exposureIdList):
...
     Returns
     ------
     flatsAtExpTime : `dict` [`float`,
-                      `list`[(`lsst.afw.image.ExposureF`, `int`)]]
-        Dictionary that groups flat-field exposures (and their IDs) that have
+                      `list`[(`lsst.pipe.base.connections.DeferredDatasetRef`,
+                              `int`)]]
+        Dictionary that groups references to flat-field exposures
+        (and their IDs) that have
         the same exposure time (seconds).
```
mfisherlevine, line 647: Maybe put this line on the line above now it's short enough to be combined.

python/lsst/cp/pipe/utils.py
```diff
@@ -631,25 +631,27 @@ def arrangeFlatsByExpTime(exposureList, exposureIdList):
...
     """
     flatsAtExpTime = {}
     assert len(exposureList) == len(exposureIdList), "Different lengths for exp. list and exp. ID lists"
-    for exp, expId in zip(exposureList, exposureIdList):
-        expTime = exp.getInfo().getVisitInfo().getExposureTime()
+    for expRef, expId in zip(exposureList, exposureIdList):
+        expTime = expRef.get(component='visitInfo').exposureTime
```
mfisherlevine, line 652: Actually, given you can do this, now I do wonder if you can use the `component='detector'` for my first comment.

mfisherlevine (approved):
Minor comments

python/lsst/cp/pipe/utils.py
timj: It is much faster to get just the detector via the butler if you don't want the entire exposure read into memory.

plazas: Merlin:  I woudl still need to get the exposure from the reference first, so it wouln not be much different to what's already above: `detector=list(inputExp.values())[0][0][0].get().getDetector()` vs `list(inputExp.values())[0][0][0].get(component='detector')` (if it works)
Tim:  In this function in `utils` we don't need the detector. Are you referring to `detector=list(inputExp.values())[0][0][0].get().getDetector()` in the `run` method: https://github.com/lsst/cp_pipe/blob/e1ecf2b791819f8a89129b60e4e17753adf1ce1f/python/lsst/cp/pipe/ptc/cpExtractPtcTask.py#L248 ? In that case, would I need to pass the butler to the `run` function as well from `runQuantum`? I wanted to just use the inputs to `run`: `inputExp` or  `inputDims`

timj: I was replying to Merlin's comment about component get vs Exposure get then getComponent  -- of course if you need the Exposure then you should get the component from the exposure.

mfisherlevine: Right, the point being that you do need to get it, but at the moment you're getting the whole thing twice, and so changing the instance above (my first comment on the PR) would mean that the first time you don't also retrieve the image. Given there's a nice syntax for it, and it doesn't involve having to get the whole thing and then special case the first pass through this loop (which is what I was worrying about) it like is worth it after all, I'd imagine (though perhaps given caching, it really isn't goig to matter that much in practice).

plazas: I see, thanks. I switched to using `get(component='detector')` above.

PR #115, opened 2021-12-22
DM-33041: Rename output to correctly match calib type

mfisherlevine:


craiglagegit (commented):
This looks good to me, although I admit I don't fully understand the connections that generate the documentation.  In reality, the only change I made to get it all to run was the change to pipelines/cpLinearitySolve.yaml .

czwa (approved):
Looks good.  I also did a local build of `pipelines_lsst_io` to check the doc changes.

PR #116, opened 2022-01-03
DM-32747: Many (>100) E2V amps had early PTC turn-off in LSSTCam

plazas:


python/lsst/cp/pipe/ptc/cpSolvePtcTask.py
```diff
@@ -453,42 +416,19 @@ def _getInitialGoodPoints(means, variances, maxDeviationPositive, maxDeviationNe
...
+        goodPoints = np.ones_like(means, dtype=bool)
+        pivotList = np.where(np.array(np.diff(variances)) < 0)[0]
+        if len(pivotList) > 0:
             # For small values, sometimes the variance decreases slightly
             # Only look when var > self.config.minVarPivotSearch
-            pivot = [p for p in pivot if variances[p] > minVarPivotSearch]
-            if len(pivot) > 0:
-                pivot = np.min(pivot)
+            pivotList = [p for p in pivotList if variances[p] > minVarPivotSearch]
+            if len(pivotList) > 0:
```
mfisherlevine, line 427: I think this needs to be `> 1` not zero - you can differentiate a list of length one.

python/lsst/cp/pipe/ptc/cpSolvePtcTask.py
```diff
@@ -432,16 +404,7 @@ def _getInitialGoodPoints(means, variances, maxDeviationPositive, maxDeviationNe
...
-        maxDeviationNegative : `float`
-            Maximum deviation from being constant for the variance/mean
-            ratio, in the negative direction.
-        minMeanRatioTest : `float`
-            Minimum signal value (in ADU) after which to start examining
-            the ratios var/mean.
-       minVarPivotSearch : `float`
+        minVarPivotSearch : `float`
             Minimum variance point (in ADU^2) after which the pivot point
             wher the variance starts decreasing should be sought.
```
mfisherlevine, line 409: `wher` → `where`. But also that sentence is just wrong somehow actually... Maybe:

`The variance (in ADU^2), above which, the point of decreasing variance should be sought.`

python/lsst/cp/pipe/ptc/cpSolvePtcTask.py
mfisherlevine: I checked and fixing this is all that's needed to make this work and not raise in situations with a single point of decrease.

python/lsst/cp/pipe/ptc/cpSolvePtcTask.py
```diff
@@ -453,42 +416,19 @@ def _getInitialGoodPoints(means, variances, maxDeviationPositive, maxDeviationNe
...
-            pivot = [p for p in pivot if variances[p] > minVarPivotSearch]
-            if len(pivot) > 0:
-                pivot = np.min(pivot)
+            pivotList = [p for p in pivotList if variances[p] > minVarPivotSearch]
+            if len(pivotList) > 0:
+                # Require that the decrease in variance happen for two
+                # consecutive signal levels
+                pivotIndex = np.min(np.where(np.diff(pivotList) == 1)[0])
+                pivot = pivotList[pivotIndex]
                 goodPoints[pivot+1:len(goodPoints)] = False
```
mfisherlevine, line 432: No need to `len(goodPoints)` here, just make the it `goodPoints[pivot+1:] = False` and you're done.

PR #117, opened 2022-01-20
DM-32746: Two amps in LSSTCam showed poor PTC curve fits.

plazas:


python/lsst/cp/pipe/ptc/cpSolvePtcTask.py
```diff
@@ -416,20 +431,27 @@ def _getInitialGoodPoints(means, variances, minVarPivotSearch):
...
+            # consecutive points. This will give a first
+            # estimate of the PTC turn-off, which
+            # may be updated (reduced) further in the code.
             if len(pivotList) > 1:
-                # Require that the decrease in variance happen for two
-                # consecutive signal levels
-                pivotIndex = np.min(np.where(np.diff(pivotList) == 1)[0])
-                pivot = pivotList[pivotIndex]
-                goodPoints[pivot+1:] = False
+                for k, g in groupby(enumerate(pivotList), lambda x: x[0]-x[1]):
```
czwa, line 448: Can this have a quick comment describing what it's doing?  I think I follow along, but even just a per-line  `# Find all indices in this group` kind of thing would clarify what's happening.

czwa (approved):
Just one minor comment on comments.

PR #121, opened 2022-02-23
DM-33512: Fix IndexError during BF kernel generation.

mfisherlevine:


craiglagegit:
Hi Chris.  Please review these fixes for this ticket.  scons and Jenkins have both passed, and Merlin and I think it is ready to merge if you agree.  Thanks.

czwa (approved):
This looks fine to me.  The quadratic unit test never looked right to me, and it now makes sense why the values were so odd.

PR #122, opened 2022-03-09
DM-31895: Rename/reorganize cp_pipe pipelines

plazas:


pipelines/cpBias.yaml
```diff
@@ -25,6 +25,6 @@ tasks:
       calibrationType: 'bias'
       exposureScaling: "Unity"
 contracts:
-  - isr.doBias == False
```
czwa, line 28: Contracts are "expressions written with Python syntax that should evaluate to a boolean" according to https://github.com/lsst/pipe_base/blob/main/doc/lsst.pipe.base/creating-a-pipeline.rst#verifying-configuration-contracts , so these should stay as `False`.

pipelines/cpDark.yaml
```diff
@@ -31,6 +31,6 @@ tasks:
       exposureScaling: "DarkTime"
       python: config.mask.append("CR")
 contracts:
-  - isr.doDark == False
```
czwa, line 34: As above, re contracts.

pipelines/cpFlat.yaml
```diff
@@ -37,4 +37,4 @@ tasks:
       exposureScaling: InputList
       scalingLevel: AMP
 contracts:
-  - isr.doFlat == False
```
czwa, line 40: As above.

pipelines/cpFlatSingleChip.yaml
```diff
@@ -4,26 +4,26 @@ tasks:
...
-      connections.outputExposure: 'cpFlatProc'
-      doBias: True
-      doVariance: True
-      doLinearize: True
-      doCrosstalk: True
-      doDefect: True
-      doNanMasking: True
-      doInterpolate: True
-      doDark: True
+      connections.outputExposure: 'cpflatProc'
```
czwa, line 7: Did a search/replace catch this accidentally?  Here and five more below.

pipelines/cpFringe.yaml
```diff
@@ -29,6 +29,6 @@ tasks:
       calibrationType: 'fringe'
       exposureScaling: "Unity"
 contracts:
-  - isr.doFringe == False
```
czwa, line 32: Contracts again.

czwa (approved):
Two issues, one with contracts and the other with what look like accidental renames.
I think this will be fine once those are fixed.

PR #123, opened 2022-03-17
DM-34101: Pipelines should not activate non-essential features

PaulPrice:


pipelines/cpFlat.yaml
```diff
@@ -5,18 +5,10 @@ tasks:
...
-      doCrosstalk: True
-      doDefect: True
-      doNanMasking: True
-      doInterpolate: True
-      doDark: True
+      doWrite: True
       doBrighterFatter: False
       doFlat: False
       doFringe: False
-      doApplyGains: False
```
czwa, line 19: `doApplyGains` needs to be disabled here, to ensure that the flat constructed doesn't have that gain scaling applied.  `doFlat` and `doApplyGains` can't both be enabled, so a flat constructed with the scaling can't be applied correctly.

pipelines/cpFlatSingleChip.yaml
```diff
@@ -5,18 +5,10 @@ tasks:
...
-      doCrosstalk: True
-      doDefect: True
-      doNanMasking: True
-      doInterpolate: True
-      doDark: True
+      doWrite: True
       doBrighterFatter: False
       doFlat: False
       doFringe: False
-      doApplyGains: False
```
czwa, line 19: Same as above.

czwa (approved):
Based on Jim's interpretation, I think this is fine, except for the `doApplyGains` issue in the flat pipelines.

PR #124, opened 2022-03-29
DM-34196: DM-31895 broke the piepelines.lsst.io build.

plazas:


ktlim (approved):
This looks like it fixes the build, but now I'm worried that the new cp_pipe.ptc tasks may not be getting pages properly built.  I'll approve this anyway.

PR #125, opened 2022-04-12
DM-33900: cp_pipe: calculate the gain using a pair of flats

plazas:


tests/test_ptc.py
```diff
@@ -394,6 +405,40 @@ def test_getGoodAmps(self):
...
+            idCounter += 2
+
+        resultsExtract = extractTask.run(inputExp=expDict, inputDims=expIds,
+                                         taskMetadata=[self.metadataContents])
+
+        for exposurePair in resultsExtract.outputCovariances:
+            for ampName in self.ampNames:
+                if exposurePair.gain[ampName] is np.nan:
+                    continue
+                self.assertAlmostEqual(exposurePair.gain[ampName], inputGain, delta=0.1)
```
czwa, line 436: Is that large of a tolerance needed?  I would think that the simulation would get better agreement.

czwa (approved):
I think most of my comments are about docstrings.  The only significant code change is moving the readnoise calculation out of a loop so it doesn't get recalculated multiple times.

python/lsst/cp/pipe/ptc/cpExtractPtcTask.py
```diff
@@ -618,3 +656,121 @@ def measureMeanVarCov(self, exposure1, exposure2, region=None):
...
+            Median of the overscan readnoise in the
+            post-ISR metadata of the input exposures (ADU).
+        """
+        # Empirical readout noise [ADU] measured from an
+        # overscan-subtracted overscan during ISR.
+        expectedKey = f"RESIDUAL STDEV {ampName}"
+
+        readNoises = []
+        for expMetadata in taskMetadata:
+            overscanNoise = expMetadata['isr'][expectedKey]
```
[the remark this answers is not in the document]
plazas: I'll check for the key `isr` and if it doesn't exist, I'll continue the loop. If in the end the median readout noise can't be calculated, I'll return `None` (with a warining), and then in the function that estimates the gain, if the read noise is None but the correction type is not `NONE`, instead of raising (as it is right now), I'll issue a warning and default to correction type `None`, which does not require the noise, to at least return an estimation.

tests/test_ptc.py
plazas: I see differences of ~0.065, especially with no correction.

PR #126, opened 2022-04-14
DM-22521: enable partial image reads in cp_pipe combine to avoid memory issues

czwa:
This uses the same tactic as pipe_tasks AssembleCoaddTask to remove the need to have all exposures in memory while doing the final calibration combine step.

### DIFF
--- a/python/lsst/cp/pipe/cpCombine.py
+++ b/python/lsst/cp/pipe/cpCombine.py
@@ -34,9 +34,9 @@ from astro_metadata_translator import merge_headers, ObservationGroup
 from astro_metadata_translator.serialize import dates_to_fits
 
 
-__all__ = ['CalibStatsConfig', 'CalibStatsTask',
-           'CalibCombineConfig', 'CalibCombineConnections', 'CalibCombineTask',
-           'CalibCombineByFilterConfig', 'CalibCombineByFilterConnections', 'CalibCombineByFilterTask']
+__all__ = ["CalibStatsConfig", "CalibStatsTask",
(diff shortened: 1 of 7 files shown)


python/lsst/cp/pipe/cpCombine.py
```diff
@@ -293,20 +289,20 @@ def run(self, inputExps, inputScales=None, inputDims=None):
...
+            inputDims = [dict() for i in inputExpHandles]
 
-        for index, (exp, dims) in enumerate(zip(inputExps, inputDims)):
+        for index, (expHandle, dims) in enumerate(zip(inputExpHandles, inputDims)):
             scale = 1.0
-            if exp is None:
-                self.log.warning("Input %d is None (%s); unable to scale exp.", index, dims)
-                continue
-
+            visitInfo = expHandle.get(component='visitInfo')
```
leeskelvin, line 296: I won't mind if you choose to ignore this, as it seems that this file is already using a mixture of `'` and `"`, but it seems like `"` is preferred more often than not?

leeskelvin (approved):

python/lsst/cp/pipe/cpCombine.py
czwa: Changed to more universally use `"`.  I will likely never be consistent, so I'm happy to fix these when they're pointed out.

timj: This is where black can really help: https://developer.lsst.io/python/formatting.html

PR #128, opened 2022-05-06
DM-34700: Change Config File for cpPTC.yaml

plazas:


czwa (changes requested):
I think this also needs to be changed in the top-level pipeline definition file as well.  That may make the changes in these files unnecessary.

plazas:
I set `doFlat` to`false` in the generic pipeline. I did leave `doDefect=true` in LATISS's pipeline, while setting `doDefect=false` in LSSTComCam's

czwa (approved):

PR #131, opened 2022-05-24
DM-34912: Include pixel clipping and masks in "getGainFromFlatPair" as in "measureMeanVarCov" of "cpPtcSolve"

plazas:


python/lsst/cp/pipe/ptc/cpExtractPtcTask.py
```diff
@@ -650,8 +627,67 @@ def measureMeanVarCov(self, exposure1, exposure2, region=None):
...
+        else:
+            im1Area = exposure1.maskedImage
+            im2Area = exposure2.maskedImage
+
+        if self.config.binSize > 1:
+            im1Area = afwMath.binImage(im1Area, self.config.binSize)
+            im2Area = afwMath.binImage(im2Area, self.config.binSize)
+
+        im1MaskVal = exposure1.getMask().getPlaneBitMask(self.config.maskNameList)
+        im1StatsCtrl = afwMath.StatisticsControl(self.config.nSigmaClipPtc,
```
czwa, line 662: Do you need two stat control objects that are essentially the same?  Is the concern that the mask planes will be in a different order?  These are also returned, but I'm not sure I see them used elsewhere.  Are they used in code that hasn't changed?

python/lsst/cp/pipe/ptc/cpExtractPtcTask.py
```diff
@@ -714,17 +744,33 @@ def getGainFromFlatPair(self, exposure1, exposure2, correctionType='NONE',
...
-            im1Area = exposure1.getImage()[region].getArray()
-            im2Area = exposure2.getImage()[region].getArray()
-        else:
-            im1Area = exposure1.getImage().getArray()
-            im2Area = exposure2.getImage().getArray()
 
-        const = np.mean((im1Area - im2Area)**2 / (im1Area + im2Area))
-        gain = 1. / const
+        (im1Area, im1MaskVal, im1StatsCtrl, im2Area,
+            im2MaskVal, im2StatsCtrl, mu1, mu2) = imageProperties.values()
```
czwa, line 749: Is there a reason to have this be a dict if it's only used as `imageProperties.values()`?  I think the ordering is guaranteed to be consistent, but a tuple would do the same thing.
If you expect to want to access these by key in the future, then ignore this comment.

czwa (approved):

python/lsst/cp/pipe/ptc/cpExtractPtcTask.py
plazas: Is there a guarantee that `exposure1.getMask().getPlaneBitMask(self.config.maskNameList)` will be equal to `exposure2.getMask().getPlaneBitMask(self.config.maskNameList)` ? Would that object be the same to be used in ` ratioImMaskVal = ratioIm.getMask().getPlaneBitMask(self.config.maskNameList)` of ` getGainFromFlatPair` and in ` diffImMaskVal = diffIm.getMask().getPlaneBitMask(self.config.maskNameList)` of `measureMeanVarCov`?

plazas: As for not being used after being returned, I think you are correct, thanks!

python/lsst/cp/pipe/ptc/cpExtractPtcTask.py
plazas: I'll change it to a tuple. I wasn't sure if a dictionary would be more compact.

PR #132, opened 2022-05-25
DM-34922: Report ptc turnoff in ptcDataset from cpPtcSolve

plazas:


python/lsst/cp/pipe/ptc/cpSolvePtcTask.py
```diff
@@ -831,6 +831,12 @@ def errFunc(p, x, y):
             goodPoints = self._getInitialGoodPoints(meanVecOriginal, varVecOriginal,
                                                     self.config.minVarPivotSearch,
                                                     self.config.consecutivePointsVarDecreases)
+
+            # Save the point where the variance starts decreasing as the
+            # PTC turnoff point
+            ptcTurnoff = meanVecOriginal[goodPoints][-1]
```
czwa, line 837: Do you need to check that there more than zero `goodPoints`?  Otherwise this will throw an `IndexError`.

czwa (approved):

python/lsst/cp/pipe/ptc/cpSolvePtcTask.py
plazas: Good catch. I think `goodPoints` will always be the same length as `meanVecOriginal`, but it could be that all its points are `False`, so this would fail. We do check for that possibility a few lines below and declare the amp as bad if it happens, so I will move these lines with the turnoff after that check is done.

PR #134, opened 2022-06-06
DM-35095: Update cp_pipe documentation to remove instrument specific pipelines from example commands

czwa:
Fix pipeline locations in documentation.


plazas (approved):
Looks good to me.

PR #136, opened 2022-06-09
DM-35135: cpCombine partial reads interferes with amplifier-level scaling

czwa:
Convert per-amp scales to an image.

This image can then be subset to match the subregion used in the data,
avoiding potential geometry mismatch.

### DIFF
--- a/python/lsst/cp/pipe/cpCombine.py
+++ b/python/lsst/cp/pipe/cpCombine.py
@@ -394,7 +394,7 @@ class CalibCombineTask(pipeBase.PipelineTask,
             raise RuntimeError("Inconsistent dimensions: %s" % dim)
         return dim.pop()
 
-    def applyScale(self, exposure, scale=None):
+    def applyScale(self, exposure, bbox=None, scale=None):
         """Apply scale to input exposure.
 
         This implementation applies a flux scaling: the input exposure is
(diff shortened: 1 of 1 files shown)


python/lsst/cp/pipe/cpCombine.py
```diff
@@ -400,17 +400,24 @@ def applyScale(self, exposure, scale=None):
...
+                # slower than only populating the region we care
+                # about, but this avoids needing to do arbitrary
+                # numbers of offsets, etc.
+                scaleExp = afwImage.MaskedImageF(exposure.getDetector().getBBox())
                 for amp, ampScale in zip(exposure.getDetector(), scale):
-                    ampIm = mi[amp.getBBox()]
-                    ampIm /= ampScale
-            else:
-                mi /= scale
+                    scaleExp.image[amp.getBBox()] = ampScale
```
erykoff, line 418: So this creates a fake full exposure and fills it with the amp scale values?

And I guess the performance penalty is that it has to do this every time we call applyScale?  Will the scale_list be different for every input exposure, or is there some caching that could be done?  (And feel free to tell me that this would be premature optimization anyway...)

erykoff (approved):
Looks good!  I am wondering about performance and if there's any caching that could be done, but it's probably not worth it unless this somehow becomes a pain point.

python/lsst/cp/pipe/cpCombine.py
czwa: That's correct.  I thought of trying to be clever, but that seems like I'm just making more places to get geometry wrong.
We do instantiate a new full-size image for each `applyScale` call, which means we make `(Nrows / NsubRows)` copies of the same image for each input.  This isn't optimal, but pulling it out of the loop will cause the same memory issues the partial reads fixed.  The scales are different for each input exposure, so I'm not sure there's any way to optimize there.

PR #137, opened 2022-06-15
DM-31938: Have PTC dataset unpad results

plazas:


czwa (approved):
No major concerns here.

czwa (approved):
As with the ip_isr pull, I think this all makes sense, provided it passes ci.

PR #138, opened 2022-06-16
DM-35252: Add DECam config overrides into cpBias/cpFlat

leeskelvin:


czwa (approved):
These look fine to me.  Do you need to convert any of the other pipelines, or do you not worry about the `dark`?

leeskelvin:
So far, DECam reductions have only made use of the bias and flat calibrations (as far as I'm aware). So yes, I think the bias and flats are all we need for now.

PR #139, opened 2022-07-18
DM-35316: Add 'detector' information (and any other missing info) to PTC dataset metadata

plazas:


python/lsst/cp/pipe/ptc/cpSolvePtcTask.py
```diff
@@ -187,10 +187,11 @@ def runQuantum(self, butlerQC, inputRefs, outputRefs):
             Output data refs to persist.
         """
         inputs = butlerQC.get(inputRefs)
-        outputs = self.run(inputCovariances=inputs['inputCovariances'], camera=inputs['camera'])
+        detId = inputRefs.inputCovariances[0].dataId['detector']
+        outputs = self.run(inputCovariances=inputs['inputCovariances'], camera=inputs['camera'], detId=detId)
```
czwa, line 191: Doesn't this need to pass `inputExpList`?  It didn't previously it seems, but doesn't that mean that it's always `None`?

python/lsst/cp/pipe/ptc/cpSolvePtcTask.py
```diff
@@ -201,7 +202,10 @@ def run(self, inputCovariances, camera=None, inputExpList=None):
             Input camera.
         inputExpList : `list` [`~lsst.afw.image.ExposureF`], optional
             List of exposures.
-
+        detId : `int`
+            Detector ID to locate the detector in teh camera and
```
czwa, line 206: typo

python/lsst/cp/pipe/ptc/cpSolvePtcTask.py
czwa: Was it only being used to find a detector before, and now that's being handled by the dimension information in `runQuantum`, so it isn't needed anymore?

plazas: I don't think it was used before (except for that one check, but then it was always ``None``), so I removed it.

PR #152, opened 2022-09-08
DM-35974: CTI code fails with unclear messages.

czwa:
Raise RuntimeError if processing cannot continue.

### DIFF
--- a/python/lsst/cp/pipe/deferredCharge.py
+++ b/python/lsst/cp/pipe/deferredCharge.py
@@ -229,6 +229,11 @@ class CpCtiSolveTask(pipeBase.PipelineTask):
         calib : `lsst.ip.isr.DeferredChargeCalib`
             Populated calibration.
 
+        Raises
+        ------
+        RuntimeError
+            Raised if no data remains after flux filtering.
+
         Notes
         -----
         The original CTISIM code uses a data model in which the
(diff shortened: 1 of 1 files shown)


python/lsst/cp/pipe/deferredCharge.py
```diff
@@ -229,6 +229,11 @@ def solveLocalOffsets(self, inputMeasurements, calib, detector):
...
             Populated calibration.
 
+        Raises
+        ------
+        RuntimeError
+            Raised if no data remains after flux filtering.
+
         Notes
         -----
         The original CTISIM code uses a data model in which the
```
plazas, line 239: Wher's the "original" CTSIM code? Should there be a reference, for readers not familiar with the context?

python/lsst/cp/pipe/deferredCharge.py
```diff
@@ -540,7 +559,9 @@ def findTraps(self, inputMeasurements, calib, detector):
                     new_signal.append(exposureDict[ampName]['LAST_MEAN'])
                 else:
                     Nskipped += 1
-            self.log.info(f"Skipped {Nskipped} exposures brighter than {self.config.maxSignalForCti}.")
+            self.log.info(f"Skipped {Nskipped} exposures brighter than {self.config.maxImageMean}.")
```
plazas, line 562: The additions look good (the new log messages), but I'm wondering if this code where you added the log message is repeated in all these functions (starting at the line with the comment that says "The signal is the mean intensity...."). Is this OK? Should it be refactored to avoid code duplication? (for another ticket, maybe)

plazas (approved):
The additions look fine. I just put a comment about potential code duplication. 

(somewhat unrelated: "__all__" should go before import statements, a ccording to the guide).

python/lsst/cp/pipe/deferredCharge.py
czwa: I wanted to keep the differences from the original code to a minimum, at least until it's been thoroughly tested.  I'm going to leave it for now, and will think about it in the future.

PR #153, opened 2022-09-09
DM-35569: ip_isr doEmpiricalReadNoise fails if an amplifier is fully masked

czwa:
Move countMaskedPixels to ip_isr.


plazas (approved):
LGTM.  

cp_pipe/utils.py only had this one test. Do we need more unit tests for those utils? Just a general comment.

PR #154, opened 2022-09-20
DM-36262: length of mask and covariance can disagree in BrighterFatterKernelSolveTask

czwa:
Remove masking that appears unnecessary.


plazas (approved):
Looks good, thanks for the fix. I only wonder if we need to change the "raw" prefix in the name, as the PTC code has "raw" when a mask has not been applied.

PR #155, opened 2022-09-26
DM-36358: "broken" amplifiers trigger a failure in setting the threshold for defects

czwa:
Handle exception caused by a single constant pixel value.

### DIFF
--- a/python/lsst/cp/pipe/defects.py
+++ b/python/lsst/cp/pipe/defects.py
@@ -35,6 +35,7 @@ from lsst.afw import cameraGeom
 from lsst.geom import Box2I, Point2I
 from lsst.meas.algorithms import SourceDetectionTask
 from lsst.ip.isr import Defects, countMaskedPixels
+from lsst.pex.exceptions import InvalidParameterError
 
 from ._lookupStaticCalibration import lookupStaticCalibration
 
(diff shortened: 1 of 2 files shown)


erykoff:
Can we add a test for this?

python/lsst/cp/pipe/defects.py
```diff
@@ -246,7 +247,15 @@ def findHotAndColdPixels(self, exp, nSigma):
 
                 threshold = afwDetection.createThreshold(nSig, 'stdev', polarity=polarity)
 
-                footprintSet = afwDetection.FootprintSet(ampImg, threshold)
+                try:
+                    footprintSet = afwDetection.FootprintSet(ampImg, threshold)
+                except InvalidParameterError:
+                    # This occurs if the image sigma value is 0.0.
+                    # Let's mask the whole area.
+                    minValue = np.min(ampImg.image.array) - 1.0
```
taranu, line 255: Just checking if this needs to be nanmin.

python/lsst/cp/pipe/defects.py
```diff
@@ -246,7 +247,15 @@ def findHotAndColdPixels(self, exp, nSigma):
 
                 threshold = afwDetection.createThreshold(nSig, 'stdev', polarity=polarity)
 
-                footprintSet = afwDetection.FootprintSet(ampImg, threshold)
+                try:
+                    footprintSet = afwDetection.FootprintSet(ampImg, threshold)
+                except InvalidParameterError:
+                    # This occurs if the image sigma value is 0.0.
```
ktlim, line 253: Is there any other way this could happen?  Testing exceptions should be the last resort.  Can't the `threshold` value be checked?

python/lsst/cp/pipe/defects.py
```diff
@@ -246,7 +247,15 @@ def findHotAndColdPixels(self, exp, nSigma):
...
+                    # Let's mask the whole area.
+                    minValue = np.nanmin(ampImg.image.array) - 1.0
+                    threshold = afwDetection.createThreshold(minValue, 'value', polarity=True)
+                    footprintSet = afwDetection.FootprintSet(ampImg, threshold)
```
ktlim, line 257: This seems like a pretty complicated (and expensive) way to create a `FootprintSet` covering the whole area.  Since a `FootprintSet` can be created from a `Box2I`, wouldn't that be simpler and faster?

python/lsst/cp/pipe/defects.py
czwa: The `threshold` itself is fine.  It's the image sigma value coming out as zero that causes the `InvalidParameterError`.  Manually calculating the image statistics seemed to be less efficient than doing the try/except, as most amplifiers should work fine.

python/lsst/cp/pipe/defects.py
czwa: I think I got lost in the `FootprintSet` definition before I saw that it could have a `Box` passed to it.  I'll refactor for that instead.

czwa: After coding this and sending to Jenkins, I'm now discovering that the box definition only sets the region, and does not create a footprint set containing footprints that can be used to mask the image.

PR #159, opened 2022-11-04
DM-36885: Ensure cp_pipe only uses a different ISR output when needed

czwa:
Switch cpSky to operate on postISRCCD datasetType.

### DIFF
--- a/pipelines/Latiss/cpSky.yaml
+++ b/pipelines/Latiss/cpSky.yaml
@@ -6,7 +6,6 @@ tasks:
   isr:
     class: lsst.ip.isr.isrTask.IsrTask
     config:
-      connections.outputExposure: cpSkyIsr
       overscan.fitType: 'MEDIAN_PER_ROW'
       doLinearize: false
       doCrosstalk: true
--- a/pipelines/cpSky.yaml
+++ b/pipelines/cpSky.yaml
@@ -2,10 +2,10 @@ description: Sky frame generation pipeline definition.
 tasks:
   isr:
     class: lsst.ip.isr.isrTask.IsrTask
-    config:
-      connections.outputExposure: cpSkyIsr
   cpSkyImage:
     class: lsst.cp.pipe.CpSkyImageTask
+    config:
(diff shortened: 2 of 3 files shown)


pipelines/Latiss/cpSky.yaml
```diff
@@ -6,7 +6,6 @@ tasks:
   isr:
     class: lsst.ip.isr.isrTask.IsrTask
     config:
-      connections.outputExposure: cpSkyIsr
       overscan.fitType: 'MEDIAN_PER_ROW'
```
leeskelvin, line 9: From what I can tell, the ISR task for Latiss in `cpSky` is configured differently to the main Latiss ISR task. When I compare the configs for this `cpSky` ISR task on this ticket against the default Latiss ISR task, this is my diff:
```
< config.overscan.fitType='MEDIAN_PER_ROW'
---
> config.overscan.fitType='MEDIAN'
204c204
< config.doCrosstalk=True
---
> config.doCrosstalk=False
```
Do these ISR task over-rides need to be upstreamed to the main Latiss ISR config over-rides, or can they be removed here?

leeskelvin: For reference, Latiss ISR task configs generated using:
```
pipetask build -p $CP_PIPE_DIR/pipelines/Latiss/cpSky.yaml#isr --show config
```

czwa: `doCrosstalk=True` has been an optimistic hope for a while.  I continue to be optimistic that we'll have a calibration soon (there are Camera Team produced ones that are probably sufficient for now).
We probably should switch the overscan method.  I'll check for concerns next week.

leeskelvin: Having chatted on Slack about this, I think these config differences aren't too problematic. They can be checked after the fact too, so perhaps this is a non-issue. If you're happy with this Chris, then I'm happy too!

leeskelvin (approved):

PR #164, opened 2023-01-18
DM-36388: Convert the PTC plotting script to a task

plazas:


python/lsst/cp/pipe/ptc/cpPlotPtcTask.py
```diff
@@ -0,0 +1,1522 @@
...
+        topPlot=False,
+    ):
+        """Plot C_ij/mu vs mu.
+
+        Figs. 8, 10, and 11 of Astier+19
+
+        Parameters
+        ----------
+        i : `int`
+            Covariane lag
```
aferte, line 785: typo and full stop to add here and below.

aferte (changes requested):
Typo to fix otherwise looks great

python/lsst/cp/pipe/ptc/cpPlotPtcTask.py
```diff
@@ -0,0 +1,1522 @@
...
+    camera = cT.PrerequisiteInput(
+        name="camera",
+        doc="Camera associated with this data.",
+        storageClass="Camera",
+        dimensions=("instrument",),
+        isCalibration=True,
+        lookupFunction=lookupStaticCalibration,
+    )
+    # ptcFitType = "FULLCOVARIANCE" produces 12 plots
+    ptcPlot1 = cT.Output(
```
czwa, line 73: Can you name these (or at least the output `dataset_type_name`) something more explicit, so that the on-disk file and the butler have something more human readable?  I'm not as worried about the internal usage, but being able to select a plot without having to look up if I want to look at `ptcPlot4` or `ptcPlot5` seems useful.

python/lsst/cp/pipe/ptc/cpPlotPtcTask.py
```diff
@@ -0,0 +1,1522 @@
...
+
+        Parameters
+        ----------
+        filenameFull : `str`
+            File name, including full path.
+        figDict : `dict` [`int`, `~matplotlib.figure.Figure]
+            Dictionary, keyed by an integer index, of
+            matplotlib PTC figures.
+        """
+        with PdfPages(filenameFull) as pdfPages:
```
czwa, line 276: This may need a try/except block;  I don't know if this will run correctly on a fully virtualized/cloud/etc node.  You can probably skip it for now, but if this gets added to a more involved pipeline definition it might cause issues.

python/lsst/cp/pipe/ptc/cpPlotPtcTask.py
plazas: I wanted to be able to plot all the individual files with a loop, something like this: 

from IPython.display import Image, display
detector = 55
for i in range(1,10):
    ref = butler.registry.findDataset(f'ptcPlot{i}', detector=detector)
    uri = butler.getURI(ref)
    display(Image(data=uri.read()))
    
I had also considered giving them more meaningful names. I can do that and will document the names in the ticket so the user can still do a loop like the above.

python/lsst/cp/pipe/ptc/cpPlotPtcTask.py
plazas: mm, so should I just delete this PDF option? I just wanted to keep the option of having a single PDF, which was convenient before for the original script.

plazas: I decided to delete this altogether, so it's more in line with the standards.  The user can make a PDF from a notebook using the retrieved PNGs.

python/lsst/cp/pipe/ptc/cpPlotPtcTask.py
aferte: Thanks, I had similar thoughts. Looks great now

aferte (approved):

PR #166, opened 2023-01-20
DM-37643: ci_cpp_gen3 fails on cpPtcExtract when noise is None

czwa:
Use NaN instead of None to prevent potential serialization issues.

### DIFF
--- a/python/lsst/cp/pipe/ptc/cpExtractPtcTask.py
+++ b/python/lsst/cp/pipe/ptc/cpExtractPtcTask.py
@@ -756,9 +756,9 @@ class PhotonTransferCurveExtractTask(pipeBase.PipelineTask):
         if correctionType not in ['NONE', 'SIMPLE', 'FULL']:
             raise RuntimeError("Unknown correction type: %s" % correctionType)
 
-        if correctionType != 'NONE' and readNoise is None:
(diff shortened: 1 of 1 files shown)


python/lsst/cp/pipe/ptc/cpExtractPtcTask.py
```diff
@@ -756,9 +756,9 @@ def getGainFromFlatPair(self, im1Area, im2Area, imStatsCtrl, mu1, mu2,
         if correctionType not in ['NONE', 'SIMPLE', 'FULL']:
             raise RuntimeError("Unknown correction type: %s" % correctionType)
 
-        if correctionType != 'NONE' and readNoise is None:
+        if correctionType != 'NONE' and not np.isfinite(readNoise):
```
plazas, line 759: Why ``np.isfinite`` instead of ``np.isnan``? Just curious, I guess that, for this case, they work the same?

plazas (approved):
Looks good.

python/lsst/cp/pipe/ptc/cpExtractPtcTask.py
czwa: No particular reason.  `isfinite` catches other values that aren't plausible as well, but there are other impossible values (negative values).  I'm going to leave it for now.

PR #167, opened 2023-02-07
DM-37902: Mask edges at the AMP level as default when calculating the PTC

plazas:


plazas:
> When I ran this, I also included isr:doSuspect=True, because the default is False. But I see you are doing it in the ptcExtract task instead of the isr task. So then the doSuspect=True is not needed, is that correct?
@craiglagegit  We can add this to the isr part of the pipeline in the new yaml file I created. Why is this option needed?

craiglagegit:
Well, maybe I am confused, but I thought the two options you added will mark the 20 pixels around the edge of each amp as "Suspect".  but unless doSuspect=True, the Suspect pixels will not actually be masked.  I could easily be wrong on how this works.

czwa:
It would be good to run a quick test to confirm that `doSuspect` is or is not needed.  I think on my read that it isn't (`ptcExtract` calls the `IsrTask.maskEdges`, and the `SUSPECT` bit is included in the `ptcExtract` `maskNameList`, so the edge should be excluded from the calculation on line 687-697), but I may be missing something subtle.
I'm marking this as Accepted on the assumption I did trace through the code correctly (and that the test will be run to check).

czwa (approved):

PR #169, opened 2023-02-09
DM-36390: calibrations should track sequencer crc

czwa:
Add updateMetadataFromExposures calls.

### DIFF
--- a/python/lsst/cp/pipe/defects.py
+++ b/python/lsst/cp/pipe/defects.py
@@ -162,6 +162,7 @@ class MeasureDefectsTask(pipeBase.PipelineTask):
         msg = "Found %s defects containing %s pixels in %s"
         self.log.info(msg, len(defects), self._nPixFromDefects(defects), datasetType)
 
+        defects.updateMetadataFromExposures([inputExp])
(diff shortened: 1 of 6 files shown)


python/lsst/cp/pipe/defects.py
```diff
@@ -162,6 +162,7 @@ def run(self, inputExp, camera):
         msg = "Found %s defects containing %s pixels in %s"
         self.log.info(msg, len(defects), self._nPixFromDefects(defects), datasetType)
 
+        defects.updateMetadataFromExposures([inputExp])
```
mfisherlevine, line 165: General comment, not really for here, but why so few items? Don't we need this for biases, darks, flats etc? I see they're all being compared in `isrTask.run()` but are they being set correctly somewhere?

czwa: These are merged in `merge_headers` for the bias, dark, and flat.  For the `IsrCalib` based products, this call adds the header information from the exposure to the per-exposure defect set, which are then combined in the `Merge` task.  I wanted to have a central method (this is part of the base class) to handle these keywords.

mfisherlevine: OK, great, thanks.

PR #171, opened 2023-02-13
Tickets/dm 37951

craiglagegit:
This is working, but I'm sure has things that need correcting.  Summary of the status is in the Jira ticket.


python/lsst/cp/pipe/ptc/cpSolvePtcTask.py
```diff
@@ -223,27 +226,27 @@ def run(self, inputCovariances, camera=None, detId=0):
             # Ignore dummy datasets
             if partialPtcDataset.ptcFitType == 'DUMMY':
                 continue
+            # Modifications here for the subregions to append new data to the old
             for ampName in ampNames:
-                datasetPtc.inputExpIdPairs[ampName].append(partialPtcDataset.inputExpIdPairs[ampName])
```
czwa, line 227: Doesn't this change the dimensionality?  Previously the partial inputs were lists of tuples, appended into a list-of-list-of-tuples.  Using `+=` flattens it into a list-of-tuples again.  I think this less of a problem below, as it looks like it was just selecting the first element of the input lists for those entries.

PR #174, opened 2023-03-01
DM-38163: Update PTC to avoid potential failures

czwa:
PTC improvements:

- Ensure all outputRefs have a matching outputCovariance (this is a problem blocking DM-37383).
- Try to get read noise from the exposure header.
- Get ampNames for the solve task from a non-dummy input, and anticipate associated bug fix on the PTC dataset in `ip_isr`.
- Update tests to pass full set of taskMetadata to mimic real butler behavior.

### DIFF
--- a/python/lsst/cp/pipe/ptc/cpExtractPtcTask.py
+++ b/python/lsst/cp/pipe/ptc/cpExtractPtcTask.py
@@ -257,9 +257,45 @@ class PhotonTransferCurveExtractTask(pipeBase.PipelineTask):
             inputs['inputExp'] = arrangeFlatsByExpId(inputs['inputExp'], inputs['inputDims'])
 
         outputs = self.run(**inputs)
+        outputs = self._guaranteeOutputs(inputs['inputDims'], outputs, outputRefs)
         butlerQC.put(outputs, outputRefs)
 
+    def _guaranteeOutputs(self, inputDims, outputs, outputRefs):
+        """Ensure that all outputRefs have a matching output, and if they do
+        not, fill the output with dummy PTC datasets.
+
+        Parameters
+        ----------
+        inputDims : `dict` [`str`, `int`]
+            Input exposure dimensions.
+        outputs : `lsst.pipe.base.Struct`
+            Outputs from the ``run`` method.  Contains the entry:
+
+            ``outputCovariances``
+                Output PTC datasets (`list` [`lsst.ip.isr.IsrCalib`])
+        outputRefs : `~lsst.pipe.base.connections.OutputQuantizedConnection`
+            Container with all of the outputs expected to be generated.
+
+        Returns
+        -------
+        outputs : `lsst.pipe.base.Struct`
+            Dummy dataset padded version of the input ``outputs`` with
+            the same entries.
+        """
+        newCovariances = []
+        for ref in outputRefs.outputCovariances:
+            outputExpId = ref.dataId['exposure']
+            if outputExpId in inputDims:
+                entry = inputDims.index(outputExpId)
+                newCovariances.append(outputs.outputCovariances[entry])
+            else:
+                newPtc = PhotonTransferCurveDataset(['no amp'], 'DUMMY', 1)
+                newPtc.setAmpValues('no amp')
+                newCovariances.append(newPtc)
+        return pipeBase.Struct(outputCovariances=newCovariances)
+
     def run(self, inputExp, inputDims, taskMetadata):
+
         """Measure covariances from difference of flat pairs
 
         Parameters
--- a/python/lsst/cp/pipe/ptc/cpExtractPtcTask.py
+++ b/python/lsst/cp/pipe/ptc/cpExtractPtcTask.py
@@ -349,7 +349,8 @@ class PhotonTransferCurveExtractTask(pipeBase.PipelineTask):
         # dimensions match.
         dummyPtcDataset = PhotonTransferCurveDataset(ampNames, 'DUMMY',
                                                      self.config.maximumRangeCovariancesAstier)
-
+        for ampName in ampNames:
+            dummyPtcDataset.setAmpValues(ampName)
         # Get read noise.  Try from the exposure, then try
         # taskMetadata.  This adds a get() for the exposures.
         readNoiseLists = {}
--- a/python/lsst/cp/pipe/ptc/cpExtractPtcTask.py
+++ b/python/lsst/cp/pipe/ptc/cpExtractPtcTask.py
@@ -350,13 +350,27 @@ class PhotonTransferCurveExtractTask(pipeBase.PipelineTask):
         dummyPtcDataset = PhotonTransferCurveDataset(ampNames, 'DUMMY',
                                                      self.config.maximumRangeCovariancesAstier)
 
+        # Get read noise.  Try from the exposure, then try
+        # taskMetadata.  This adds a get() for the exposures.
+        readNoiseLists = {}
+        for pairIndex, expRefs in inputExp.items():
+            # This yields an index (exposure_time, seq_num, or flux)
+            # and a pair of references at that index.
+            for expRef, expId in expRefs:
+                # This yields an exposure ref and an exposureId.
+                exposure = expRef.get()
+                metadataIndex = inputDims.index(expId)
+                thisTaskMetadata = taskMetadata[metadataIndex]
+
+                for ampName in ampNames:
+                    if ampName not in readNoiseLists:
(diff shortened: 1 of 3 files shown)


python/lsst/cp/pipe/ptc/cpExtractPtcTask.py
```diff
@@ -313,14 +349,29 @@ def run(self, inputExp, inputDims, taskMetadata):
...
-            readNoiseDict[ampName] = self.getReadNoiseFromMetadata(taskMetadata, ampName)
+        # Get read noise.  Try from the exposure, then try
+        # taskMetadata.  This adds a get() for the exposures.
+        readNoiseLists = {}
+        for pairIndex, expRefs in inputExp.items():
+            # This yields an index (exposure_time, seq_num, or flux)
+            # and a pair of references at that index.
+            for expRef, expId in expRefs:
+                # This yields an exposure ref and an exposureId.
+                exposure = expRef.get()
```
erykoff, line 362: Can't we just get the metadata component rather than the full exposure?  Or is there something else we need?

python/lsst/cp/pipe/ptc/cpExtractPtcTask.py
```diff
@@ -257,9 +257,45 @@ def runQuantum(self, butlerQC, inputRefs, outputRefs):
             inputs['inputExp'] = arrangeFlatsByExpId(inputs['inputExp'], inputs['inputDims'])
 
         outputs = self.run(**inputs)
+        outputs = self._guaranteeOutputs(inputs['inputDims'], outputs, outputRefs)
         butlerQC.put(outputs, outputRefs)
 
+    def _guaranteeOutputs(self, inputDims, outputs, outputRefs):
```
plazas, line 263: Two questions: 
1) Why do we need this? I thought that `cpPtcExtract` was already introducing dummy exposures in order to match the inputs: https://github.com/lsst/cp_pipe/blob/main/python/lsst/cp/pipe/ptc/cpExtractPtcTask.py#L311 
or are you building a pipeline that will only use the classes in `cpPtcSolve` without running `cpPtcExtract` first? 
2) It's not that clear to me how the function is implementing the matching.

python/lsst/cp/pipe/ptc/cpExtractPtcTask.py
```diff
@@ -792,40 +843,35 @@ def getGainFromFlatPair(self, im1Area, im2Area, imStatsCtrl, mu1, mu2,
 
         return gain
 
-    def getReadNoiseFromMetadata(self, taskMetadata, ampName):
+    def getReadNoise(self, exposure, taskMetadata, ampName):
         """Gets readout noise for an amp from ISR metadata.
```
plazas, line 847: Update dosctring with explanation if the function is now doing more.

python/lsst/cp/pipe/ptc/cpExtractPtcTask.py
```diff
@@ -792,40 +843,35 @@ def getGainFromFlatPair(self, im1Area, im2Area, imStatsCtrl, mu1, mu2,
...
-            Median of the overscan readnoise in the
-            post-ISR metadata of the input exposures (ADU).
-            Returns 'None' if the median could not be calculated.
+            The read noise for this set of exposure/amplifier.
         """
-        # Empirical readout noise [ADU] measured from an
-        # overscan-subtracted overscan during ISR.
-        expectedKey = f"RESIDUAL STDEV {ampName}"
+        # Try from the exposure first.
+        expectedKey = f"LSST ISR OVERSCAN RESIDUAL SERIAL STDEV {ampName}"
```
plazas, line 864: Hoes does this noise from the exposure metadata potentially differ from the noise form the task metadata? And why did you need to add this extra check?

python/lsst/cp/pipe/ptc/cpSolvePtcTask.py
```diff
@@ -219,8 +219,14 @@ def run(self, inputCovariances, camera=None, detId=0):
                 means, variances, and exposure times
                 (`lsst.ip.isr.PhotonTransferCurveDataset`).
         """
+        # Find the ampNames from a non-dummy ptc.
```
plazas, line 222: In `cpPtcExtract`, the dummy datasets had the ampNames: https://github.com/lsst/cp_pipe/blob/main/python/lsst/cp/pipe/ptc/cpExtractPtcTask.py#L314 
So are you actually building a pipeline that does not have `cpPtcExtract` as a previous step of `cpPTcSolve`?

plazas (changes requested):
Thanks for these changes. It looks good in general, but I would just like to understand them a bit better their motivation (and probably add a bit more explanation in the code) before accepting.

python/lsst/cp/pipe/ptc/cpExtractPtcTask.py
czwa: The issue arises when running in the `cp_testing` environment.  There is a check during the ISR stage to restrict the inputs to flat exposures only, returning `pipeBase.NoWorkFound` for the other exposures.  This is fine in other stages, such as flat combine, as whatever exposures are passed as input are used to construct a single output.  However, `cpPtcExtract` is expecting to write an output for all inputs to the pipeline, so it ends up in a situation where (for my test case) `len(inputRefs.inputExp) = 124` but `len(outputRefs.outputCovariances) = 229`.  The existing code can supply a dummy dataset for each input, but that leaves 105 expected outputs that are not generated.  This new function iterates over the full set of `outputRefs.outputCovariances`, constructing the missing outputs and inserting them into the expected output location.
As for the matching:  the function iterates over all of the outputs, and if that output's `exposureId` is found in the `inputDims` list, then it needs to extract that entry from the existing outputs (which are currently indexed by the ordering of `exposureId` in `inputDims`) and insert it into the location expected in the list of `outputRefs`.  If the expected output is not in the existing output list, then the input exposure was ignored during ISR, and so a dummy dataset needs to be inserted into that location.

python/lsst/cp/pipe/ptc/cpExtractPtcTask.py
czwa: I simply didn't think of this at the time.  Switched to get just the metadata.

python/lsst/cp/pipe/ptc/cpExtractPtcTask.py
czwa: They should be identical.  I added this because I was finding that all my read noise values were being set to NaN.  The `cp_testing` pipeline renames the `IsrTask` stage of the pipeline to `cptPtcIsr` to make it unique within the pipeline, and that changes the key within the `taskMetadata` object that stores that information.  I originally tried making that key configurable, but decided that that was a work-around to a work-around, so I added the lookup in the exposure metadata instead.  As this header key should be present in all ISR processed exposures from now on, this change is also the first step to deprecating and removing the `taskMetadata` lookup and connection entirely.

python/lsst/cp/pipe/ptc/cpSolvePtcTask.py
czwa: The original code was relying on all input PTC datasets having the same unique set of `ampNames`.  The bug in `ip_isr` allowed each instance of the `PhotonTransferCurveDataset` to update the global list of `ampNames` for all instances, as it updated the class definition instead of just the instance definition.  This meant that if someone was lazy about constructed the additional padding datasets (me), then the `Solve` task would attempt to iterate over the union of all `ampNames` found in all datasets.  Fixing the bug ensured that the inputs only knew about their own `ampNames`, so the `np.unique` no longer was needed to prune the duplicates from the first input, but it did require that the first input have the same set of `ampNames` as all other inputs.  Adding the requirement the `ampNames` come from a not-DUMMY dataset allows the additional padding datasets to be lazy, and makes the content of the DUMMY datasets unimportant; they only serve to make the middleware accounting correct, and have no science value.

python/lsst/cp/pipe/ptc/cpExtractPtcTask.py
plazas: How do you end up with `len(outputRefs.outputCovariances) = 229` after running `cpPtcExtract`, if the input number was only `124`?

python/lsst/cp/pipe/ptc/cpExtractPtcTask.py
plazas: "is also the first step to deprecating and removing the taskMetadata lookup and connection entirely"--> :) :)

plazas (approved):

PR #175, opened 2023-03-09
DM-38301: Sort by y values to ensure intermittent bad pixels are correctly masked.

erykoff:


python/lsst/cp/pipe/defects.py
```diff
@@ -397,7 +398,6 @@ def _markBlocksInBadColumn(self, x, y, multipleX, defects):
                         limits.append(multipleY[gapIndex])
                         limits.append(multipleY[gapIndex+1])
                     limits.append(maxY)  # maximum last
-                    assert len(limits)%2 == 0, 'limits is even by design, but check anyways'
```
czwa, line 400: Isn't this still the case?  Is this just dropping a check that isn't useful?

czwa (approved):
Looks fine.  I'm going to file a separate ticket to fix up the docstrings, as I'm sure it's not just this method that needs updating.

python/lsst/cp/pipe/defects.py
erykoff: I mean ... I guess I could bring it back?  It just seems strange to construct something and then assert that it's true.

PR #178, opened 2023-04-03
DM-37684: Enable cp_pipe defect code to run on combined exposures

plazas:


python/lsst/cp/pipe/defects.py
```diff
@@ -74,6 +77,23 @@ class MeasureDefectsTaskConfig(pipeBase.PipelineTaskConfig,
...
+        default='STDEV',
+        allowed={'STDEV': "Use number of sigma given standard deviation.",
+                 'VALUE': "Use pixel value."},
+    )
+    darkCurrentThreshold = pexConfig.Field(
+        dtype=float,
+        doc="Dark current threshold (in e-/sec) to define hot/bright pixels in dark images.",
+        default=5,
+    )
+    fracThreshold = pexConfig.Field(
```
erykoff, line 92: This should probably be something like `flatFracThreshold`

python/lsst/cp/pipe/defects.py
```diff
@@ -74,6 +77,23 @@ class MeasureDefectsTaskConfig(pipeBase.PipelineTaskConfig,
...
+    thresholdType = pexConfig.ChoiceField(
+        dtype=str,
+        doc=("Defects threshold type: 'STDEV' or `VALUE`."),
+        default='STDEV',
+        allowed={'STDEV': "Use number of sigma given standard deviation.",
+                 'VALUE': "Use pixel value."},
+    )
+    darkCurrentThreshold = pexConfig.Field(
+        dtype=float,
+        doc="Dark current threshold (in e-/sec) to define hot/bright pixels in dark images.",
```
erykoff, line 89: Is this only applicable if `thresholdType='Value'`?  If so it should be in the docstring.

python/lsst/cp/pipe/defects.py
```diff
@@ -74,6 +77,23 @@ class MeasureDefectsTaskConfig(pipeBase.PipelineTaskConfig,
...
+        default=5,
+    )
+    fracThreshold = pexConfig.Field(
+        dtype=float,
+        doc="Fractional threshold to define cold/dark pixels in flat images.",
```
erykoff, line 94: Same as above on the docstring.

python/lsst/cp/pipe/defects.py
```diff
@@ -190,7 +208,7 @@ def _nPixFromDefects(defects):
             nPix += defect.getBBox().getArea()
         return nPix
 
-    def findHotAndColdPixels(self, exp, nSigma):
+    def findHotAndColdPixels(self, exp):
```
erykoff, line 211: A small thing, but if this is not a public interface (and if we can change the interface like this I hope not!) then we can make it `_findHotAndColdPixels`.

python/lsst/cp/pipe/defects.py
```diff
@@ -238,10 +253,41 @@ def findHotAndColdPixels(self, exp, nSigma):
...
+            meanClip = afwMath.makeStatistics(ampImg, afwMath.MEANCLIP, ).getValue()
+            ampImg -= meanClip
 
+            # Determine thresholds
+            stDev = afwMath.makeStatistics(ampImg, afwMath.STDEVCLIP, ).getValue()
+            expTime = exp.getInfo().getVisitInfo().getExposureTime()
+            thresholdType = self.config.thresholdType
+            datasetType = exp.getMetadata().get('IMGTYPE', 'UNKNOWN')
+            if thresholdType == 'VALUE':
+                hotPixelThreshold = self.config.darkCurrentThreshold*expTime/amp.getGain()
```
erykoff, line 265: I would think that in the `value` land, the hot pixel threshold is for darks only, and the cold pixel threshold is for flats only.  (This is certainly what was done by EOtest in the past, and what has been vetted so far.)

python/lsst/cp/pipe/defects.py
```diff
@@ -238,10 +253,41 @@ def findHotAndColdPixels(self, exp, nSigma):
...
+                coldPixelThreshold = self.config.nSigmaDark
+                if datasetType.lower() == 'dark':
+                    nSigmaList = [hotPixelThreshold]
+                else:
+                    nSigmaList = [hotPixelThreshold, coldPixelThreshold]
+
+            # Go back to pixel values to provide log info below
+            valuesList = [np.array(nSigmaList[0])*stDev*amp.getGain()/expTime]
+            if len(nSigmaList) == 2:
+                valuesList.append(np.array(nSigmaList[1])*stDev/meanClip + 1.)
```
erykoff, line 284: I don't understand the above computations or why scalars are wrapped in an `np.array`.  We have the hot and cold pixel thresholds above, and maybe this is reverse engineering that, with very specific assumptions.  Or the calculations could be done in the `else` block above.  But they really are just "hot" and "cold" values and they are not the same so shouldn't be listed together.

python/lsst/cp/pipe/defects.py
```diff
@@ -716,3 +762,154 @@ def run(self, inputDefects, camera):
...
+
+# Subclass the MergeDefects task to reduce the input dimensions
+# from ("instrument", "detector", "exposure") to
+# ("instrument", "detector").
+
+
+class MergeDefectsCombinedConnections(MergeDefectsConnections,
+                                      dimensions=("instrument", "detector")):
+    inputDefects = cT.Input(
+        name="singleExpDefects",
```
erykoff, line 774: This name should be `cpBrightPartialDefectsCombined` as defined in the default pipeline, and then the pipeline doesn't need the override.  That is, the default shouldn't be useless and immediately overridden.  However, I don't know if I like the name.  Are these *bright* defects or are they defects from the *dark*.  E.g, should it be `cpPartialDefectsFromDarkCombined` or some combination of those words?

python/lsst/cp/pipe/defects.py
```diff
@@ -716,3 +762,154 @@ def run(self, inputDefects, camera):
...
+        doc="Measured defect lists.",
+        storageClass="Defects",
+        dimensions=("instrument", "detector",),
+        multiple=False,
+    )
+    secondaryInputDefects = cT.Input(
+        name="AdditionalSingleExpDefects",
```
erykoff, line 781: This default name should be `cpPartialDefectsCombinedWithFilter`.  Or maybe `cpPartialDefectsFromFlatCombined` and that implies the filter.  Furthermore, I don't know how this works ... are we only doing one flat combination in here, or does this need to support `physical_filter` dimension and `multiple=True`?

python/lsst/cp/pipe/defects.py
```diff
@@ -716,3 +762,154 @@ def run(self, inputDefects, camera):
...
+        name='camera',
+        doc="Camera associated with this exposure.",
+        storageClass="Camera",
+        dimensions=("instrument", ),
+        isCalibration=True,
+        lookupFunction=lookupStaticCalibration,
+    )
+
+    outputDefects = cT.Output(
+        name="combinedExpDefects",
```
erykoff, line 854: This name should be the default instead of in the pipeline (and should match `cpPartialDefectsFromDarkCombined` or whatever we settle on).

python/lsst/cp/pipe/defects.py
```diff
@@ -716,3 +762,154 @@ def run(self, inputDefects, camera):
...
```
erykoff, line 896: This name should be the default instead of in the pipeline (and should match `cpPartialDefectsFromFlatCombined ` or whatever we settle on).

python/lsst/cp/pipe/defects.py
```diff
@@ -716,3 +762,154 @@ def run(self, inputDefects, camera):
...
+        doc="Output measured defects.",
+        storageClass="Defects",
+        dimensions=("instrument", "detector"),
```
erykoff, line 899: Doesn't this need the `physical_filter` dimension, or else you could clobber yourself if you ran multiple filters of data through the defect pipeline?

tests/test_defects.py
```diff
@@ -701,8 +698,7 @@ def test_badImage(self):
         config.nPixBorderLeftRight = 0
 
         task = cpPipe.defects.MeasureDefectsTask(config=config)
-        defects = task.findHotAndColdPixels(testImage, [config.nSigmaBright,
-                                                        config.nSigmaDark])
+        defects = task.findHotAndColdPixels(testImage)
```
erykoff, line 701: I don't see any tests for the combined tasks, or the `VALUE` option.

erykoff (changes requested):
Looks like this is coming together, but I think there are some clarifications that need to happen with what the thresholds mean; more tests are required; and probably tests in ci_cpp_gen3; and some care taken with the flat dimensions.

python/lsst/cp/pipe/defects.py
plazas: Correct. But in standard deviation land here in DM and as we have been doing so far, we do have cold and dark pixels from flats. We were discussing this with @czwa, and decided to keep value and stdev lands doing analogous things here in DM. Of course, then this would become a difference in definition with `eoTest` and I have pointed this out to Jim.   
I think we can make a decision here at this point. If we decide to be completely analogous to `eoTest` then we would need to do hot pixels for darks and only cold for flats, both in VALUE and STDEV modes.

python/lsst/cp/pipe/defects.py
plazas: In the `if  thresholdType == 'VALUE'` ... `else` conditional, `nSigmaList` will always have the list of thresholds in `STDEV` units (with a bit of overloading of the name in the first part of the conditional). After the conditional block, we'd like to report in the log both the `STDEV` and equivalent pixel `VALUE` values, so we do that conversion.  We put things in arrays because, in principle, the thresholds are a list (hot a cold for flats, currently) and there is an iteration over that list/array in the algorithm below (we wanted to modify the current algorithm as little as possible).  As for the actual formulas from the thresholds, these come directly from the `eoTest` code. I can add a few more comments explaining the `eoTest` requirements, I think Jim has shared these with us somewhere.

python/lsst/cp/pipe/defects.py
plazas: I can make a config option for this .

python/lsst/cp/pipe/defects.py
plazas: The defects measured from darks are referred to bright or hot (above dark current threshold).  I'll use `cpPartialDefectsFromDarkCombined` and put it in the code, deleting it from the yaml file.

python/lsst/cp/pipe/defects.py
erykoff: This still needs to be updated.

python/lsst/cp/pipe/defects.py
erykoff: This name also needs updating.

python/lsst/cp/pipe/defects.py
```diff
@@ -716,3 +784,154 @@ def run(self, inputDefects, camera):
...
+        # which is what MergeDefectsTask expects.
+        tempList = [inputs['inputDefects'], inputs['secondaryInputDefects']]
+        # Rename inputDefects
+        inputsCombined = {'inputDefects': tempList, 'camera': inputs['camera']}
+
+        outputs = super().run(**inputsCombined)
+        butlerQC.put(outputs, outputRefs)
+
+
+class MeasureDefectsCombinedConnections(MeasureDefectsConnections,
```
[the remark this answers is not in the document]
erykoff: Yes, great point!

python/lsst/cp/pipe/defects.py
plazas: Yes, thanks! I'm working on it.

python/lsst/cp/pipe/defects.py
plazas: I changed the name to `cpPartialDefectsFromFlatCombined`.

plazas: This was failing before and @czwa suggested the solution "with Filter".

python/lsst/cp/pipe/defects.py
plazas: Changed the name to `cpPartialDefectsFromFlatCombined`. What you say about the filter dimension makes sense to me. @czwa , what are your thoughts?

czwa: I hadn't thought of the case where multiple flats could be found.  Keeping the filter dimension makes sense to me now as well.  The associated connection in the merge code needs to be updated as well to match dimensions, but that task can still have dimensions of `(instrument, detector)`.

python/lsst/cp/pipe/defects.py
czwa: I think a config option so we can run in both modes (cold pixels only from flats, or cold and hot both from flats) is the least confusing options.

plazas: I added the config option.

tests/test_defects.py
plazas: I added a new file for the combined defects task that repeats the tests in the existing defects task and added a couple of tests to the `VALUE` option.

pipelines/findDefectsCombined.yaml
```diff
@@ -0,0 +1,15 @@
...
+    config:
+      connections.inputExp: 'dark'
+  measureDefectsFlat:
+    class: lsst.cp.pipe.defects.MeasureDefectsCombinedWithFilterTask
+    config:
+      connections.inputExp: 'flat'
+  mergeDefectsCombined:
+    class: lsst.cp.pipe.defects.MergeDefectsCombinedTask
+    config:
+      connections.mergedDefects: 'defects'
```
erykoff, line 14: This seems like it isn't correct, right?

python/lsst/cp/pipe/defects.py
```diff
@@ -74,16 +77,39 @@ class MeasureDefectsTaskConfig(pipeBase.PipelineTaskConfig,
...
     """
 
+    thresholdType = pexConfig.ChoiceField(
+        dtype=str,
+        doc=("Defects threshold type: 'STDEV' or `VALUE`. If 'VALUE', cold pixels will be found "
+             "in flats, and hot pixels in darks. If 'STDEV', cold and hot pixels will be found "
+             "in flats, and hot pixels in darks."),
+        default='STDEV',
+        allowed={'STDEV': "Use number of sigma given standard deviation.",
+                 'VALUE': "Use pixel value."},
```
erykoff, line 87: Maybe these descriptions should say "for threshold" (but that may be redundant in context).  It just seems a bit terse as it is.

python/lsst/cp/pipe/defects.py
```diff
@@ -74,16 +77,39 @@ class MeasureDefectsTaskConfig(pipeBase.PipelineTaskConfig,
...
+    )
+    darkCurrentThreshold = pexConfig.Field(
+        dtype=float,
+        doc=("If thresholdType='VALUE', dark current threshold (in e-/sec) to define "
+             "hot/bright pixels in dark images."),
```
erykoff, line 92: Should probably add `"Unused if thresholdType==``STDEV``"`.

python/lsst/cp/pipe/defects.py
```diff
@@ -74,16 +77,39 @@ class MeasureDefectsTaskConfig(pipeBase.PipelineTaskConfig,
...
+        default=5,
+    )
+    fracThresholdFlat = pexConfig.Field(
+        dtype=float,
+        doc=("If thresholdType='VALUE', fractional threshold to define cold/dark "
+             "pixels in flat images (fraction of the mean value per amplifier)."),
```
erykoff, line 98: Should probably add `"Unused if thresholdType==``STDEV``"`.

python/lsst/cp/pipe/defects.py
```diff
@@ -74,16 +77,39 @@ class MeasureDefectsTaskConfig(pipeBase.PipelineTaskConfig,
...
+        default=0.8,
+    )
     nSigmaBright = pexConfig.Field(
         dtype=float,
-        doc=("Number of sigma above mean for bright pixel detection. The default value was found to be"
-             " appropriate for some LSST sensors in DM-17490."),
+        doc=("If thresholdType='STDEV', number of sigma above mean for bright/hot "
+             "pixel detection. The default value was found to be "
+             "appropriate for some LSST sensors in DM-17490."),
```
erykoff, line 105: Should probably add `"Unused if thresholdType==``VALUE``.`.

python/lsst/cp/pipe/defects.py
```diff
@@ -74,16 +77,39 @@ class MeasureDefectsTaskConfig(pipeBase.PipelineTaskConfig,
...
         default=4.8,
     )
     nSigmaDark = pexConfig.Field(
         dtype=float,
-        doc=("Number of sigma below mean for dark pixel detection. The default value was found to be"
-             " appropriate for some LSST sensors in DM-17490."),
+        doc=("If thresholdType='STDEV', number of sigma below mean for dark/cold pixel "
+             "detection. The default value was found to be "
+             "appropriate for some LSST sensors in DM-17490."),
```
erykoff, line 112: Should probably add `"Unused if thresholdType==``VALUE``"`.

python/lsst/cp/pipe/defects.py
```diff
@@ -716,3 +858,70 @@ def run(self, inputDefects, camera):
...
+        storageClass="Defects",
+        dimensions=("instrument", "detector",),
+        multiple=False,
+    )
+    secondaryInputDefects = cT.Input(
+        name="cpPartialDefectsFromFlatCombinedWithFilter",
+        doc="Additional measured defect lists.",
+        storageClass="Defects",
+        dimensions=("instrument", "detector", "physical_filter"),
+        multiple=False,
```
erykoff, line 881: Right now we only send one flat in, so this `multiple=False` is okay.  I think this is okay, but I would like to know from @jchiang87 or @czwa if there is any possibility that we would be sending multiple combined flats from different filters through this defect pipeline.

erykoff (commented):
This is very close, a few docstrings to touch up, and a couple of defaults need updating.  I think the new tests are good, and while I was wondering if the new combined defect tests should use VALUE as the default, if STDEV remains the task default then these are correct; and the new VALUE code is tested here.  (Whether we should replace STDEV with VALUE as the default is a bigger question for us and the camera team.)

python/lsst/cp/pipe/defects.py
```diff
@@ -74,16 +77,39 @@ class MeasureDefectsTaskConfig(pipeBase.PipelineTaskConfig,
     """Configuration for measuring defects from a list of exposures
     """
 
+    thresholdType = pexConfig.ChoiceField(
+        dtype=str,
+        doc=("Defects threshold type: 'STDEV' or `VALUE`. If 'VALUE', cold pixels will be found "
+             "in flats, and hot pixels in darks. If 'STDEV', cold and hot pixels will be found "
+             "in flats, and hot pixels in darks."),
```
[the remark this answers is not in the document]
plazas: With @czwa we decided to leave it like this, as trying to use the `eoTest` dark current threshold also as a threshold for hot pixels in the flat didn't seem to work straightforwardly and was creating complications with the tests and definitions.  It should be a more general question for us (camera + DM) to decide whether we keep what `eoTest` or `DM` have been doing so far (as implemented in `VALUE` and `STDEV`). as I have mentioned before.

python/lsst/cp/pipe/defects.py
erykoff: Yeah, something like "Use pixel value to determine detection threshold" or "Use image standard deviation to determine detection threshold" or something like that.

python/lsst/cp/pipe/defects.py
jchiang87: I can't think of any reason for sending multiple combined flats through this pipeline.

python/lsst/cp/pipe/defects.py
```diff
@@ -716,3 +861,70 @@ def run(self, inputDefects, camera):
...
         )
+
+# Subclass the MergeDefects task to reduce the input dimensions
+# from ("instrument", "detector", "exposure") to
+# ("instrument", "detector").
+
+
+class MergeDefectsCombinedConnections(MergeDefectsConnections,
+                                      dimensions=("instrument", "detector")):
+    inputDefects = cT.Input(
```
erykoff, line 872: This is a minor point, and may be irrelevant.  But now that we have different dimensions, I wonder if the name here should be `darkDefects` and the secondary should be `flatDefects`?

tests/test_defectsCombined.py
```diff
@@ -0,0 +1,766 @@
...
+        boxesMeasured = []
+        for defect in defects:
+            boxesMeasured.append(defect.getBBox())
+
+        for expectedBBox in shouldBeFound:
+            self.assertIn(expectedBBox, boxesMeasured)
+
+        shouldBeMissed = self.darkBBoxes[self.onlyEdges] + self.brightBBoxes[self.onlyEdges]
+        for boxMissed in shouldBeMissed:
+            self.assertNotIn(boxMissed, boxesMeasured)
```
erykoff, line 623: How are the tests above here any different from the tests in the original file?  Isn't it only the value tests below that matter?  Or am I missing something?

plazas: They are mostly the same, as I mentioned in another comment above: `I added a new file for the combined defects task that repeats the tests in the existing defects task and added a couple of tests to the VALUE option.`

erykoff: But it looks as if they are identical, so what's different?  I'm trying to figure out if anything is different except for the couple of VALUE tests.  If so, then those are the only tests that are needed here and the duplicated tests should be deleted.  But if something else is different, please let me know!

python/lsst/cp/pipe/defects.py
plazas: I like long and descriptive names, like the ones that are already there: `cpPartialDefectsFromDarkCombined` and `cpPartialDefectsFromFlatCombinedWithFilter`. But I would be happy to change them to what you suggest if you believe that's better. Please let me know.

erykoff: I mean the name on the left side of the equation, `inputDefects` -> `inputDarkDefects` which I think is more descriptive, and below `secondaryInputDefects` -> `inputFlatDefects`.

tests/test_defectsCombined.py
plazas: They are mostly identical. The default class and configurations are the combined ones, e.g., `self.defaultTask = cpPipe.defects.MeasureDefectsCombinedTask()`, in addition to the `VALUE` tests.

python/lsst/cp/pipe/defects.py
plazas: Ah, I see! Yeah, I can change that :)

python/lsst/cp/pipe/defects.py
```diff
@@ -716,3 +861,70 @@ def run(self, inputDefects, camera):
         return pipeBase.Struct(
             mergedDefects=merged,
         )
+
+# Subclass the MergeDefects task to reduce the input dimensions
+# from ("instrument", "detector", "exposure") to
+# ("instrument", "detector").
+
+
+class MergeDefectsCombinedConnections(MergeDefectsConnections,
```
erykoff, line 870: This should not inherit from MergeDefectsConnections, it has different connections.  It should just inherit from the real base class.

tests/test_defectsCombined.py
erykoff: The task is different, but it's a sub-task, and the tests here are only calling the individual methods, and not the task.  So the code paths are essentially identical.  I would remove all of these except for the VALUE tests which are new and useful.  And the combined part of the pipeline (changing connections) is done with ci_cpp_gen3.

plazas: Sounds good. I have implemented and pushed these changes.

erykoff (approved):

PR #179, opened 2023-04-04
DM-38486: Combined dark seems to not have exposure time

czwa:
Populate a visitInfo for combined calibration.

### DIFF
--- a/python/lsst/cp/pipe/cpCombine.py
+++ b/python/lsst/cp/pipe/cpCombine.py
@@ -341,18 +341,6 @@ class CalibCombineTask(pipeBase.PipelineTask):
         self.combineHeaders(inputExpHandles, combinedExp,
                             calibType=self.config.calibrationType, scales=expScales)
 
-        # Populate a visitInfo.  Set the exposure time and dark time
-        # to 0.0 or 1.0 as appropriate, and copy the instrument name
-        # from one of the inputs.
-        expTime = 1.0
-        if self.config.connections.outputData.lower() == 'bias':
-            expTime = 0.0
-        inputVisitInfo = inputExpHandles[0].get(component="visitInfo")
-        visitInfo = afwImage.VisitInfo(exposureTime=expTime, darkTime=expTime,
-                                       instrumentLabel=inputVisitInfo.getInstrumentLabel())
-
-        combinedExp.getInfo().setVisitInfo(visitInfo)
-
         # Set the detector
         combinedExp.setDetector(inputDetector)
 
@@ -498,7 +486,8 @@ class CalibCombineTask(pipeBase.PipelineTask):
 
     def combineHeaders(self, expHandleList, calib, calibType="CALIB", scales=None):
         """Combine input headers to determine the set of common headers,
-        supplemented by calibration inputs.
+        supplemented by calibration inputs.  The calibration header is
+        set in-place.
 
         Parameters
         ----------
@@ -559,6 +548,17 @@ class CalibCombineTask(pipeBase.PipelineTask):
             if scales is not None:
                 header.set("CPP_INPUT_SCALE_%d" % (i,), scales[i])
 
+        # Populate a visitInfo.  Set the exposure time and dark time
+        # to 0.0 or 1.0 as appropriate, and copy the instrument name
+        # from one of the inputs.
+        expTime = 1.0
+        if self.config.connections.outputData.lower() == 'bias':
+            expTime = 0.0
+        inputVisitInfo = visitInfoList[0]
+        visitInfo = afwImage.VisitInfo(exposureTime=expTime, darkTime=expTime,
+                                       instrumentLabel=inputVisitInfo.instrumentLabel)
+        calib.getInfo().setVisitInfo(visitInfo)
+
         # Not yet working: DM-22302
         # Create an observation group so we can add some standard headers
         # independent of the form in the input files.
@@ -569,7 +569,7 @@ class CalibCombineTask(pipeBase.PipelineTask):
             self.log.warning("Exception making an obs group for headers. Continuing.")
             # Fall back to setting a DATE-OBS from the calibDate
             dateCards = {"DATE-OBS": "{}T00:00:00.00".format(calibDate)}
-            comments["DATE-OBS"] = "Date of start of day of calibration midpoint"
+            comments["DATE-OBS"] = "Date of start of day of calibration creation"
         else:
             oldest, newest = group.extremes()
             dateCards = dates_to_fits(oldest.datetime_begin, newest.datetime_end)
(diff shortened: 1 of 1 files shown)


timj:
In theory, if we are propagating input headers to output and adjusting all the times properly, then ObservationInfo should still work (because it will have the "raw" headers understood by it) and then you could use the obsInfo to visitInfo routine from obs_base.

czwa:
Checking this manually to see if the header in an existing calibration could be converted fails with a `ValueError: None of the registered translation classes ['DECam', 'SuprimeCam', 'HSC', 'MegaPrime', 'SDSS'] understood this header`.  The `TELESCOP` parameter is possibly the problem, with `LSST AuxTelescope` found in the header not matching the `Rubin Auxiliary Telescope` that the LATISS translator in `obs_lsst` has.
Constructing the `visitInfo` manually seemed the right solution, as the exposure time is ill defined for some products: flats represent some product of throughputs that isn't really time dependent, so the use of 1.0s times there (and in the fringe which has a similar issue) is to supply a time that is safe against dividing by the exposure time.  The dark should scale correctly, but that scale isn't propagated into the input metadata (although it is in the output metadata).  1.0s cannot be the correct times for the bias, which is why that is special cased to 0.0.

timj:
> ValueError

Okay. The problem here is that the metadata translator for LATISS hasn't been registered yet because nothing knows it has to `import lsst.obs.lsst.translators`. The only way this would be imported is if the code asked butler for the instrument class and imported that (using `Instrument.fromName()`) and then it would work. The advantage of the ObservationInfo approach is you get a fully populated combo `visitInfo` but I understand that this might not be workable (especially if you are stripping headers like EXPTIME out the combined header rather than updating it) [I hope you aren't leaving the EXPTIME header in there with a confusing value]

Also, we need to fix the LATISS headers so they have the right telescope name...

python/lsst/cp/pipe/cpCombine.py
```diff
@@ -341,6 +341,18 @@ def run(self, inputExpHandles, inputScales=None, inputDims=None):
...
 
+        # Populate a visitInfo.  Set the exposure time and dark time
+        # to 0.0 or 1.0 as appropriate, and copy the instrument name
+        # from one of the inputs.
+        expTime = 1.0
+        if self.config.connections.outputData.lower() == 'bias':
+            expTime = 0.0
+        inputVisitInfo = inputExpHandles[0].get(component="visitInfo")
+        visitInfo = afwImage.VisitInfo(exposureTime=expTime, darkTime=expTime,
+                                       instrumentLabel=inputVisitInfo.getInstrumentLabel())
```
timj, line 352:
```suggestion
                                       instrumentLabel=inputVisitInfo.instrumentLabel)
```

python/lsst/cp/pipe/cpCombine.py
```diff
@@ -341,6 +341,18 @@ def run(self, inputExpHandles, inputScales=None, inputDims=None):
         self.combineHeaders(inputExpHandles, combinedExp,
                             calibType=self.config.calibrationType, scales=expScales)
 
+        # Populate a visitInfo.  Set the exposure time and dark time
+        # to 0.0 or 1.0 as appropriate, and copy the instrument name
+        # from one of the inputs.
+        expTime = 1.0
```
timj, line 347: How come the exposure time isn't the sum of all the dark exposure times? We could easily calculate that in combineHeaders since we already loop over visitInfo to store the exposure times in the header.

python/lsst/cp/pipe/cpCombine.py
```diff
@@ -341,6 +341,18 @@ def run(self, inputExpHandles, inputScales=None, inputDims=None):
         self.combineHeaders(inputExpHandles, combinedExp,
```
timj, line 341: I'm a bit confused by this call because we don't do anything with the header this calculates.

I also just looked in combineHeaders and I wonder if we can make use of the `first` and `last` parameters to `merge_headers` -- those are designed to allow you to pick the first DATE-OBS you encounter and the last DATE-END you encounter such that the ObservationGroup code at the end might not be needed. Also we now use DATE-BEG in addition to DATE-OBS.

python/lsst/cp/pipe/cpCombine.py
```diff
@@ -341,6 +341,18 @@ def run(self, inputExpHandles, inputScales=None, inputDims=None):
...
                             calibType=self.config.calibrationType, scales=expScales)
 
+        # Populate a visitInfo.  Set the exposure time and dark time
+        # to 0.0 or 1.0 as appropriate, and copy the instrument name
+        # from one of the inputs.
+        expTime = 1.0
+        if self.config.connections.outputData.lower() == 'bias':
+            expTime = 0.0
+        inputVisitInfo = inputExpHandles[0].get(component="visitInfo")
```
timj, line 350: Would be convenient if we didn't have to do this because combineHeaders returned some kind of combo VisitInfo as well as the modified header.

timj (changes requested):
I have many questions. I've also re-read combineHeaders and it confuses me a lot because it seems to be making life hard for itself in many places and then not being used anywhere. There are two places it's called and none of them use the return value.

python/lsst/cp/pipe/cpCombine.py
czwa: It's 1.0s as the scale factors used to combine the input darks are the `dark_time` values.  This normalizes the combined dark to have unit exposure time, which is then scaled to match the `dark_time` of the exposures that will be corrected (in `IsrTask`).

python/lsst/cp/pipe/cpCombine.py
czwa: I've moved this into the `combineHeaders` call, as that has the `VisitInfo` list already available, to avoid the unnecessary extra `get`.

python/lsst/cp/pipe/cpCombine.py
timj: Okay. It's all very confusing and seems to be working differently to what it's documented to do. There is the added confusion that it generates a merged header but then copies everything that got dropped back into the original header but that's exactly what the `first` mode option does in `merge_headers` anyhow, and if the `first` and `last` parameters are used it will automatically makes sure it picks up the first DATE-OBS and the last DATE-END.  The way it's now written you don't want `merged` to exist at all because you are wanting to use the other header you got from the exposure and then copy everything back out of `merged` into the alias.

I think ObservationGroup is used because that can sort the headers into date order before it reads back the start time and end time (which it gets from header translation).

timj (approved):
This looks okay. Thanks.

Maybe change the documentation for combineHeaders to say that it will update the header in place for `calib` as well as returning it.

PR #180, opened 2023-04-04
DM-38562:  cp_pipe test fails with lmfit 1.1.0

czwa:
Update deferred charge test value.


parejkoj (approved):
With the caveat that I wish we understood how sensitive these tolerances are to the details of the fitter, this is fine.

PR #181, opened 2023-04-17
DM-38736: Fix index mismatches in output when computing PTC full covariance.

erykoff:


python/lsst/cp/pipe/ptc/cpSolvePtcTask.py
```diff
@@ -435,24 +435,30 @@ def fitDataFullCovariance(self, dataset):
             if len(maskAtAmp) == 0:
                 maskAtAmp = np.repeat(True, len(muAtAmp))
 
-            muAtAmp = muAtAmp[maskAtAmp]
-            covAtAmp = np.nan_to_num(dataset.covariances[ampName])[maskAtAmp]
-            covSqrtWeightsAtAmp = np.nan_to_num(dataset.covariancesSqrtWeights[ampName])[maskAtAmp]
+            muAtAmpMasked = muAtAmp[maskAtAmp]
```
czwa, line 438: I was initially confused by this, but this renaming is to ensure that the unmasked values are the ones that are written out, and not the masked values that previously replaced those, correct?

czwa (approved):
One quick question to ensure I'm understanding correctly.

python/lsst/cp/pipe/ptc/cpSolvePtcTask.py
erykoff: Yeah, I thought that reusing names to mean masked or unmasked was confusing.  And this way we make sure we have the unmasked values from the individual PTC files to set in the solved/combined PTC file.

PR #182, opened 2023-04-24
DM-38777:  LinearizeSpline linearity corrections do not anchor the spline at zero flux

czwa:
Anchor spline linearity corrections.

### DIFF
--- a/python/lsst/cp/pipe/linearity.py
+++ b/python/lsst/cp/pipe/linearity.py
@@ -453,6 +453,15 @@ class LinearitySolveTask(pipeBase.PipelineTask):
                     binCenters = binCenters[numPerBin > 0]
 
                 self.debugFit('splineFit', binCenters, np.abs(values), values, None, ampName)
+                # Anchor the spline to have zero correction at zero
+                # flux as well as at the lowest measured flux bin.
+                if binCenters[0] != 0.0:
+                    if values[0] != 0.0:
+                        offset = values[0]
+                        values -= offset
+                    np.concatenate(([0.0], binCenters))
+                    np.concatenate(([0.0], values))
+
                 interp = afwMath.makeInterpolate(binCenters.tolist(), values.tolist(),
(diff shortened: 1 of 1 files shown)


python/lsst/cp/pipe/linearity.py
```diff
@@ -453,6 +453,15 @@ def run(self, inputPtc, dummy, camera, inputDims, inputPhotodiodeData=None,
                     binCenters = binCenters[numPerBin > 0]
 
                 self.debugFit('splineFit', binCenters, np.abs(values), values, None, ampName)
+                # Anchor the spline to have zero correction at zero
+                # flux as well as at the lowest measured flux bin.
+                if binCenters[0] != 0.0:
```
taranu, line 458: Should this check if any binCenters are negative?

python/lsst/cp/pipe/linearity.py
```diff
@@ -453,6 +453,15 @@ def run(self, inputPtc, dummy, camera, inputDims, inputPhotodiodeData=None,
...
+                    if values[0] != 0.0:
```
taranu, line 459: Does this behave appropriately in both cases of `values[0]<0` and `values[0]>0`? Just wondering if you can end up with negative `values` and if that's worth checking or if that's only possible if you did something obviously wrong in the first place.

czwa: Negative values in the correction is fine; the `values` correspond to amounts to be added to pixels at that flux, and negative values simply means some amount needs to be subtracted.

python/lsst/cp/pipe/linearity.py
czwa: There shouldn't be any negative binCenters, as these are measured from a series of flat field images.  I've added a `raise` if there are negative binCenters, as that means something has gone wrong.

taranu (approved):

PR #184, opened 2023-04-25
DM-38826: ZeroDivisionError in lsst.cp.pipe.defects.MeasureDefectsCombinedWithFilterTask

plazas:


tests/test_defects.py
```diff
@@ -635,10 +634,18 @@ def valueThreshold(self, fileType):
...
-            # Change the default a bit so it works for the
-            # existing simulated defects.
-            task.config.fracThresholdFlat = 0.9
+            if saturateFlat:
+                exp.image.array[:] = 0.0
+                x = self.defaultConfig.nPixBorderUpDown
+                y = self.defaultConfig.nPixBorderLeftRight
+                shapeY, shapeX = exp.image.array.shape
+                width, height = shapeX - 2*x, shapeY - 2*x
+                shouldBeFound = [Box2I(corner=Point2I(x, y), dimensions=Extent2I(width, height))]
```
erykoff, line 643: So this is supposed to find all defects in this case?  Shouldn't it find none?

tests/test_defects.py
```diff
@@ -650,8 +657,10 @@ def valueThreshold(self, fileType):
             self.assertIn(expectedBBox, boxesMeasured)
 
     def test_valueThreshold(self):
-        for fileType in ['flat', 'flat']:
+        for fileType in ['dark', 'flat']:
```
erykoff, line 660: Thanks for catching this!

erykoff (commented):
Can you clarify if it's finding all defect or no defect?

tests/test_defects.py
plazas: I found out empirically by running this test that it is declaring the whole detector as a defect (one box). I could only do it perhaps for just one amp?

erykoff: I think we want it to mask out one amp, and then I think it should return no defects, at least through this code path.  Can you figure out where it's marking everything as a defect?

plazas: I think that's the footprint that the ``afw`` code returns with that sigma: 

`ipdb>  footprintList
[995488 peaks, area=1000486, centroid=(257.5, 2996)]
ipdb> defects = Defects.fromFootprintList(footprintList)
ipdb> defects
<lsst.ip.isr.defects.Defects object at 0x7f91ddd9c430>
ipdb> defects.toDict()
{'metadata': <lsst.daf.base.propertyContainer.propertyList.PropertyList object at 0x7f91b8103530>, 'x0': [7], 'y0': [2000], 'width': [502], 'height': [1993]}`

The above is for amp `C10` in the example that Jim paste din the ticket. 

As for the unit test, I'll do it only for one amp.

erykoff: Okay, so that's not from afw, that's from the code that explicitly makes a bad amp if sigma=0.0.  And that is currently the intended behavior, but we may revisit it.  So this is actually fine as it is.

erykoff (approved):

PR #189, opened 2023-05-09
DM-39048: Use Gaussian histogram KS-test for good point selection in PTC fit.

erykoff:
In addition to changing to the Gaussian histogram ks-test p-value for good point / turnoff selection, this PR also fixes a few other issues:

* Min/max point selection has been moved from the extraction task to the solver task, as this is a more convenient place for testing different cuts.
* When doing exponential or polynomial fit, the finalMeans and finalVars had an index error due to the way padding was done.  This has been fixed.
* The final ptc dataset now has the same exposure ID sorting for every amp.  This wasn't a bug but was potentially confusing.

### DIFF
--- a/python/lsst/cp/pipe/ptc/cpExtractPtcTask.py
+++ b/python/lsst/cp/pipe/ptc/cpExtractPtcTask.py
@@ -968,6 +968,14 @@ class PhotonTransferCurveExtractTask(pipeBase.PipelineTask):
             Chi-squared per degree of freedom of Gaussian fit.
         kspValue : `float`
             The KS test p-value for the Gaussian fit.
+
+        Notes
+        -----
+        The algorithm here was originally developed by Aaron Roodman.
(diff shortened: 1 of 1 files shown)


czwa (approved):

python/lsst/cp/pipe/ptc/cpExtractPtcTask.py
```diff
@@ -899,3 +936,72 @@ def getReadNoise(self, exposureMetadata, taskMetadata, ampName):
...
+            Clipped mean of im2Area (ADU).
+
+        Returns
+        -------
+        varFit : `float`
+            Variance from the Gaussian fit.
+        chi2Dof : `float`
+            Chi-squared per degree of freedom of Gaussian fit.
+        kspValue : `float`
+            The KS test p-value for the Gaussian fit.
```
plazas, line 964: Could you comment in Notes on what values are expected for Gaussian distribution? Or what are the range of values considered in this code?

PR #191, opened 2023-05-11
DM-39178: Add deterministic seed to ks-test model sampling.

erykoff:


python/lsst/cp/pipe/ptc/cpExtractPtcTask.py
```diff
@@ -1009,7 +1009,15 @@ def computeGaussianHistogramParameters(self, im1Area, im2Area, imStatsCtrl, mu1,
             sigmaFit = out.params["sigma"].value
 
             # Calculate KS test p-value for the fit.
-            gSample = scipy.stats.norm.rvs(size=numOk, scale=sigmaFit, loc=out.params["center"].value)
+            # Seed this with the mean value, so that the same data will get the
+            # same result.
+            randomSeed = int((mu1 + mu2)/2.)
```
mfisherlevine, line 1014: It's an interesting thought. But is it really what you want? Upstream changes which cause changes in the inputs will cause a seed-change - why not just hard code it like we do everywhere else? I mean, I think it's fine, just kinda curious if you really do want to do that...

mfisherlevine (approved):

python/lsst/cp/pipe/ptc/cpExtractPtcTask.py
erykoff: I don't know if I fully believe in completely hard-coding seeds outside tests; in fact we often use exp number or something that changes but is deterministic.  I just wanted something quick that would give the same output for the same input.

mfisherlevine: Ah, I'd forgotten that it was tests I was thinking of. OK, nice, thanks.

mfisherlevine: The fact the fit is so on the hairy edge that seeds cause it to fail is a different worry though, but for now, fixing random Jenkins failures is clearly more important.

erykoff: Well, the hairy-edge part is going to be fixed on DM-39167.

mfisherlevine: OK, I figured, thanks 🙂

PR #193, opened 2023-05-15
DM-39167: Improve outlier rejection in PTC fitting and add additional tests.

erykoff:
This PR adds:
* A revised implementation of the outlier rejection algorithm that correctly implements the original Digel eotest code where the initial fit is only on the lower flux points, to prevent high flux outliers from ruining the fit.
* Additional tests of PTC dataset reordering and raw value invariance, and fixes for bugs that these tests uncovered.
* Robustification of ks-test computation.
* Update PTC turnoff point to be highest point after outlier rejection, and not before (following the original eotest code).

I also renamed the short-lived `doLegacyTurnoffAndOutlierSelection` to `doLegacyTurnoffSelection` because I updated the outlier rejection for both the legacy and updated versions.


tests/test_ptc.py
```diff
@@ -183,7 +186,16 @@ def test_covAstier(self):
         resultsExtract.outputCovariances[-2].rawMeans['C:0,0'] = np.array([np.nan])
         resultsExtract.outputCovariances[-2].rawVars['C:0,0'] = np.array([np.nan])
 
-        resultsSolve = solveTask.run(resultsExtract.outputCovariances,
+        # Force the next-to-last PTC dataset to have a decreased variance to
+        # ensure that the outlier fit rejection works.
+        rawVar = resultsExtract.outputCovariances[-4].rawVars['C:0,0']
```
czwa, line 191: Is this comment incorrect here?  `outputCovariances[-4]` isn't next-to-last, unless I'm confused on the indexing.

tests/test_ptc.py
```diff
@@ -193,6 +205,49 @@ def test_covAstier(self):
...
+            # Check that the PTC turnoff is correctly computed.
+            # This will be different for the C:0,0 amp.
+            if amp == 'C:0,0':
+                self.assertAlmostEqual(ptc.ptcTurnoff[amp], ptc.rawMeans[ampName][-3])
+            else:
+                self.assertAlmostEqual(ptc.ptcTurnoff[amp], ptc.rawMeans[ampName][-1])
+
+            # Test that all the quantities are correctly ordered and have
+            # not accidentally been masked.
+            for i, extractPtc in enumerate(resultsExtract.outputCovariances[::2]):
```
czwa, line 217: Another indexing question to make sure I'm not confused:  This is skipping the "dummy" entries, correct?

czwa (approved):

python/lsst/cp/pipe/ptc/cpExtractPtcTask.py
```diff
@@ -196,6 +196,12 @@ class PhotonTransferCurveExtractConfig(pipeBase.PipelineTaskConfig,
         doc="Minimum number of good data values to compute KS test histogram.",
         default=100,
     )
+    ksHistSampleMultiplier = pexConfig.Field(
+        dtype=int,
+        doc="Multiplier for how many more points to sample from the model to reduce measurement "
+            "noise.",
+        default=10,
+    )
```
czwa, line 204: (Comment I forgot on the first pass)
Is there a reason to sample the model for this test instead of using the known Gaussian CDF with `ks_1samp`?  I doubt it matters, but this multiplier made me wonder if just using the CDF would be easier.

erykoff: That is a great point!  I have refactored to use the `ks_1samp` and gaussian cdf.

czwa (approved):

PR #194, opened 2023-05-15
DM-39212: Move no-instrument pipelines in cp_pipe to an "ingredients" directory

czwa:
Move pipeline defaults to ingredients directory.


leeskelvin:
Thanks @czwa, these changes look good. Would you be able to add a README similar to https://github.com/lsst/drp_pipe/blob/main/pipelines/README.md into the `pipelines` directory, and similarly another README into `_ingredients` similar to https://github.com/lsst/drp_pipe/blob/main/ingredients/README.md? (Those README's haven't been updated to reflect the new layout, but they should hopefully be useful). Otherwise, I think this is good, cheers!

leeskelvin (approved):

PR #199, opened 2023-06-13
DM-39604: Record full noise matrix in PTC dataset

plazas:


erykoff (changes requested):
Please add a check of the noise matrix in the `test_ptc.py` code.  It can be empirical, but do check that it was computed and the numbers are sane and consistent.

tests/test_ptc.py
```diff
@@ -253,6 +258,30 @@ def test_covAstier(self):
                     extractPtc.covariancesSqrtWeights[ampName][0],
                     ptc.covariancesSqrtWeights[ampName][i],
                 )
```
erykoff, line 260: `assertFloatsAlmostEqual` can work with arrays, this can be just one call for each; or maybe you're only checking the diagonals?  Wait, this is a dict ...  that isn't correct.

```
noiseMatrixExpected = np.array(
    [[29.37, 0, 0],
     [0, -23.209, 0],
     [0, 0, 35.69]]
)
```
above (or maybe do the full matrix with off-diagonals?) would certainly be a cleaner comparison (and also for the noB case).

plazas: Thanks. I now use the full matrix too.

erykoff:
I assume you ran black on that file to clean up the formatting?  If so, that must be a separate commit, or else it's very hard to follow what is substance and what is style in the git history.  Thanks!

plazas:
@erykoff Yes, I had run Black along with the changes. I have reverted the commit, and now there are two separate commits. Please let me know if it looks good now and if I can merge after running Jenkins!

erykoff (approved):

PR #202, opened 2023-06-22
DM-39639: Refactor linearity task, add Astier spline-based fit, and add full tests.

erykoff:


tests/test_linearity.py
```diff
@@ -0,0 +1,407 @@
...
+
+            # Check that the spline parameters recovered are consistent,
+            # with input to some low-grade precision.
+            # The first element should be identically zero.
+            self.assertFloatsEqual(linearizer.linearityCoeffs[amp_name][0], 0.0)
+
+            self.assertFloatsAlmostEqual(
+                linearizer.linearityCoeffs[amp_name][n_nodes:],
+                non_lin_spline_values,
+                atol=7.0,
```
czwa, line 380: Is this tolerance needed for all of the spline parameters?  I'm less concerned about a 70/7 or 185/7 S/N, but a 1/7 would be concerning.

python/lsst/cp/pipe/utils.py
```diff
@@ -764,3 +765,262 @@ def ddict2dict(d):
...
+        n_arr = np.zeros(len(self._nodes), dtype=int)
+        np.add.at(tot_arr, bins, ratio_model2)
+        np.add.at(n_arr, bins, 1)
+
+        ratio = np.ones(len(self._nodes))
+        ratio[n_arr > 0] = tot_arr[n_arr > 0]/n_arr[n_arr > 0]
+        ratio[0] = 1.0
+        p0[0: len(self._nodes)] = (ratio - 1) * self._nodes
+
+        # ratio_model3 = self.compute_ratio_model(
```
czwa, line 887: Add a comment explaining why if this should be retained.

python/lsst/cp/pipe/utils.py
```diff
@@ -764,3 +765,262 @@ def ddict2dict(d):
...
+    Sum[(S(mu_i) + mu_i)/(k_j * D_i) - 1]**2, where mu_i is the
+    observed flat-pair mean; D_j is the photo-diode measurement
+    corresponding to that flat-pair; and k_j is a constant of
+    proportionality which is over index j as it is allowed to
+    be different based on different photodiode settings (e.g.
+    CCOBCURR).
+
+    The fit has additional constraints to ensure that the spline
+    goes through the (0, 0) point, as well as a normalization
+    condition so that the average of the spline over the full
```
czwa, line 788: Can you explain this normalization condition more?  Is this trying to conserve charge?  I ask because if we attempted to spline fit something with a simple squared correction, it seems like this would cause problems.

czwa (approved):
Mostly documentation questions, but I have a few questions interspersed.  The new detailed tests are great.

python/lsst/cp/pipe/linearity.py
```diff
@@ -330,6 +345,19 @@ def run(self, inputPtc, dummy, camera, inputDims, inputPhotodiodeData=None,
             if self.config.applyPhotodiodeCorrection:
                 abscissaCorrections = inputPhotodiodeCorrection.abscissaCorrections
 
+        if self.config.linearityType == 'Spline':
+            if self.config.splineGroupingColumn is not None:
+                if self.config.splineGroupingColumn not in inputPtc.auxValues:
```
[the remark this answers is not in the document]
erykoff: So the answer is ... sure?  If you would like me to file a ticket so we add multiple groups to the other linearity fits, I can do that, but that's significantly more work and not urgent in any way.

python/lsst/cp/pipe/utils.py
```diff
@@ -764,3 +765,262 @@ def ddict2dict(d):
...
+class AstierSplineLinearityFitter:
+    """Class to fit the Astier spline linearity model.
+
+    This is a spline fit with photodiode data based on a model
+    from Pierre Astier.
+    This model fits a spline with (optional) nuisance parameters
+    to allow for different linearity coefficients with different
+    photodiode settings.  The minimization is a least-squares
+    fit with the residual of
+    Sum[(S(mu_i) + mu_i)/(k_j * D_i) - 1]**2, where mu_i is the
```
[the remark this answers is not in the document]
erykoff: So I can put in a link to the code, though I don't know how long it will live there.

python/lsst/cp/pipe/utils.py
erykoff: This is from Pierre's code, and what I think it's basically doing is what I said: the total sum of the spline term should be zero, so that the linear part is picked up by the linear fit.  I think this is basically the same thing that happens in the squared correction as well, as it's a squared deviation from linear, but the linear term is dropped in the application.

tests/test_linearity.py
erykoff: It's not, it's just at the large ones.  But the ratio comparison wasn't looking good.  The proof is in the linearity of the residuals, but I wanted to add some sort of test here.  If you think it's better to do different tolerances at low and high flux I can do that instead.

python/lsst/cp/pipe/utils.py
erykoff: I've added this comment:
```
    The fit has additional constraints to ensure that the spline
    goes through the (0, 0) point, as well as a normalization
    condition so that the average of the spline over the full
    range is 0. The normalization ensures that the spline only
    fits deviations from linearity, rather than the linear
    function itself which is degenerate with the gain.
```

python/lsst/cp/pipe/utils.py
```diff
@@ -764,3 +765,262 @@ def ddict2dict(d):
...
+            Array of photodiode measurements.
+        mu : `np.ndarray` (N,)
+            Array of flat means.
+        return_spline : `bool`, optional
+            Return the spline interpolation as well as the model ratios?
+
+        Returns
+        -------
+        ratio_models : `np.ndarray` (N,)
+            Model ratio, (mu_i + S(mu_i))/(k_j * D_i)
```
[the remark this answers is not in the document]
erykoff: Great catch!

python/lsst/cp/pipe/linearity.py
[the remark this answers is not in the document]
czwa: No, that's fine.  I mostly wanted to check that there's nothing spline-specific.

tests/test_linearity.py
czwa: Ok.  If my specific worry isn't justified, then this is fine with me.

PR #203, opened 2023-07-05
DM-32138: Remove exposureId from VisitInfo

parejkoj:


python/lsst/cp/pipe/utils.py
```diff
@@ -171,8 +171,8 @@ def makeMockFlats(expTime, gain=1.0, readNoiseElectrons=5, fluxElectrons=1000,
         flatExp1.image.array[:] = flatData1/gain   # ADU
         flatExp2.image.array[:] = flatData2/gain   # ADU
 
-    visitInfoExp1 = lsst.afw.image.VisitInfo(exposureId=expId1, exposureTime=expTime)
-    visitInfoExp2 = lsst.afw.image.VisitInfo(exposureId=expId2, exposureTime=expTime)
+    visitInfoExp1 = lsst.afw.image.VisitInfo(exposureTime=expTime)
+    visitInfoExp2 = lsst.afw.image.VisitInfo(exposureTime=expTime)
```
kfindeisen, line 175: Should this one use `id`? Seems a bit odd that it used to need an ID, but now it doesn't.

kfindeisen (approved):

python/lsst/cp/pipe/utils.py
parejkoj: The ids are set in the ExposureInfo just below, so the ids are preserved. It's not clear from context whether these were supposed to be detector+visit ids, or just visit ids, but looking at the use of this (in tests), it probably doesn't matter either way.

parejkoj: I also vaguely recall the exposureId field being the only required field early on, so this might be vestigial from that.

PR #204, opened 2023-07-10
DM-38944: Move photodiode computations from linearity to ptc code.

erykoff:
This moves the photodiode connections and integration code from the linearity code to the PTC code.  This will use the photoCharges part of the PTC dataset which was previously unfilled.

This also updates the default PTC and linearity pipelines for LSSTCam as appropriate.

### DIFF
--- a/tests/test_linearity.py
+++ b/tests/test_linearity.py
@@ -285,6 +285,12 @@ class LinearityTaskTestCase(lsst.utils.tests.TestCase):
             pd_values_offset[group2] *= pd_offset_factors[2]
             pd_values_offset[group3] *= pd_offset_factors[3]
 
+        # Add one bad photodiode value, but don't put it at the very
+        # end because that would change the spline node positions
+        # and make comparisons to the "truth" here in the tests
+        # more difficult.
+        pd_values_offset[-2] = np.nan
+
         ptc = self._create_ptc(
             self.amp_names,
             time_values,
--- a/tests/test_linearity.py
+++ b/tests/test_linearity.py
@@ -30,7 +30,7 @@ import numpy as np
 import lsst.utils
 import lsst.utils.tests
 
-from lsst.ip.isr import PhotonTransferCurveDataset, PhotodiodeCalib
+from lsst.ip.isr import PhotonTransferCurveDataset
 
 import lsst.afw.image
 import lsst.afw.math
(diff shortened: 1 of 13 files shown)


tests/test_linearity.py
```diff
@@ -85,6 +87,9 @@ def _create_ptc(self, amp_names, exp_times, means, ccobcurr=None):
         """
         exp_id_pairs = np.arange(len(exp_times)*2).reshape((len(exp_times), 2)).tolist()
 
+        if photo_charges is None:
+            photo_charges = np.zeros(len(exp_times)) + np.nan
```
czwa, line 91: Is this more efficient than `np.full(len(exp_times), np.nan)`?

czwa (approved):
One minor question.

tests/test_linearity.py
erykoff: Oddly enough, for small N the zeros is faster (but negligible time), for medium N full is faster, and for large N they're very close.  I don't think it matters much, I can change to full if you like.

erykoff: (small ~10; medium ~10000; large ~1000000)

czwa: I assumed they were both extremely optimized, I think I just find `full` easier to read.

tests/test_linearity.py
```diff
@@ -279,40 +285,21 @@ def _check_linearity_spline(self, do_pd_offsets=False):
...
-        for i, exp_id_pair in enumerate(exp_id_pairs):
-            time_samples = np.linspace(0, 20.0, 100)
-            current_samples = np.zeros(100)
-            current_samples[50] = -1.0*pd_values_offset[i]
-
-            pd_calib = PhotodiodeCalib(timeSamples=time_samples, currentSamples=current_samples)
-            pd_calib.currentScale = -1.0
-            pd_calib.integrationMethod = "CHARGE_SUM"
+        # Add one bad photodiode value, but don't put it at the very
+        # end because that will mess up the spline node positions.
```
czwa, line 289: Is this a problem that we need to check for and deal with?  I'm assuming it's not a trivial fix given this note.

czwa (approved):
Looks fine to me.  One question about the bad photodiode entry in the linearity tests.

tests/test_linearity.py
erykoff: Ah, sorry this is not clear.  The issue is simply that I want the spline positions in the test code ("truth") to match the spline positions in the data, which are computed from 0 to max.  So I don't want to mess with max or else the tests won't be comparing the same numbers.

PR #210, opened 2023-09-06
DM-37383: Add package to support DM processing of camera runs

czwa:
Use explicit names for the ISR task calls.

### DIFF
--- a/python/lsst/cp/pipe/defects.py
+++ b/python/lsst/cp/pipe/defects.py
@@ -876,14 +876,14 @@ class MergeDefectsCombinedConnections(pipeBase.PipelineTaskConnections,
         doc="Measured defect lists.",
         storageClass="Defects",
         dimensions=("instrument", "detector",),
-        multiple=False,
+        multiple=True,
     )
     inputFlatDefects = cT.Input(
         name="cpPartialDefectsFromFlatCombinedWithFilter",
         doc="Additional measured defect lists.",
         storageClass="Defects",
         dimensions=("instrument", "detector", "physical_filter"),
-        multiple=False,
+        multiple=True,
     )
     camera = cT.PrerequisiteInput(
         name='camera',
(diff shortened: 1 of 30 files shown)


erykoff (approved):
First of all, do like the fact that all these `isr` tasks are getting unique names; I think this will lead to less confusion (as well as enabling single pipelines without clashes).  However, I worry that this will break existing workflows (outside of cp_pipe).  Does this restructuring need an RFC? (Or at least a wide announcement).

python/lsst/cp/pipe/defects.py
```diff
@@ -605,7 +605,8 @@ def debugHistogram(self, stepname, ampImage, nSigmaUsed, exp):
             plt.close()
 
 
-class MeasureDefectsCombinedConnections(MeasureDefectsConnections,
+class MeasureDefectsCombinedConnections(pipeBase.PipelineTaskConnections,
+                                        # MeasureDefectsConnections,
```
erykoff, line 609: Is this comment necessary?

python/lsst/cp/pipe/defects.py
```diff
@@ -645,8 +646,9 @@ class MeasureDefectsCombinedTask(MeasureDefectsTask):
     _DefaultName = "cpDefectMeasureCombined"
 
 
-class MeasureDefectsCombinedWithFilterConnections(MeasureDefectsCombinedConnections,
-                                                  dimensions=("instrument", "detector")):
+class MeasureDefectsCombinedWithFilterConnections(pipeBase.PipelineTaskConnections,
+                                                  # MeasureDefectsCombinedConnections,
```
erykoff, line 650: What does this comment mean?

python/lsst/cp/pipe/defects.py
```diff
@@ -919,11 +921,28 @@ class MergeDefectsCombinedTask(MergeDefectsTask):
...
     def runQuantum(self, butlerQC, inputRefs, outputRefs):
         inputs = butlerQC.get(inputRefs)
-        # Turn inputFlatDefects and inputDarkDefects into a list
-        # which is what MergeDefectsTask expects.
-        tempList = [inputs['inputFlatDefects'], inputs['inputDarkDefects']]
+        # Turn inputFlatDefects and inputDarkDefects into a list which
+        # is what MergeDefectsTask expects.  If there are multiple,
+        # use the one with the most inputs.
+        tempList = [self.chooseBest(inputs['inputFlatDefects']),
+                    self.chooseBest(inputs['inputDarkDefects'])]
```
erykoff, line 944: So the decision is to just use the one with the most inputs, rather than try to merge all of them from the different filters?  Have you confirmed that this does something sensible?

erykoff (approved):

python/lsst/cp/pipe/defects.py
czwa: This comment and the previous one removed, as we're avoiding potential connection inheritance issues.

python/lsst/cp/pipe/defects.py
czwa: In the {{cp_testing}} processing, all flats are generated prior to these defects, so the task needs to handle the existence of multiple inputs.  I didn't want to simply {{OR}} all of the inputs, as the verify ticket for combined defects showed that the dust spots can trigger masking if they're fainter than the threshold.  {{OR}}ing these together means that the output defects may have all faint areas in all dust spots in all filters contributing to the masks.
I will admit that choosing the one with the most inputs is completely arbitrary.  However, this is somewhat artificial problem; In general processing, the flats would be inspected prior to use, and a single optimal input selected.  Only in the {{cp_testing}} case does this decision need to be made automatically.

PR #213, opened 2023-10-11
DM-40904: Certify updated calibrations based on combined defects

czwa:
Defects shouldn't be applied to image calibs.


plazas (approved):
LGTM!

PR #218, opened 2023-11-21
DM-41681: Implement function from P. Astier's code to subtract covariances at large lags in the PTC code of the LSST Science Pipelines.

plazas:


python/lsst/cp/pipe/ptc/cpSolvePtcTask.py
```diff
@@ -178,7 +207,7 @@ class PhotonTransferCurveSolveConfig(pipeBase.PipelineTaskConfig,
         doc="Use bootstrap for the PTC fit parameters and errors?.",
         default=False,
     )
-    binSize = pexConfig.Field(
+    binSide = pexConfig.Field(
```
erykoff, line 210: This is a typo

python/lsst/cp/pipe/ptc/cpSolvePtcTask.py
```diff
@@ -260,6 +289,21 @@ def run(self, inputCovariances, camera=None, detId=0):
...
         """
+        # Check parameters
+        fitMatrixSide = self.config.maximumRangeCovariancesAstierFullCovFit
+        measureMatrixSide = self.config.maximumRangeCovariancesAstier
+        if self.config.ptcFitType == "FULLCOVARIANCE":
+            if fitMatrixSide > measureMatrixSide:
+                raise RuntimeError("Covariance fit size %s is larger than"
+                                   "measurement size %s.",
+                                   fitMatrixSide, measureMatrixSide)
+            if self.config.doSubtractLongRangeCovariances:
```
czwa, line 300: If this only makes sense for the FULLCOVARIANCE model, then that can be done in the config validate as well.

python/lsst/cp/pipe/ptc/cpSolvePtcTask.py
```diff
@@ -500,7 +546,8 @@ def fitDataFullCovariance(self, dataset):
         have r^2 fewer entries.
         """
         matrixSide = self.config.maximumRangeCovariancesAstier
-        lenParams = matrixSide*matrixSide
+        matrixSideFit = self.config.maximumRangeCovariancesAstierFullCovFit
```
czwa, line 549: Should this parameter be pulled from the dataset itself?  It's already been set during initialization, right?

python/lsst/cp/pipe/ptc/cpSolvePtcTask.py
```diff
@@ -919,8 +1033,9 @@ def fitPtc(self, dataset):
         Astier+19 approximation (Eq. 16).
 
         Fit the photon transfer curve with either a polynomial of
-        the order specified in the task config, or using the
-        exponential approximation in Astier+19 (Eq. 16).
+        the order specified in the task config (POLNOMIAL),
```
czwa, line 1036: Typo.

python/lsst/cp/pipe/ptc/cpSolvePtcTask.py
```diff
@@ -955,22 +1070,23 @@ def fitPtc(self, dataset):
             ptcFitType = dataset.ptcFitType
         else:
             raise RuntimeError("ptcFitType is None of empty in PTC dataset.")
-        matrixSide = self.config.maximumRangeCovariancesAstier
-        nanMatrix = np.empty((matrixSide, matrixSide))
-        nanMatrix[:] = np.nan
+        # For FULLCOVARIANCE model fit
+        matrixSideFit = self.config.maximumRangeCovariancesAstierFullCovFit
```
czwa, line 1074: Isn't this going to be used for all model types?  If this is used for all model types, what do we gain by changing all of these variable names?

python/lsst/cp/pipe/utils.py
```diff
@@ -784,6 +785,94 @@ def ddict2dict(d):
     return dict(d)
 
 
+class Pol2D:
```
czwa, line 788: Can we use the afw/math polynomial code here instead?  Adding a duplicate class seems like a bad idea.

czwa (commented):
Some questions and suggestions.  It would also be good to have unit tests for this new functionality.

python/lsst/cp/pipe/ptc/cpExtractPtcTask.py
```diff
@@ -417,7 +417,7 @@ def run(self, inputExp, inputDims, taskMetadata, inputPhotodiodeData=None):
         # used to ensure that the number of output and input
         # dimensions match.
         dummyPtcDataset = PhotonTransferCurveDataset(ampNames, 'DUMMY',
-                                                     self.config.maximumRangeCovariancesAstier)
+                                                     self.config.maximumRangeCovariancesAstier, 1)
```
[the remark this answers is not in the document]
plazas: I added the keyword for the measured matrix size, and deleted the fit size, since now the default behavior takes care of it (if it's `None`, the measured one is assigned).

python/lsst/cp/pipe/ptc/cpSolvePtcTask.py
plazas: Just for the `FULLCOVARIANCE` model type.

python/lsst/cp/pipe/utils.py
plazas: I filed a ticket: DM-42196

czwa (approved):

PR #222, opened 2023-12-05
DM-42024: Allow for low gains and allow initial outlier cutoff to have units of electrons.

erykoff:
This fixes a bug where low gains (<0.5) were not allowed in the fit.  It also now changes `maxADUInitialPtcOutlierFit` to default of units of electrons that can then be scaled by an initial estimate of the gain.  This allows the cutoff to be approximately invariant across a range of gain settings.

### DIFF
--- a/python/lsst/cp/pipe/ptc/cpSolvePtcTask.py
+++ b/python/lsst/cp/pipe/ptc/cpSolvePtcTask.py
@@ -139,8 +139,17 @@ class PhotonTransferCurveSolveConfig(pipeBase.PipelineTaskConfig,
     maxSignalInitialPtcOutlierFit = pexConfig.Field(
         dtype=float,
         doc="Maximum signal considered for intial outlier fit. This should be below "
-            "the PTC turnoff to ensure accurate outlier rejection.",
-        default=30_000.,
+            "the PTC turnoff to ensure accurate outlier rejection. If "
+            "scaleMaxSignalInitialPtcOutlierFit=True then the units are electrons; "
+            "otherwise ADU.",
+        default=50_0000.,
+    )
+    scaleMaxSignalInitialPtcOutlierFit = pexConfig.Field(
+        dtype=bool,
+        doc="Scale maxSignalInitialPtcOutlierFit by approximate gain?  If yes then "
+            "maxSignalInitialPtcOutlierFit is assumed to have units of electrons, "
+            "otherwise ADU.",
+        default=True,
     )
     minVarPivotSearch = pexConfig.Field(
         dtype=float,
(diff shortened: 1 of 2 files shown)


python/lsst/cp/pipe/ptc/cpSolvePtcTask.py
```diff
@@ -1005,8 +1014,8 @@ def errFunc(p, x, y):
                 if self.config.binSize > 1:
                     bounds = self._boundsForAstier(parsIniPtc)
                 else:
-                    bounds = self._boundsForAstier(parsIniPtc, lowers=[-1e-4, 0.5, -2000],
```
plazas, line 1008: How necessary are these bounds for the convergence of this initial fit? In any case, I'm just wondering if even the new values (min=0.2, and max=3.0) will give us trouble in the future. Or if this code is used for other cameras with different settings (I think DECam has gains of 4, so I'm not even sure at this point how this has worked in the past for DECam...)

python/lsst/cp/pipe/ptc/cpSolvePtcTask.py
```diff
@@ -1022,6 +1031,18 @@ def errFunc(p, x, y):
             # This algorithm was initially developed by Seth Digel for
             # the EO Testing pipeline.
 
+            if self.config.scaleMaxSignalInitialPtcOutlierFit:
+                approxGain = np.nanmedian(meanVecOriginal/varVecOriginal)
+                maxADUInitialPtcOutlierFit = self.config.maxSignalInitialPtcOutlierFit/approxGain
+                self.log.info(
```
plazas, line 1037: We were concerned in another ticket about too much info in the logs, especially per amp. Will it be OK in this case? Should we change the level to "debug"?

plazas (approved):
LGTM, just a couple of questions.

python/lsst/cp/pipe/ptc/cpSolvePtcTask.py
erykoff: I don't think anybody has ever run this on DECam.  I'll change this to a bigger range.

python/lsst/cp/pipe/ptc/cpSolvePtcTask.py
erykoff: I did think about this, and could make it debug.  But it's not that much.  The difference is that this is a log that is done per amp in a PTC fit (16 messages per hundreds of flat pairs) and not a log per amp per ISR (16 messages times hundreds of flat pairs).

python/lsst/cp/pipe/ptc/cpSolvePtcTask.py
plazas: I did!

PR #226, opened 2024-01-12
DM-42160: eo_pipe/cp_verify parity: dark

czwa:
Add calibration distribution statistics.


python/lsst/cp/pipe/cpCombine.py
```diff
@@ -612,6 +615,33 @@ def setFilter(exp, filterLabel):
         """
         pass
 
+    def calibStats(self, exp, calibrationType):
```
arunkannawadi, line 618: This could be a static method, I guess.

python/lsst/cp/pipe/cpCombine.py
```diff
@@ -612,6 +615,33 @@ def setFilter(exp, filterLabel):
...
+                                                             [0, 5, 16, 50, 84, 95, 100])
+
+            metadata.set(f"LSST CALIB {calibrationType.upper()} {amp.getName()} DISTRIBUTION 0-PCT", p0)
+            metadata.set(f"LSST CALIB {calibrationType.upper()} {amp.getName()} DISTRIBUTION 5-PCT", p5)
+            metadata.set(f"LSST CALIB {calibrationType.upper()} {amp.getName()} DISTRIBUTION 16-PCT", p16)
+            metadata.set(f"LSST CALIB {calibrationType.upper()} {amp.getName()} DISTRIBUTION 50-PCT", p50)
+            metadata.set(f"LSST CALIB {calibrationType.upper()} {amp.getName()} DISTRIBUTION 84-PCT", p84)
+            metadata.set(f"LSST CALIB {calibrationType.upper()} {amp.getName()} DISTRIBUTION 95-PCT", p95)
+            metadata.set(f"LSST CALIB {calibrationType.upper()} {amp.getName()} DISTRIBUTION 100-PCT", p100)
+
```
arunkannawadi, line 644: Instead of hard-coding these, could we take the percentile values to compute as an input, and loop over those?

arunkannawadi (approved):

python/lsst/cp/pipe/cpCombine.py
czwa: You're right, hard coding this is likely to be annoying in the future.  I've moved the percentile levels to a config option, and have this method `zip` the levels and measured values into a second loop.  Then, if someone decides the quartiles are useful too, it's just a config change.

python/lsst/cp/pipe/cpCombine.py
czwa: Left alone, as this now uses a config parameter.

python/lsst/cp/pipe/cpCombine.py
timj: PS `metadata[key] = value` should work fine.

PR #227, opened 2024-01-17
DM-42238: Propagates list of flat pair read noises to ptcDataset

Alex-Broughton:


python/lsst/cp/pipe/ptc/cpSolvePtcTask.py
```diff
@@ -463,6 +466,17 @@ def run(self, inputCovariances, camera=None, detId=0):
...
+        # Initial validation of PTC fit.
+        for ampName in ampNames:
+            noise = np.nanmedian(datasetPtc.noiseList[ampName])**2
+            noiseFitted = datasetPtc.noise[ampName]
+
+            # Check if noise is close to noiseFitted
+            if not np.isclose(noiseFitted, noise, rtol=0.05, atol=0.0):
+                self.log.warning(f"Read noise from PTC fit ({noiseFitted}) is not consistent "
+                                 f"with read noise measured from overscan ({noise}) for "
+                                 f"amplifier {ampName}. Try adjusting the fit range.")
```
youtsumi, line 478: I like this suggestion!

python/lsst/cp/pipe/ptc/cpExtractPtcTask.py
```diff
@@ -534,6 +509,24 @@ def run(self, inputExp, inputDims, taskMetadata, inputPhotodiodeData=None):
...
+                thisTaskMetadata1 = taskMetadata[metadataIndex1]
+
+                expMetadata2 = expRef2.get(component="metadata")
+                metadataIndex2 = inputDims.index(expId2)
+                thisTaskMetadata2 = taskMetadata[metadataIndex2]
+
+                readNoise1[ampName] = self.getReadNoise(expMetadata1, thisTaskMetadata1, ampName)
+                readNoise2[ampName] = self.getReadNoise(expMetadata2, thisTaskMetadata2, ampName)
+
+                meanReadNoise[ampName] = np.nanmean([readNoise1[ampName], readNoise2[ampName]])
```
youtsumi, line 528: Initially I didn't understand why you need this change but now I probably do.  
Since you want to save each read noise at each flux level, you only have two measurement at each flux. Then there is no way to take median. That's why you need to switch to `nanmean`?

youtsumi (commented):
Increasing the number of comments to say why you need this change might help me to understand.

python/lsst/cp/pipe/ptc/cpSolvePtcTask.py
```diff
@@ -463,6 +466,17 @@ def run(self, inputCovariances, camera=None, detId=0):
             # PhotonTransferCurveDataset object.
             datasetPtc = self.fitMeasurementsToModel(datasetPtc)
 
+        # Initial validation of PTC fit.
+        for ampName in ampNames:
+            noise = np.nanmedian(datasetPtc.noiseList[ampName])**2
+            noiseFitted = datasetPtc.noise[ampName]
+
+            # Check if noise is close to noiseFitted
+            if not np.isclose(noiseFitted, noise, rtol=0.05, atol=0.0):
```
czwa, line 475: I think we want this comparison in ADU (dispersion-space) not ADU^2 (variance-space).  That matches the comment text better, and corresponds to how we use "read noise" elsewhere.

czwa (approved):
One small suggestion.

python/lsst/cp/pipe/ptc/cpExtractPtcTask.py
Alex-Broughton: Yep!

python/lsst/cp/pipe/ptc/cpSolvePtcTask.py
Alex-Broughton: Good suggestion

czwa (approved):

PR #229, opened 2024-01-20
Add PTC to linearity and BFK header provenance

Alex-Broughton:


python/lsst/cp/pipe/linearity.py
```diff
@@ -265,7 +270,28 @@ def runQuantum(self, butlerQC, inputRefs, outputRefs):
...
         inputs['inputDims'] = dict(inputRefs.inputPtc.dataId.required)
 
+        # Add calibration provenance info to header.
+        if self.config.doHeaderProvenance:
+
+            kwargs = dict()
+            reference = getattr(inputRefs, "inputPtc", None)
+            self.log.info(str(reference.run) + " and " + str(reference.id))
+            if reference is not None and hasattr(reference, "run"):
+                runKey = "PTCRUN"
```
czwa, line 280: It's annoying that astropy can't handle the long names.  Can these start with `PTC_` so they're somewhat easier to read?  They'll stay below 8 characters even with that.

python/lsst/cp/pipe/linearity.py
```diff
@@ -239,6 +239,11 @@ class LinearitySolveConfig(pipeBase.PipelineTaskConfig,
         doc="Maximum number of rejections per iteration for spline fit.",
         default=5,
     )
+    doHeaderProvenance = pexConfig.Field(
```
czwa, line 242: Let's not make this configurable.  Adding this information is always useful.

python/lsst/cp/pipe/makeBrighterFatterKernel.py
```diff
@@ -135,6 +135,11 @@ class BrighterFatterKernelSolveConfig(pipeBase.PipelineTaskConfig,
         doc="Slope of the correlation model for radii larger than correlationModelRadius",
         default=-1.35,
     )
+    doHeaderProvenance = pexConfig.Field(
```
czwa, line 138: Same as above.

python/lsst/cp/pipe/makeBrighterFatterKernel.py
```diff
@@ -159,9 +164,31 @@ def runQuantum(self, butlerQC, inputRefs, outputRefs):
...
+
+        # Add calibration provenance info to header.
+        if self.config.doHeaderProvenance:
+
+            for inputName in sorted(list(inputs.keys())):
+                kwargs = dict()
+                reference = getattr(inputRefs, inputName, None)
+
+                if reference is not None and hasattr(reference, "run"):
+                    runKey = f"LSST CALIB RUN {inputName.upper()}"
```
czwa, line 177: Does this not have the same issue as the linearity, or is the issue that the string "linearity" is long enough to trigger the issue?

python/lsst/cp/pipe/makeBrighterFatterKernel.py
```diff
@@ -548,3 +575,28 @@ def successiveOverRelax(self, source, maxIter=None, eLevel=None):
             self.log.info("Success: SuccessiveOverRelaxation converged in %s iterations."
                           "\noutError: %s, inError: %s", nIter//2, outError, inError*eLevel)
         return func[1: -1, 1: -1]
+
+    @staticmethod
+    def extractCalibDate(calib):
```
czwa, line 580: Move this function to the `utils.py` file, and call that in these two places.  It might be worth moving it to `isrFunctions.py` in the future, but that can be done on a future ticket (https://jira.lsstcorp.org/browse/DM-42622).

czwa (changes requested):
A few small changes.

python/lsst/cp/pipe/linearity.py
Alex-Broughton: I was following the same thing in `isrTask`, where doHeaderProvenance` is an option [here](https://github.com/lsst/ip_isr/blob/main/python/lsst/ip/isr/isrTask.py#L399) . Should there be another ticket to remove that as well?

python/lsst/cp/pipe/makeBrighterFatterKernel.py
Alex-Broughton: This function also appear [here](https://github.com/lsst/ip_isr/blob/main/python/lsst/ip/isr/isrTask.py#L399) as a static method of `isrTask`. Should I also remove it from there and import it into `isrTask.py`?

czwa (approved):

python/lsst/cp/pipe/linearity.py
czwa: No, but I acknowledge that I'm being a bit arbitrary.  Header provenance is something we may want to disable for alert production if we're not hitting the required processing times (even though the expected gain would be small).  These are being called during calib construction, when we theoretically have infinite time.

PR #235, opened 2024-02-29
DM-43022: Refactor calibration and defect production to ensure masking is orthogonal to calibration frames.

erykoff:


python/lsst/cp/pipe/cpCombine.py
```diff
@@ -464,7 +484,7 @@ def _subBBoxIter(bbox, subregionSize):
                                        (bbox, subregionSize, colShift, rowShift))
                 yield subBBox
 
-    def combine(self, target, expHandleList, expScaleList, stats):
+    def combine(self, target, expHandleList, expScaleList, stats, maskForNoData="BAD"):
```
czwa, line 487: Does `maskForNoData` get used elsewhere?  Was this a change from prior to adding `self.config.noGoodPixelsMask`?

pipelines/_ingredients/cpDarkForDefects.yaml
```diff
@@ -0,0 +1,31 @@
+description: cp_pipe DARK calibration construction for defect finding.
```
czwa, line 1: Can  this inherit from the default `cpDark.yaml`?  It'd be safer if we only needed to update one location, and that this contains only the difference from the standard cp dark processing.

python/lsst/cp/pipe/defects.py
```diff
@@ -89,6 +89,12 @@ class MeasureDefectsTaskConfig(pipeBase.PipelineTaskConfig,
              "hot/bright pixels in dark images. Unused if thresholdType==``STDEV``."),
         default=5,
     )
+    biasThreshold = pexConfig.Field(
```
czwa, line 92: I don't like that this does nothing in the `STDEV` case (meaning that entire task call is a NOOP), but I can't articulate why, so I think this is just a complaint with nothing to do.

czwa (approved):
No major issues.

pipelines/_ingredients/cpDarkForDefects.yaml
erykoff: I originally did that, but I first forgot to remove the `cpDark` part and it got all confused.  So we have to inherit, exclude, and override names and some configs and I thought that was more confusing to follow.  However, I could do that if you think it would be a net positive.

python/lsst/cp/pipe/cpCombine.py
erykoff: Good catch.  This was from before I realized I needed to set it in the `stats` object.

python/lsst/cp/pipe/defects.py
erykoff: I don't love the way that VALUE/STDEV is smushed into one, and I refrained from doing anything more drastic to refactor this part because it could go on forever.  I'll add a warning at least if it's called that way.

pipelines/_ingredients/cpDarkForDefects.yaml
czwa: If it's more complicated than a simple inherit, then it's fine to leave them separate, but please put a "if you edit this, you must also edit the other one" comments in both places so they can hopefully stay in sync.

czwa (approved):
This had a "review changes since last view" button, and I'm still happy with this new view.

PR #236, opened 2024-03-06
DM-43045: Stand-up per-camera cpSky configurations

leeskelvin:


python/lsst/cp/pipe/cpSkyTask.py
```diff
@@ -89,12 +89,7 @@ class CpSkyImageConfig(pipeBase.PipelineTaskConfig,
...
     def setDefaults(self):
-        # These values correspond to the HSC recommendation.  As noted
-        # below, the sizes are in millimeters, and correspond to an
-        # background image 8192x8192 pixels (8192*0.015=122.88).
-        self.largeScaleBackground.xSize = 122.88  # in mm
-        self.largeScaleBackground.ySize = 122.88  # in mm
-        self.largeScaleBackground.pixelSize = 0.015  # in mm per pixel
+        # These values correspond to the HSC recommendation.
         self.largeScaleBackground.minFrac = 0.1
         self.largeScaleBackground.mask = ['BAD', 'SAT', 'INTRP', 'DETECTED', 'DETECTED_NEGATIVE',
```
leeskelvin, line 94: If preferred, these could also be moved out of the task and into a pipelines `_ingredients` YAML, which would allow us to retire the `setDefaults` method here. I haven't done that at this time however, but I'm happy to do so if that reads better.

czwa: As these remaining fields are the same as the original config values, we can drop them here, as well as this method.

pipelines/Latiss/cpSky.yaml
```diff
@@ -10,3 +10,12 @@ tasks:
       doLinearize: false
       doCrosstalk: true
       doBrighterFatter: false
+  cpSkyImage:
+    class: lsst.cp.pipe.cpSkyTask.CpSkyImageTask
+    config:
+      # These values roughly split a single raft into 4x4 superpixels.
+      # As noted below, the sizes are in millimeters, and correspond to a
+      # background image of 6144*6144 pixels (6144*0.01=61.44).
+      largeScaleBackground.xSize: 61.44  # in mm
```
czwa, line 19: If these are known not to work (as shown by the figures on the ticket), then we should file a ticket to figure out correct values, and reference that ticket here and below (for just the cameras that need more development work).

czwa (approved):
A few suggestions in comments.

pipelines/Latiss/cpSky.yaml
czwa: But this one shouldn't be larger than the available pixels, probably.

leeskelvin: I've filed ticket DM-43321 to capture future effort on this, and left a note in each non-HSC YAML referencing this ticket.

All 6k superpixel configs have been reverted to their prior 4k dimensions on this ticket.

PR #238, opened 2024-04-17
DM-43377: Add more features to defect finding code and update LATISS pipeline.

erykoff:
This also changes the general defect pipeline to mark saturated pixels but not count them as bad/interpolate/anything so that the defect finding code can know which pixels were saturated.


czwa (approved):
Just one question on the inclusion of the `DETECTED` mask plane.

PR #239, opened 2024-04-18
DM-42766: Update linearity fitting code to fit temperature dependence and other updates recommended by Pierre Astier

erykoff:


python/lsst/cp/pipe/utils.py
```diff
@@ -1032,33 +1128,47 @@ def compute_ratio_model(nodes, group_indices, pars, pd, mu, return_spline=False)
...
         group_indices : `list` [`np.ndarray`]
             List of group indices, one array for each group.
+        par_indices : `dict`
+            Dictionary showing which indices in the pars belong to
+            each set of fit values.
         pars : `np.ndarray`
             Parameter array, with spline values (one for each node) followed
-            by proportionality constants (one for each group.)
+            by proportionality constants (one for each group.), followed by
+            (optionally) one offset O.
```
czwa, line 1137: Similar comment here: I don't think these docstrings were updated after the fit parameters were extended.

python/lsst/cp/pipe/utils.py
```diff
@@ -1075,7 +1185,7 @@ def fit(self, p0, min_iter=3, max_iter=20, max_rejection_per_iteration=5, n_sigm
         ----------
         p0 : `np.ndarray`
             Initial fit parameters (one for each knot, followed by one for
-            each grouping).
+            each grouping, and optionally one for the offset.)
```
czwa, line 1188: And here.

tests/test_linearity.py
```diff
@@ -324,6 +373,11 @@ def _check_linearity_spline(self, do_pd_offsets=False, n_points=200):
             self.input_dims,
         ).outputLinearizer
 
+        if do_weight_fit:
+            # These checks currently fail, and weight fitting is not
```
czwa, line 377: Is there a ticket to investigate/fix the weight fitting that could be added here?

python/lsst/cp/pipe/linearity.py
```diff
@@ -215,6 +220,42 @@ class LinearitySolveConfig(pipeBase.PipelineTaskConfig,
...
+    )
+    doSplineFitWeights = pexConfig.Field(
+        dtype=bool,
+        doc="Fit linearity weight parameters in the spline fit.",
+        default=False,
+    )
+    splineFitWeightParsStart = pexConfig.ListField(
+        dtype=float,
+        doc="Starting parameters for weight fit, if doSplineFitWeights=True. "
+            "Parameters are such that sigma = sqrt(par[0]**2. + par[1]**2./mu).",
```
czwa, line 236: I have questions about this, but after reading the rest of the changes (and the test comment), it seems like these aren't being fit as part of the standard pipeline code, so I think my questions aren't relevant at the moment.

czwa (approved):
Mostly docstring comments.

python/lsst/cp/pipe/linearity.py
```diff
@@ -463,13 +525,18 @@ def run(self, inputPtc, dummy, camera, inputDims,
                 # to allow for different linearity coefficients with different
                 # photodiode settings.  The minimization is a least-squares
                 # fit with the residual of
-                # Sum[(S(mu_i) + mu_i)/(k_j * D_i) - 1]**2, where S(mu_i) is
-                # an Akima Spline function of mu_i, the observed flat-pair
+                # Sum[(S(mu_i) + mu_i - O)/(k_j * D_i) - 1]**2, where S(mu_i)
+                # is an Akima Spline function of mu_i, the observed flat-pair
                 # mean; D_j is the photo-diode measurement corresponding to
```
czwa, line 530: `D_i`?

python/lsst/cp/pipe/linearity.py
```diff
@@ -463,13 +525,18 @@ def run(self, inputPtc, dummy, camera, inputDims,
...
                 # that flat-pair; and k_j is a constant of proportionality
                 # which is over index j as it is allowed to
                 # be different based on different photodiode settings (e.g.
-                # CCOBCURR).
+                # CCOBCURR); and O is a constant offset to allow for light
```
czwa, line 534: Is this `O` a single value for all pairs?  I'll find out when I get to the fitting code, but I'm just checking if this needs any index as well.

python/lsst/cp/pipe/linearity.py
```diff
@@ -565,6 +662,13 @@ def run(self, inputPtc, dummy, camera, inputDims,
 
             linearizer.fitResiduals[ampName] = residuals
 
+            finite = np.isfinite(residuals)
+            if finite.sum() == 0:
+                sigmad = np.nan
+            else:
+                sigmad = median_abs_deviation(residuals[finite]/inputOrdinate[finite], scale="normal")
+            linearizer.fitResidualsSigmaMad[ampName] = sigmad
```
czwa, line 670: I'm not clear on the logic here.  Is this checking for NaN (in which case any NaN value will result in `sigmad=NaN`), or is this checking for infinite values (which I see `scipy.stats.median_abs_deviation` does handle).

python/lsst/cp/pipe/utils.py
```diff
@@ -1032,33 +1131,50 @@ def compute_ratio_model(nodes, group_indices, pars, pd, mu, return_spline=False)
...
         )
 
-        numerator = mu - spl.interpolate(mu)
+        # Check if we want to do just the left or both with temp scale.
+        if len(par_indices["temperature_coeff"]) == 1:
+            mu_corr = mu*(1. + pars[par_indices["temperature_coeff"]]*temperature_scaled)
+        else:
+            mu_corr = mu
+
+        numerator = mu_corr - spl.interpolate(mu_corr)
```
czwa, line 1172: Does this line contradict the docstrings that list `S(mu_i) + mu_i - O`?  It's also entirely possible I've confused myself.

czwa: I think I've convinced myself that I am not confused in this case.

python/lsst/cp/pipe/linearity.py
czwa: And I agree now that this is `O`, with no indexing.

czwa (approved):
A few comments as I worked my way through the fitting process, with only docstring changes I think.

czwa:
I continue to be happy with approval, including the two most recent commits (on failed amp/length matching),

PR #240, opened 2024-04-24
DM-44025: Update ptc turnoff computation code and improve tests with real data.

erykoff:
This also adds a computation of the "sampling error" on the ptc turnoff.


python/lsst/cp/pipe/ptc/cpSolvePtcTask.py
```diff
@@ -174,11 +174,21 @@ class PhotonTransferCurveSolveConfig(pipeBase.PipelineTaskConfig,
             "otherwise ADU.",
         default=50_000.,
     )
+    maxDeltaInitialPtcOutlierFit = pexConfig.Field(
+        dtype=float,
+        doc="If there are any outliers in the initial fit above "
+            "maxSignalInitialPtcOutlierFit, then no points that have this delta "
+            "from the previous ``good`` point are allowed. If "
```
czwa, line 181: Can you clarify that this delta is in terms of the `rawMeans` vector (or `finalMeans`).  The units being to first order implies this (so that it's not based on the variance).

python/lsst/cp/pipe/ptc/cpSolvePtcTask.py
```diff
@@ -1236,8 +1276,10 @@ def errFunc(p, x, y):
 
                     count += 1
```
czwa, line 1277: Do we not warn if we've reached the iteration threshold and are still finding outliers?  I'm guessing we converge very quickly, so it's possible 2 is always sufficient, but I wanted to check if this was an oversight.

tests/test_ptc.py
```diff
@@ -679,6 +679,45 @@ def ptcFitAndCheckPtc(
...
+                configSolve = copy.copy(self.defaultConfigSolve)
+                configSolve.doFitBootstrap = False
+                configSolve.doLegacyTurnoffSelection = False
+
+                solveTask = cpPipe.ptc.PhotonTransferCurveSolveTask(config=configSolve)
+
+                solvedDataset = solveTask.fitMeasurementsToModel(dataset)
+
+                # Check that the ptcTurnoff is what is expected.
+                self.assertFloatsAlmostEqual(solvedDataset.ptcTurnoff[ampName], ptcTurnoff)
```
czwa, line 710: Can you add `msg` arguments to all of these asserts with the `dense` and `mode` parameters?  That should help identify problems when these tests fail in the future.

tests/test_ptc.py
```diff
@@ -923,6 +962,694 @@ def test_ptcFitBootstrap(self):
...
+            Array of raw mean values.
+        rawVars : `np.ndarray`
+            Array of raw variance values.
+        ptcTurnoff : `float`
+            PTC turnoff determined from fitting by eye.
+        """
+        if dense and mode == "normal":
+            # Taken from dense run 13591, detector 94, amplifier C02
+            ptcTurnoff = 92239.4794
+            rawMeans = np.array([
```
czwa, line 991: I have a slight preference for storing these as external PTC dataset files in `tests/data`, but it's probably not worth the effort on this ticket.

czwa (approved):
Just a few clarifications.

python/lsst/cp/pipe/ptc/cpSolvePtcTask.py
erykoff: We never have, but I can add a warning.

tests/test_ptc.py
erykoff: Since we didn't have a test data directory I was too lazy to make one ...

PR #242, opened 2024-05-02
DM-44114: Add filter scan task to cp_pipe

czwa:
This adds CpFilterScanTask to create filter scan reports.
There's a small change to the `combineHeaders` method to allow it to populate the metadata independent of the final output calib type.


pipelines/_ingredients/cpFilterScan.yaml
```diff
@@ -0,0 +1,26 @@
...
+      doBias: true
+      doDark: true
+      doInterpolate: true
+      doSetBadRegions: true
+  cpFilterScanMerge:
+    class: lsst.cp.pipe.cpFilterScan.CpFilterScanTask
+    config:
+      connections.inputExpHandles: 'cpFilterScanIsrExp'
+      # This is still an intermediate product, although it's the end of this pipeline.
+      connections.outputData: 'cpFilterScan'
```
erykoff, line 26: I was going to comment that I don't like the name `cpFilterScan` as a final product but you say this is an intermediate product so 👍

python/lsst/cp/pipe/cpFilterScan.py
```diff
@@ -0,0 +1,377 @@
...
+    )
+    useElectrometer = pexConfig.Field(
+        dtype=bool,
+        doc="Use electrometer data?",
+        default=False,
+    )
+    referenceFilter = pexConfig.Field(
+        dtype=str,
+        doc="Filter to use as baseline reference.",
+        default="empty~empty",
```
erykoff, line 86: I know this is the default right now for LATISS, but I think the dual naming is LATISS-only.  I would leave this as `empty` or better yet make it required to be overridden in the instrumental pipelines.

python/lsst/cp/pipe/cpFilterScan.py
```diff
@@ -0,0 +1,377 @@
...
+
+        for handle in inputExpHandles:
+            visitInfo = handle.get(component="visitInfo")
+            metadata = handle.get(component="metadata")
+
+            physical_filter = handle.dataId['physical_filter']
+            exposureId = handle.dataId['exposure']
+            exposureTime = visitInfo.exposureTime
+
+            values = [v for k, v in metadata.items() if "LSST ISR FINAL MEAN" in k]
```
erykoff, line 147: I don't love using the `MEAN` here over the full frame.  Either the `median` or (like the PTC) we need to compute the mean with excluding the edges.  (Though it may be that because we're computing a ratio with and without the filter it cancels, I still don't like using the MEAN).

python/lsst/cp/pipe/cpFilterScan.py
```diff
@@ -0,0 +1,377 @@
...
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
```
erykoff, line 20: A module or (probably) class docstring which explains the algorithm would be very useful.

python/lsst/cp/pipe/cpFilterScan.py
```diff
@@ -0,0 +1,377 @@
...
+            scan = {
+                'physical_filter': physical_filter,
+                'scale': scaleFactor,
+                'flux': np.median(values),
+            }
+            filterSet.add(physical_filter)
+
+            _, key = efdClient.parseMonochromatorStatus(monochromatorData,
+                                                        visitInfo.date.toString(DateTime.TAI))
+            key = float(key)
```
erykoff, line 167: I think that `key` should be `wavelength` or `wavelengthKey`

python/lsst/cp/pipe/cpFilterScan.py
```diff
@@ -0,0 +1,377 @@
...
+            if key in filterScanResults:
+                filterScanResults[key].append(scan)
+            else:
+                filterScanResults[key] = [scan]
+
+        filterScan = []
+        for wavelength in filterScanResults.keys():
```
erykoff, line 174: So now we have a dict keyed by wavelength, and each will have (probably) two entries, with and without the filter?

python/lsst/cp/pipe/cpFilterScan.py
```diff
@@ -0,0 +1,377 @@
...
+            scans = filterScanResults[wavelength]
+
+            referenceScan = [x for x in scans if x['physical_filter'] == self.config.referenceFilter]
+            if len(referenceScan) == 0:
+                # No reference scan at this wavelength.
```
erykoff, line 179: This should probably log a warning, since the plan would be to have pairs, correct?

python/lsst/cp/pipe/cpFilterScan.py
```diff
@@ -0,0 +1,377 @@
...
+            referenceScan = referenceScan[0]
+            referenceValue = referenceScan['scale'] / referenceScan['flux']
+
+            wavelengthScan = {'wavelength': float(wavelength)}
+            for filterName in filterSet:
+                wavelengthScan[filterName] = np.nan
+
+            for scan in scans:
+                if np.isfinite(wavelengthScan[scan['physical_filter']]):
+                    self.log.warn(f"Multiple instances of filter: {scan['physical_filter']}")
```
erykoff, line 190: This needs to be `log.warning()` not deprecated `log.warn()`

python/lsst/cp/pipe/cpFilterScan.py
```diff
@@ -0,0 +1,377 @@
...
+            scans = filterScanResults[wavelength]
+
+            referenceScan = [x for x in scans if x['physical_filter'] == self.config.referenceFilter]
+            if len(referenceScan) == 0:
+                # No reference scan at this wavelength.
+                continue
+            referenceScan = referenceScan[0]
+            referenceValue = referenceScan['scale'] / referenceScan['flux']
+
+            wavelengthScan = {'wavelength': float(wavelength)}
```
erykoff, line 184: I'm having trouble parsing what these dicts are supposed to contain.  Below I see that it always sets the wavelengthScan to nan and then checks if it's finite and how is it ever finite?

python/lsst/cp/pipe/cpFilterScan.py
```diff
@@ -0,0 +1,377 @@
...
+    peakBoxSize = pexConfig.Field(
+        dtype=int,
+        doc="Half-size of the box used for fitting the spectrum peak.",
+        default=50,
+    )
+
+
+class CpMonochromatorScanTask(pipeBase.PipelineTask):
+    """Compare EFD monochromator results to fiber spectrograph spectra.
+    """
```
erykoff, line 250: Is this used in any of the pipelines?  I don't know what this task is.

python/lsst/cp/pipe/cpFilterScan.py
```diff
@@ -0,0 +1,377 @@
...
+                'peak': peakWavelength,
+                'fwhm': fwhm,
+                'fit_range': fitRange,
+                'exp_date': date,
+                'mono_date': monoDate.isoformat(),
+                'mono_state': monoValue,
+            }
+            monochromatorScanResults.append(entry)
+
+        # Will this silence the errors?
```
erykoff, line 298: I must know!  Will it?

python/lsst/cp/pipe/utilsEfd.py
```diff
@@ -0,0 +1,151 @@
...
+
+        Returns
+        -------
+        indexDate : `str`
+            Date string indicating the monochromator state change.
+        wavelength : `float`
+            Monochromator commanded peak.
+        """
+        dateValue = Time(dateStr, format='isot', scale='utc')
+        # I have words here.
```
erykoff, line 124: ??

python/lsst/cp/pipe/utilsEfd.py
```diff
@@ -0,0 +1,151 @@
...
+        if data.index[0] > dateValue or data.index[-1] < dateValue:
+            raise RuntimeError("Requested date is outside of data range.")
+
+        # I have more words.
```
erykoff, line 128: ???

erykoff (commented):
I have words (and questions)...

pipelines/_ingredients/cpFilterScan.yaml
czwa: My thought is that these are results that can be converted into `transmission_filter` objects, likely using the monochromator scans as well, but that is beyond the scope of this task and ticket.

python/lsst/cp/pipe/cpFilterScan.py
czwa: Done here, and duplicated for `CpMonochromatorScanTask`.

python/lsst/cp/pipe/cpFilterScan.py
czwa: I've switched to `empty` so it's more generic.

python/lsst/cp/pipe/cpFilterScan.py
czwa: Yes, that's the intention.

python/lsst/cp/pipe/cpFilterScan.py
czwa: Definitely, this is an oversight.  I've also updated the other warning with the wavelength.

python/lsst/cp/pipe/cpFilterScan.py
czwa: After the first loop, we have a `filterScanResults` dictionary, keyed by the wavelength of the monochromator setting, and containing a list of scans.  Each scan in that list maps to one input exposure, and contains the filter of that exposure, the measured exposure flux, and a scale factor to normalize exposure time and the electrometer's measurement of the illumination.
This second loop then iterates over all scans taken at a given monochromator wavelength, with the intention of pairing them up.  In the generic case, there can be any number of filters taken at that wavelength, so the set of all filters observed at that wavelength is determined, and the throughput of all found filters at that wavelength are set to NaN.  Ideally, the wavelengthScan looks like `{wavelength: 12345, g: NaN, r: NaN, i: NaN, empty: NaN}`.  The code then loops over the entries of `scans`, and if the value for a particular filter is NaN, that filter has not yet been filled, and is then set to the appropriate throughput.  If the value for that filter is not NaN, then a previous scan already populated that field, and the log message is triggered.  This task has no way to know which of the multiple scans are correct, but if there are matching electrometer values for everything, they should normalize to the same value.  There's something of a check in that `empty` is both a filter with a scan and the reference, so the throughput of the reference filter should be universally 1.0.

python/lsst/cp/pipe/cpFilterScan.py
czwa: Documentation updated as mentioned above.  This is used in an `_ingredients/cpMonochromatorScan.yaml`, which I did not add a camera specific task for, as we currently only have one fiber spectrograph (and so the config is specific to that instrument).

python/lsst/cp/pipe/cpFilterScan.py
czwa: The answer is yes, and the comment deleted.

python/lsst/cp/pipe/utilsEfd.py
czwa: "Check that the date we want to consider is contained in the EFD data."  Writing code me and remembering to fix my comments me didn't meet that day.

python/lsst/cp/pipe/utilsEfd.py
czwa: "Binary search through the EFD entries in date, until the most recent monochromator state update prior to the spectrum in question is found."
If there's a better way to search through a pandas dataframe, I'm happy to use that, but I was unable to find one with some searching.

python/lsst/cp/pipe/cpFilterScan.py
```diff
@@ -0,0 +1,417 @@
...
+    )
+
+
+class CpFilterScanTask(pipeBase.PipelineTask):
+    """Create filter scan from appropriate data.
+
+    This task constructs a filter scan from pairs of flat exposures,
+    one with a filter in the beam and one without a filter.  A
+    monochromator is also included in the beam for both exposures, so
+    that the beam can be limited to a narrow band of wavelengths small
```
erykoff, line 105: small -> smaller

python/lsst/cp/pipe/cpFilterScan.py
```diff
@@ -0,0 +1,417 @@
...
+    than the full bandpass of the filter.
+
+    From these pairs of exposures, we can determine the filter
+    throughput by calculating the flux per second with the filter:
+        F_filter(\lambda0) = median(f_amplifiers) / t_exposure
```
erykoff, line 110: These are errors that need to be fixed with either an `r` in front of the triple backtick, or perhaps double backslash?

python/lsst/cp/pipe/cpFilterScan.py
```diff
@@ -0,0 +1,417 @@
...
+    We can correct for any illumination changes by optionally using
+    the electrometer measurements, E, which provide an independent
+    measure of the incident flux for the two exposures, such that:
+        throughput(\lambda0) = throughput_raw * E_reference / E_filter
+
+    Repeating this procedure at multiple monochromator settings builds
+    up a catalog of throughput measurements across the filter
+    bandpass.  Additional differences between the monochromator
+    setting (retrieved here from the EFD) and the actual wavelengths
+    of light that are permitted can exist, so a matching
```
erykoff, line 127: "the actual wavelengths of light that are permitted can exist" doesn't make sense.

python/lsst/cp/pipe/cpFilterScan.py
erykoff: Can you add a comment to this effect in the code?

python/lsst/cp/pipe/cpFilterScan.py
erykoff: This is all wonderfully clear and very helpful.  If you could put it into the code comments then in 6 months we'll have a lot easier time remembering what it is doing.

python/lsst/cp/pipe/cpFilterScan.py
czwa: Switching to MEDIAN (which should also already be in the headers).

python/lsst/cp/pipe/cpFilterScan.py
czwa: Additional comments added.

python/lsst/cp/pipe/cpFilterScan.py
czwa: Let me know if further comments would help.

python/lsst/cp/pipe/cpFilterScan.py
```diff
@@ -0,0 +1,443 @@
...
+            for filterName in filterSet:
+                wavelengthScan[filterName] = np.nan
+
+            for scan in scans:
+                # If the entry for this filter already exists, then we
+                # have multiple entries at this wavelength.
+                if np.isfinite(wavelengthScan[scan['physical_filter']]):
+                    self.log.warning(
+                        f"Multiple instances of filter {scan['physical_filter']} at {wavelength}"
+                    )
```
erykoff, line 243: I guess right now it just overrides with the "last" one?

python/lsst/cp/pipe/cpFilterScan.py
```diff
@@ -0,0 +1,443 @@
...
+
+                wavelengthScan[scan['physical_filter']] = referenceValue * scan['flux'] / scan['scale']
+
+            filterScan.append(copy.copy(wavelengthScan))
+
+        outMetadata = PropertyList()
+        self.combine.combineHeaders(inputExpHandles, calib=None, metadata=outMetadata,
+                                    calibType="FilterScan", scales=None)
+        filteredMetadata = {k: v for k, v in outMetadata.toDict().items() if v is not None}
+
```
erykoff, line 253: Adding a comment here for what fields we will have in the final `filterScan` would be really useful for the future/downstream processing.

PR #243, opened 2024-05-21
DM-41442: Add focal plane mosaic construction to cp_pipe/cp_verify

czwa:
Update pipelines to call binning/mosaic tasks.

### DIFF
--- a/pipelines/_ingredients/cpBias.yaml
+++ b/pipelines/_ingredients/cpBias.yaml
@@ -46,14 +46,14 @@ tasks:
   cpBiasMosaic1:
     class: lsst.pipe.tasks.visualizeVisit.VisualizeMosaicCalibTask
     config:
-      connections.inputExp: 'biasBin1'
-      connections.outputExp: 'biasMosaic1'
+      connections.inputExps: 'biasBin1'
+      connections.outputData: 'biasMosaic1'
       binning: 8
   cpBiasMosaic2:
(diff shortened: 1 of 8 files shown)


pipelines/_ingredients/cpBias.yaml
```diff
@@ -31,7 +31,33 @@ tasks:
       calibrationType: 'bias'
       exposureScaling: "Unity"
       mask: ["DETECTED", "INTRP"]
+  cpBiasBin1:
+    class: lsst.pipe.tasks.visualizeVisit.VisualizeBinCalibTask
+    config:
+      connections.inputExp: 'bias'
+      connections.outputExp: 'biasBin1'
+      binning: 8
```
erykoff, line 39: Can we call this `biasBin8`?

pipelines/_ingredients/cpBias.yaml
```diff
@@ -31,7 +31,33 @@ tasks:
...
+  cpBiasBin2:
+    class: lsst.pipe.tasks.visualizeVisit.VisualizeBinCalibTask
+    config:
+      connections.inputExp: 'bias'
+      connections.outputExp: 'biasBin2'
+      binning: 64
```
erykoff, line 45: And this `biasBin64` (and so on)

erykoff (commented):
I think the names could be more evocative, including the binning rather than an arbitrary index would be useful.  I don't know if it should be called `8` or `08` though.

PR #245, opened 2024-06-13
DM-44259: Update pipeline names and code for RFC-1013

erykoff:


python/lsst/cp/pipe/ptc/__init__.py
```diff
@@ -21,6 +21,6 @@
 # see <https://www.lsstcorp.org/LegalNotices/>.
 #
 
-from .cpExtractPtcTask import *
-from .cpSolvePtcTask import *
-from .cpPlotPtcTask import *
+from .cpPtcExtract import *
```
czwa, line 24: To be consistent, I'd suggest these be alphabetized as well.

tests/test_pipelines.py
```diff
@@ -0,0 +1,141 @@
...
+    has_obs_decam = False
+
+
+class CalibrationPipelinesTestCase(lsst.utils.tests.TestCase):
+    """Test case for building the pipelines."""
+
+    def setUp(self):
+        self.pipeline_path = os.path.join(lsst.utils.getPackageDir("cp_pipe"), "pipelines")
+
+    def _get_pipelines(self, exclude=[]):
```
czwa, line 56: I think this either needs to be able to auto-discover pipelines (possibly with a `glob(os.path.join(self.pipeline_path, "*", "*.yaml"))`, or we need a note in the `README.md` that newly added pipelines need to be added to this list.
Hopefully we won't have many more to add, but we don't want things to get out of sync.

czwa (approved):
One likely typo, two "alphabetize __init__"s, and a comment on the new `test_pipelines.py`.

pipelines/LSSTCam-imSim/fpDefectsIndividual.yaml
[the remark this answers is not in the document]
erykoff: That should not have been there at all.

PR #246, opened 2024-06-15
DM-44259-hotfix-docs: Remove extraneous paths from task names in docs.

erykoff:
The tasks are all available in the `lsst.cp.pipe` namespace and the docs need to reference them there.


ktlim (approved):
Since it builds locally, 👍

PR #247, opened 2024-06-17
DM-44548  Patches for calibration rehearsal 1 (CR1)

czwa:
Patches that were needed for the calibration rehearsal, that we now want to include on main:

- Adds LSSTComCamSim specific pipeline definitions.
- Adds LSSTComCamSim to the pipeline unit tests.


erykoff (approved):
Looks good!

PR #251, opened 2024-07-08
DM-45135: Update task names in cp_pipe yaml files for LSSTCam

jchiang87:


timj:
Can I push a quick numpy 2 fix to this branch to fix this:

```
cp_pipe/python/lsst/cp/pipe/ptc/cpPtcSolve.py:971:23: NPY201 [*] `np.NINF` will be removed in NumPy 2.0. Use `-np.inf` instead.
cp_pipe/python/lsst/cp/pipe/ptc/cpPtcSolve.py:980:23: NPY201 [*] `np.NINF` will be removed in NumPy 2.0. Use `-np.inf` instead.```

?

czwa (approved):
Thanks for catching this inconsistency.
Numpy fix looks good as well.

PR #253, opened 2024-07-19
DM-45300: Begin deprecation of task metadata in cp_pipe/cp_verify

czwa:
Move `getReadNoise` to lsst.cp.pipe.utils, and make it able to find read noise entries in either the exposure or task metadata.

### DIFF
--- a/python/lsst/cp/pipe/ptc/cpPtcExtract.py
+++ b/python/lsst/cp/pipe/ptc/cpPtcExtract.py
@@ -30,7 +30,7 @@ import lsst.pipe.base as pipeBase
 from lsst.geom import (Box2I, Point2I, Extent2I)
 from lsst.cp.pipe.utils import (arrangeFlatsByExpTime, arrangeFlatsByExpId,
                                 arrangeFlatsByExpFlux, sigmaClipCorrection,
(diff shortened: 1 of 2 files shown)


python/lsst/cp/pipe/ptc/cpPtcExtract.py
```diff
@@ -515,16 +515,14 @@ def run(self, inputExp, inputDims, taskMetadata, inputPhotodiodeData=None):
                 readNoise2 = dict()
                 meanReadNoise = dict()
 
-                expMetadata1 = expRef1.get(component="metadata")
```
erykoff, line 518: I now see that we already had `exp1` and `exp2` loaded so this was just extra i/o.

erykoff (approved):
Are we planning on deprecating the metadata input to the ptc extractor as well?

python/lsst/cp/pipe/ptc/cpPtcExtract.py
czwa: I definitely want to deprecate all `taskMetadata` inputs, so I've added the notice to that connection as well.

PR #254, opened 2024-07-22
DM-45322: Correctly handle linearity bboxes for both trimmed and untrimmed exposures

czwa:


python/lsst/cp/pipe/cpLinearitySolve.py
```diff
@@ -152,6 +152,13 @@ class LinearitySolveConfig(pipeBase.PipelineTaskConfig,
         doc="Number of spline knots to use in fit.",
         default=10,
     )
+
+    trimmedState = pexConfig.Field(
```
erykoff, line 156: I don't think that this is quite the right name.  But then again the docs are clear, and I don't have a better suggestion off the top of my head.

erykoff (approved):
Minor comment: I think the PR title should describe the PR not the motivation (e.g. "Add configuration to linearizer creation to specify if it will be used on un/trimmed exposures")

PR #256, opened 2024-07-30
DM-44826: Add EFD query support for electrometer data.

czwa:


python/lsst/cp/pipe/utilsEfd.py
```diff
@@ -202,11 +206,67 @@ def selectTimeSeries(self, topicName, fields=[], startDate=None, endDate=None):
         table = Table(rows=series["values"], names=series["columns"], dtype=tableDtype)
         table["time"] = Time(table["time"], scale="utc")
         table.sort("time")
+        if 'private_sndStamp' in table.columns:
```
erykoff, line 209: Please use consistent double quotes here and below.  (Sorry to be difficult)

python/lsst/cp/pipe/utilsEfd.py
```diff
@@ -202,11 +206,67 @@ def selectTimeSeries(self, topicName, fields=[], startDate=None, endDate=None):
         table = Table(rows=series["values"], names=series["columns"], dtype=tableDtype)
         table["time"] = Time(table["time"], scale="utc")
         table.sort("time")
```
erykoff, line 208: Can you add a comment here about what the `time` column is and what the `private_sndStamp` column is and why we would like one over the other?

python/lsst/cp/pipe/utilsEfd.py
```diff
@@ -202,11 +206,67 @@ def selectTimeSeries(self, topicName, fields=[], startDate=None, endDate=None):
...
+        if 'private_sndStamp' in table.columns:
+            table['private_sndStamp'] = Time(table['private_sndStamp'], format='unix_tai')
+            table.sort("private_sndStamp")
 
         return table
 
+    def searchResults(self, data, dateStr, scale='tai'):
+        """Determine the entry for a specific date.
```
erykoff, line 216: This docstring is too vague to be useful.

python/lsst/cp/pipe/utilsEfd.py
```diff
@@ -202,11 +206,67 @@ def selectTimeSeries(self, topicName, fields=[], startDate=None, endDate=None):
...
+        scale : `str`, optional
+            Time scale to use.  Default is 'tai'.
+
+        Returns
+        -------
+        result = `astropy.table.Row`
+            The row of the data table corresponding to ``dateStr``.
+        """
+        dateValue = Time(dateStr, format='isot', scale=scale)
+        # Table is now sorted on "time", which is in UTC.
```
erykoff, line 233: I thought it was TAI?

python/lsst/cp/pipe/utilsEfd.py
```diff
@@ -202,11 +206,67 @@ def selectTimeSeries(self, topicName, fields=[], startDate=None, endDate=None):
...
+        # most recent monochromator state update prior to the spectrum
+        # in question is found.
+        low = 0
+        high = len(data)
+        idx = (high + low) // 2
+        found = False
+        iteration = 0
+        while not found:
+            if idx < 0 or idx > len(data) or iteration > 20:
+                raise RuntimeError("Search for date failed?")
```
erykoff, line 250: Question?  Also maybe put in the specific dateValue here.

python/lsst/cp/pipe/utilsEfd.py
```diff
@@ -202,11 +206,67 @@ def selectTimeSeries(self, topicName, fields=[], startDate=None, endDate=None):
...
+            if myTime <= dateValue:
+                low = idx
+            elif myTime > dateValue:
+                high = idx
+
+            idx = (high + low) // 2
+            iteration += 1
+            if high - low == 1:
+                found = True
+            self.log.info("parse search %d %d %d %d %s %s",
```
erykoff, line 262: Is this an info or a debug?

python/lsst/cp/pipe/utilsEfd.py
```diff
@@ -219,7 +279,7 @@ def getEfdMonochromatorData(self, dataSeries=None, dateMin=None, dateMax=None):
 
         Returns
         -------
-        results : `pandas.DataFrame`
+        results : `astropy.table.Table`
```
erykoff, line 282: How did this datatype ever make it through.  Thanks for fixing it!

python/lsst/cp/pipe/utilsEfd.py
```diff
@@ -257,41 +315,163 @@ def parseMonochromatorStatus(self, data, dateStr):
...
+            Maximum date to retrieve from EFD.
 
-            myTime = data["private_sndStamp"][idx]
+        Returns
+        -------
+        results : `astropy.table.Table`
+            The table of results returned from the EFD.
+        """
+        # All electrometer data gets written to the same series.
+        defaultSeries = "lsst.sal.Electrometer.logevent_intensity"
```
erykoff, line 340: Okay this is where it gets very dicey.  What if we have comcam and latiss calibrations both going???

python/lsst/cp/pipe/utilsEfd.py
```diff
@@ -257,41 +315,163 @@ def parseMonochromatorStatus(self, data, dateStr):
...
 
-        # End binary search.
+        results = self.selectTimeSeries(dataSeries, [],
+                                        startDate, stopDate)
+        if dataSeries == 'lsst.sal.Electrometer.logevent_logMessage':
+            results = self.rewriteElectrometerStatus(results)
+        return results
+
+    def rewriteElectrometerStatus(self, inResults):
+        """Rewrite intermediate electrometer data from the EFD.
```
erykoff, line 363: Maybe specify that this is log data?

python/lsst/cp/pipe/utilsEfd.py
```diff
@@ -257,41 +315,163 @@ def parseMonochromatorStatus(self, data, dateStr):
...
+        Returns
+        -------
+        outResults : `astropy.table.Table`
+            The rewritten table containing only electrometer summary
+            status events.
+        """
+        # This is fragile against upstream changes.
+        # Ignore all entries that are not the ones we care about.
+        outResults = inResults[np.where(inResults['functionName'] == 'write_fits_file')]
+        outResults = outResults[np.where(outResults['level'] == 20)]
```
erykoff, line 379: The `np.where` is superfluous here.  You can just do `inResults[inResults["functionName"] == "write_fits_fit"]`, etc.

python/lsst/cp/pipe/utilsEfd.py
```diff
@@ -257,41 +315,163 @@ def parseMonochromatorStatus(self, data, dateStr):
...
+                filename, intensity_str, time_str = magic[0]
+                mean, median, stdev = intensity_str.split(',')
+                time_mean, time_median = time_str.split(',')
+                mean = float(mean)
+                median = float(median)
+                stdev = float(stdev)
+                time_mean = float(time_mean)
+                time_median = float(time_median)
+                # Censor the saturated points so plots look nice.
+                if mean > 1e37:
```
erykoff, line 410: Could this also be negative?

python/lsst/cp/pipe/utilsEfd.py
```diff
@@ -257,41 +315,163 @@ def parseMonochromatorStatus(self, data, dateStr):
...
+            mean = np.nan
+            stdev = np.nan
+            time_mean = np.nan
+            filename = "REGEXP_FAIL"
+
+            # Find the last "filename\.extension" before the first newline;
+            #      the last [] grouped values before the second newline;
+            #      the last [] grouped values before the end of the string.
+            magic = re.findall(r'\b\w+.+?(\w+?\.\w+)\n\b\w+.+\[(.*?)\]\n\b\w+.+\[(.*?)\]$',
+                               row['message'])
```
erykoff, line 398: oh wow this is ... not ideal but I understand why you did it.

tests/test_utilsEfd.py
```diff
@@ -0,0 +1,153 @@
...
+
+class UtilsEfdTestCase(lsst.utils.tests.TestCase):
+    """Unit test for EFD access code."""
+
+    def setUp(self):
+        # Initialize a client.
+        try:
+            self.client = CpEfdClient()
+        except Exception as e:
+            self.log.warning(f"Could not initialize EFD client: {e}")
```
erykoff, line 37: I don't think we warn on skipped tests.  I don't know if it's possible to initialize the client at the main level and use skip directives ... what does Merlin do for summit utils tests?

python/lsst/cp/pipe/utilsEfd.py
czwa: `time` is UTC, `private_sndStamp` should be TAI.  Except for the electrometers.  Which are UTC and so needed this scale addition.  I've chosen the path that was most likely to get things working, and we can revisit this if we can get all of the devices using a consistent time.

python/lsst/cp/pipe/utilsEfd.py
czwa: I don't think this can trigger outside of when I was debugging the search algorithm, but I've updated it to give information on what the state is when it failed.

python/lsst/cp/pipe/utilsEfd.py
czwa: I thought I'd already demoted it to debug.  Hopefully this time sticks.

python/lsst/cp/pipe/utilsEfd.py
czwa: They're distinguished by the `salIndex`.  I've updated the docstring in `parseElectrometerStatus` to list the values for the two telescopes.

python/lsst/cp/pipe/utilsEfd.py
czwa: Added absolute value.

python/lsst/cp/pipe/utilsEfd.py
czwa: Yeah.  I'm hoping we can deprecate this function as quickly as possible.

tests/test_utilsEfd.py
czwa: Looks like a raise that skips the test in setup: https://github.com/lsst-sitcom/summit_utils/blob/main/tests/test_efdUtils.py#L67  I've switched to do the same.

python/lsst/cp/pipe/utilsEfd.py
erykoff: The default and the comments don’t make sense to me, but these time issues are always hard. It says utc but above the default is tai…

python/lsst/cp/pipe/utilsEfd.py
```diff
@@ -257,41 +319,164 @@ def parseMonochromatorStatus(self, data, dateStr):
...
+            If true, take the average of all samples between
+            ``dateStr`` and ``dateEnd``.
+        index : `int`
+            The salIndex of the device we want to read.  For LATISS,
+            this should be 201.  For the main telescope, 101.
+
+        Returns
+        -------
+        indexDate : `str`
+            Date string indicating the monochromator state change.
```
erykoff, line 452: Is this utc or tai in the return?

python/lsst/cp/pipe/utilsEfd.py
czwa: tl;dr: Everything is in TAI now, and it appears to be consistent.

I hope I never end up in another UTC/TAI rabbit hole.  That said:

1. The filter scan task uses of the cp_pipe EFD client explicitly cast `visitInfo.date` to `DateTime.TAI` and uses that for the query.
2. The monochromator scan task uses the `DATE-BEG` header keyword, as the fiber spectrograph doesn't make a valid visitInfo yet, but that is defined as TAI in the headers.
3. The column `time` is a string representation of UTC, whether accessed via electrometer `intensity` or `logMessage` or via the monochromator `wavelength` series.  I've messed up the numbers here.
6. The column `private_sndStamp` is an float `unix_tai` value which is indeed in TAI.
7. `blah = Time(data['private_sndStamp'], format='unix_tai')` gives an `astropy.Time` version of the TAI value, which can be converted to a useful string representation with `blah.isot` and `blah.to_value('isot')`.
8. `blah.utc.isot` and `blah.utc.to_value('isot')` yield UTC string representations, as does `Time(data['private_sndStamp'], scale='utc', format='unix_tai').isot`.

Point 3 is forced, as the raw value received from the EFD is a string like `2022-10-05T14:47:49.764833Z`, which, due to the trailing `Z` must have `scale='utc'` or astropy complains.  Because that is what the EFD values are, we need to cast any start/end dates to UTC, which we do in `selectTimeSeries` when doing the query.  However, we rewrite the `time` column, but leave it in UTC.  I've now changed the code so that `time` is now TAI, making Point 3 no longer true.  Now the only method that has any conversions is `selectTimeSeries`, which now expects TAI start/end dates, will do the cast to UTC for the query, and converts the returned values in `time` to TAI, so `time` should now match `private_sndStamp`.  We've already had a `sort` on `time`, so this means the table is now monotonic, and universally in TAI.

If everything is in TAI, then `searchResults` no longer needs a `scale` argument, so I've removed that.  We actually do the search there against `private_sndStamp`, so this should have been in TAI all along.  Everything user-facing is now TAI.  Everything is TAI.

Running the tests, `test_electrometer` now fails, with `-1.5232e-07 != -1.8932e-07` where it's failing on the first lookup.  This shows that I have to ensure that the search string is converted with `scale='tai'`.  Which now causes issues because the test dates were poorly chosen.  Fixing that by ensuring the first date is included in the data set.

Now the `test_electrometer_alternate` is failing, but I can see by eye that's because the test dates are UTC, and not TAI as I've now asserted everything is in.  Fixing that results in all tests passing, so I'm going to commit and push.

python/lsst/cp/pipe/utilsEfd.py
czwa: As per previous comment: Everything is now in TAI.

python/lsst/cp/pipe/utilsEfd.py
```diff
@@ -157,9 +158,9 @@ def selectTimeSeries(self, topicName, fields=[], startDate=None, endDate=None):
             List of fields to return.  If empty, all fields are
             returned.
         startDate : `astropy.time.Time`, optional
-            Start date (in UTC) to limit the results returned.
+            Start date (in TAI) to limit the results returned.
```
erykoff, line 161: Ok for here I don’t think this should specify. An astropy time can do conversions between tai and utc as needed, right? If it’s astropy it’s not a problem. Only strings become problems.

python/lsst/cp/pipe/utilsEfd.py
```diff
@@ -248,50 +308,167 @@ def parseMonochromatorStatus(self, data, dateStr):
...
+        """Determine electrometer status for a specific date.
 
-        myTime = data["private_sndStamp"][idx]
-        return myTime.strftime("%Y-%m-%dT%H:%M:%S.%f"), data['wavelength'][idx]
+        Parameters
+        ----------
+        data : `astropy.table.Table`
+            The dataframe of electrometer results from the EFD.
+        dateStr : `str`
+            The date to look up in the status for.
```
erykoff, line 437: Specify tai here?

erykoff (approved):

python/lsst/cp/pipe/utilsEfd.py
timj: Right. Use startDate.tai

PR #258, opened 2024-08-07
DM-45465: Add EFD utility access to photodiode measurements part of cpPtcExtract

czwa:


aferte (approved):
This looks good to me. The definitions of obsMin and obsMax seem confusing to me but  it seems fine if it is needed.

czwa:
> This looks good to me. The definitions of obsMin and obsMax seem confusing to me but it seems fine if it is needed.

The EFD query needs to be time-restricted so that it doesn't blow up in memory (especially with the `logMessage` series, which contains far more information than we need).  I didn't want to use the strict start and end of the exposures because the timing between electrometer and camera isn't always going to line up, but I also didn't want to let the dates stretch far away from the observation we're looking for.  Using the length of the observation as the buffer seemed the least-bad option.

Since this still needs to be used to generate PTC/linearity curves, we'll see if this buffer is sufficient when we are processing that data.

PR #259, opened 2024-08-15
DM-45464: Fix handling of deprecated taskMetadata in cp_pipe

czwa:


python/lsst/cp/pipe/ptc/cpPtcExtract.py
```diff
@@ -62,7 +62,7 @@ class PhotonTransferCurveExtractConnections(pipeBase.PipelineTaskConnections,
         deferLoad=True,
     )
     taskMetadata = cT.Input(
-        name="isr_metadata",
+        name="cpPtcIsr_metadata",
```
erykoff, line 65: This won't work with the new pipelines, or with all the verification stuff, I think.  I _think_ what we can do is define the connection here so that it can be overridden by somebody with an old config, and then del the connection in the init below so it never actually gets used/connected by the graph generator.

python/lsst/cp/pipe/ptc/cpPtcExtract.py
```diff
@@ -62,7 +62,7 @@ class PhotonTransferCurveExtractConnections(pipeBase.PipelineTaskConnections,
         deferLoad=True,
     )
     taskMetadata = cT.Input(
```
erykoff, line 64: We need a TODO and a ticket to do the removal of this code that is blocked on the v28 release.

python/lsst/cp/pipe/ptc/cpPtcExtract.py
```diff
@@ -400,7 +400,7 @@ def run(self, inputExp, inputDims, taskMetadata, inputPhotodiodeData=None):
             sequentially by their exposure id.
         inputDims : `list`
             List of exposure IDs.
-        taskMetadata : `list` [`lsst.pipe.base.TaskMetadata`]
+        taskMetadata : `list` [`lsst.pipe.base.TaskMetadata`], optional
             List of exposures metadata from ISR.
```
erykoff, line 404: This should be commented that this is unused even if specified.  Also, below there should be a check if somebody hands it in and do a deprecation warning if so.  (And also tagged with the same TODO as above).

erykoff (changes requested):

python/lsst/cp/pipe/ptc/cpPtcExtract.py
```diff
@@ -81,7 +82,7 @@ class PhotonTransferCurveExtractConnections(pipeBase.PipelineTaskConnections,
     def __init__(self, *, config=None):
         if not config.doExtractPhotodiodeData:
             del self.inputPhotodiodeData
-
```
erykoff, line 84: This should have the TODO attached.

erykoff (approved):
There needs to be a corresponding PR to cp_verify.

PR #261, opened 2024-08-22
DM-45325: Move CTI simulation classes to lsst.ip.isr.deferredCharge

Alex-Broughton:


erykoff (commented):
This is a little bit tricky, but we need to deprecate the old imports in case somebody has been lifting `OverscanModel` from here.  See https://developer.lsst.io/stack/deprecating-interfaces.html in general, and see https://github.com/lsst/pipe_tasks/blob/c100eb463524ac2dc860a101d48d552f65cd4135/python/lsst/pipe/tasks/assembleCoadd.py#L32 for a possible pattern to lift the imports and warn if somebody imports from the old location.  You also need to file a ticket to remove the deprecated usage and add that as a TODO.

python/lsst/cp/pipe/__init__.py
```diff
@@ -35,3 +35,17 @@
...
+
+@deprecated(
+    reason="lsst.cp.pipe.OvercsanModel will be moved to lsst.ip.isr.deferredCharge.OverscanModel in v28.0.",
+    version="v28.0",
+    category=FutureWarning,
+)
+class OverscanModel(_OverscanModel):
+    pass
+
+del deprecated
```
erykoff, line 51: To be clear, we need to do this import/subclass/deprecate pattern for each of the moved classes.  We also need to file a ticket to clean up the deprecation, which should be a blocker for v29 release (if that exists as a ticket), and add a `TODO: Remove deprecated classes on DM-XXXXX`.  Also make sure there's a newline at the end of the file.

erykoff (changes requested):
The `__init__.py` needs a final newline.  This isn't triggering the linter because we generally skip `__init__.py` (because they have unused imports).

python/lsst/cp/pipe/__init__.py
```diff
@@ -35,3 +35,57 @@
...
+
+@deprecated(
+    reason="lsst.cp.pipe.FloatingOutputAmplifier will be moved to lsst.ip.isr.deferredCharge.FloatingOutputAmplifier in v28.0.",
+    version="v28.0",
+    category=FutureWarning,
+)
+class FloatingOutputAmplifier(_FloatingOutputAmplifier):
+    pass
+
+del deprecated
```
erykoff, line 91: Please make sure the file ends with a newline.

erykoff (approved):
Looks good pending the formatting problem on the `__init__`

erykoff (approved):

PR #262, opened 2024-09-03
DM-45976: Fix units bugs with camera read noise / ptc read noise / effective ptc / variance plane creation

Alex-Broughton:
This will introduce a new PTC version (1.9) which standardizes the PTC noise attribute units to electron. Previously, FULLCOVARIANCE fit mistakenly output in electron^2. Changing the PTC verison allows for new checks in ISR tasks to use the correct noise measurement. This previously caused the variance plane to be if PTC version < 1.9 and ptcFitType == FULLCOVARIANCE.


python/lsst/cp/pipe/ptc/cpPtcSolve.py
```diff
@@ -685,30 +686,35 @@ def fitDataFullCovariance(self, dataset):
...
-            readoutNoiseSqrt = np.sqrt(np.fabs(readoutNoise))
+            readoutNoiseSquared = fitResults['fullModel']['noiseMatrix'][0][0]
+            readoutNoise = np.sqrt(np.fabs(readoutNoiseSquared))
             dataset.noise[ampName] = readoutNoise
-            readoutNoiseSigma = fitResults['fullModel']['paramsErr'][2*lenParams]
-            dataset.noiseErr[ampName] = 0.5*(readoutNoiseSigma/np.fabs(readoutNoise))*readoutNoiseSqrt
-            dataset.noiseMatrix[ampName] = fitResults['fullModel']['noise']
-            dataset.noiseMatrixNoB[ampName] = fitResults['fullModelNoB']['noise']
+            readoutNoiseSquaredSigma = fitResults['fullModel']['paramsErr'][2*lenParams]
+            noiseErr = 0.5*(readoutNoiseSquaredSigma/np.fabs(readoutNoiseSquared))*readoutNoise
```
erykoff, line 714: If I'm reading this correctly, this is the same as before just with clearer variable names.  Which means this is `noiseSquaredErr` and not `noiseErr`.

erykoff (commented):
I worry about the lack of tests of these units.  And running `test_ptc.py` gives pages and pages of warnings (not new on this ticket!).  I think that most of them come from missing metadata keywords regarding the noise, which can be put in on this ticket.  Furthermore, we should confirm that the noise that is returned is what we think it should be for both EXPAPPROXIMATION and FULLCOVARIANCE.  This isn't currently done which is part of what led to this bug surviving for so long.  (Similarly for the `noiseErr` field).

tests/test_ptc.py
```diff
@@ -260,193 +255,25 @@ def test_covAstier(self):
...
-                    19.64121933,
-                    -36.09794843,
-                    -34.27205933,
-                    -25.16574105,
-                    23.80460972,
-                ],
-            ]
+            [[8.71049338, 12.48584043, -37.06585088],
+             [-4.80523971, -23.29102809, -66.37815343],
+             [7.48654766, -4.10168337, 35.64469824]],
```
erykoff, line 276: In the checks below I think we should have explicit tests of the `noise` and `noiseErr`.  The `noise` is implicitly handled by the no-warnings, but it's good to make it explicit.  Plus we need the error test for the full fit, not just the polynomial and expapproximation done later.

erykoff (commented):
One more missing test and then you're ready to go.

erykoff (approved):

PR #264, opened 2024-09-16
DM-46308: Updates to PTC fitting, including doAmpOffsetGainRatioFixup for new IsrTaskLSST

erykoff:
This PR implements a bunch of small updates found in the development of DM-45856, including the `doAmpOffsetGainRatioFixup` that uses flat-pair amp offsets to fix up noisy gain ratios determined from the individual amplifier PTC fitting.

### DIFF
--- a/python/lsst/cp/pipe/utils.py
+++ b/python/lsst/cp/pipe/utils.py
@@ -1387,6 +1387,9 @@ def ampOffsetGainRatioFixup(ptc, minAdu, maxAdu, log=None):
     """
     # We need to find the reference amplifier.
     # Find an amp near the middle to use as a pivot.
+    if log is None:
+        log = logging.getLogger(__name__)
+
     gainArray = np.zeros(len(ptc.ampNames))
     for i, ampName in enumerate(ptc.ampNames):
         gainArray[i] = ptc.gain[ampName]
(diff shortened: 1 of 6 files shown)


python/lsst/cp/pipe/utils.py
```diff
@@ -1367,3 +1373,97 @@ def getReadNoise(exposure, ampName, taskMetadata=None, log=None):
...
+    """
+    # We need to find the reference amplifier.
+    # Find an amp near the middle to use as a pivot.
+    if log is None:
+        log = logging.getLogger(__name__)
+
+    gainArray = np.zeros(len(ptc.ampNames))
+    for i, ampName in enumerate(ptc.ampNames):
+        gainArray[i] = ptc.gain[ampName]
+    good, = np.where(np.isfinite(gainArray))
```
Alex-Broughton, line 1402: Is this the best way to define the pivot amp? Pretty much every amp will have a finite gain, even if it's bad. Perhaps do some kind of gain outlier rejection in case there is a really weird amp?

erykoff: I've added more comments around here that this is just finding the "good" (not NaN) amps and below we choose the pivot as the one that is closest to the median, which is a simple case of outlier rejection.  I also updated the test to make one of the amps way out of family and it still does work.

Alex-Broughton (approved):

PR #265, opened 2024-09-16
DM-46189: Add missing cpDarkForDefects pipelines.

erykoff:


pipelines/LSSTCam/cpDarkForDefects.yaml
```diff
@@ -0,0 +1,11 @@
...
+# cpDark.yaml.
+description: LSSTCam dark calibration construction for defects.
+instrument: lsst.obs.lsst.LsstCam
+imports:
+  - location: $CP_PIPE_DIR/pipelines/_ingredients/cpDarkForDefects.yaml
+tasks:
+  cpDarkForDefectsIsr:
+    class: lsst.ip.isr.IsrTask
+    config:
+      doCrosstalk: false
```
jchiang87, line 11: `doLinearize: false` isn't needed?

jchiang87 (approved):
lgtm, though I do have one question about the `LSSTCam` config.

pipelines/LSSTCam/cpDarkForDefects.yaml
erykoff: Hmm.  It isn't there for the regular `LSSTCam/cpDark.yaml`.  I'm wondering what is going on in the pipeline right now...

PR #268, opened 2024-09-18
DM-45897: Add spectroFlat generation and application for LATISS

czwa:


python/lsst/cp/pipe/cpSpectroFlat.py
```diff
@@ -0,0 +1,324 @@
...
+        -------
+        results : `lsst.pipe.base.Struct`
+            The results struct containing:
+
+            ``outputData``
+                Final modified flat generated from the inputs
+                (`lsst.afw.image.Exposure`).
+        """
+        # Verify input
+        exposureFilter = dummyExpRef.get(component="filter")
```
aferte, line 194: is `dummyExpRef` really needed? It seems that `inputFlat` should be enough.

python/lsst/cp/pipe/cpSpectroFlat.py
```diff
@@ -0,0 +1,324 @@
...
+        # Todo: Get temperature dependence terms.
+        if self.config.scaleGainsByMean:
+            gainMean = np.mean(list(gains.values()))
+        else:
+            gainMean = 1.0
+
+        for amp in detector:
+            bbox = amp.getBBox()
+            gainFlat[bbox].image.array[:, :] = 1 / (gains[amp.getName()] / gainMean)
+            scaledGains[amp.getName()] = gains[amp.getName()] / gainMean
```
aferte, line 222: It would be nice to move this before `gainFlat` and use `scaledGains` in `gainFlat` directly.

python/lsst/cp/pipe/cpSpectroFlat.py
```diff
@@ -0,0 +1,324 @@
...
+
+        # Todo: Get temperature dependence terms.
+        if self.config.scaleGainsByMean:
+            gainMean = np.mean(list(gains.values()))
+        else:
+            gainMean = 1.0
+
+        for amp in detector:
+            bbox = amp.getBBox()
+            gainFlat[bbox].image.array[:, :] = 1 / (gains[amp.getName()] / gainMean)
```
aferte, line 221: Isn't gainFlat = scaledGains rather than 1/scaledGains? It is just an exposure with the gain so it should just be the gain rescaled.

python/lsst/cp/pipe/cpSpectroFlat.py
```diff
@@ -0,0 +1,324 @@
...
+    )
+    smoothingOutlierPercentile = pexConfig.Field(
+        dtype=float,
+        default=1.0,
+        doc="Percentile of extreme values to clip from top and bottom of distribution.",
+    )
+
+    # makePixelFlat
+    # photutils.background Background2D, MedianBackground
+    # astropy.stats.SigmaClip
```
aferte, line 121: Are the above 2 comments needed?

aferte (changes requested):
I have a few comments to address but otherwise looks good.

python/lsst/cp/pipe/cpSpectroFlat.py
czwa: No.  Deleted.

python/lsst/cp/pipe/cpSpectroFlat.py
czwa: The exposure is needed to add a time constraint on the butler lookup for the flat and PTC.  Without it, there's nothing constraining which calibrations in the input calibration collections to select.

python/lsst/cp/pipe/cpSpectroFlat.py
czwa: Not based on the example code/notebook supplied.  The original method ( https://github.com/lsst/atmospec/blob/u/jneveu/doFlat/python/lsst/atmospec/spectroFlat.py#L120 ) has an `invertGains` option that allows the gains to be in the denominator (if `True`), or in the numerator (if `False`).  The defaults have this set to `False`, but the example notebook has this flipped.  

I see now that I added the config option `applyGainDirection`, but never used it.  I've fixed that, correcting the docstring to indicate which direction `True` denotes, and wrapped this operation in an if/else block.

I think this complicates the next ticket, that will update this for the task change, but I haven't thought it completely through yet.  I did add some more details on the JIRA ticket, to preserve my thinking from today.

aferte (approved):
Looks good to me, thanks!

PR #269, opened 2024-09-25
DM-42863: Add method to evaluate PTC model in ptcDataset

Alex-Broughton:


tests/test_ptc.py
```diff
@@ -414,6 +414,15 @@ def notest_covAstier(self):
...
                 )
+            # Check if evalPtcModel produces expected values
+            nanMask = ~np.isnan(ptc.finalMeans[ampName])
+            means = ptc.finalMeans[ampName][nanMask]
+            covModel = ptc.covariancesModel[ampName][nanMask]
+            covariancesModel = ptc.evalPtcModel(means, setBtoZero=False)[ampName]
+            self.assertFloatsAlmostEqual(covariancesModel, covModel, atol=1e-13)
+            # Note that the PhotoTransferCurveDataset does not store the gain
+            # fit parameter for FULLCOVARIANCE with b=0, so we can't compare
+            # exactly.
```
erykoff, line 425: Is that something that we need to add?

erykoff (approved):

tests/test_ptc.py
Alex-Broughton: Yes we do, but I want to keep this ticket small and then filed this ticket to update the ptc class https://rubinobs.atlassian.net/browse/DM-46509

PR #272, opened 2024-09-27
DM-46546: Linearity fit residual test should use linear fit parameters as initial guess

erykoff:
This makes sure that when using the photodiode (much larger scaling) that things converge correctly when checking residuals.


czwa (approved):
This makes sense to me.

PR #273, opened 2024-09-30
DM-46426: Create new IsrTaskLSST pipeline yaml for CTI

Alex-Broughton:


czwa (approved):

pipelines/_ingredients/cpCtiLSST.yaml
```diff
@@ -0,0 +1,28 @@
...
+      connections.ccdExposure: "raw"
+      connections.outputExposure: "cpCtiIsrExp"
+      connections.outputStatistics: "cpCtiStatistics"
+      python: |
+        from lsst.cp.pipe import configureIsrTaskLSSTForCalibrations
+
+        configureIsrTaskLSSTForCalibrations(config)
+
+        config.doCalculateStatistics = True
+
```
czwa, line 15: Don't you also need a `config.isrStatistics.doCtiStatistics = True`?

czwa (approved):
I'm generally happy here, but I'm concerned that this is missing a config options which suggests this hasn't been tested (at least to check that it runs completely).  I see this is already in the ticket description, so I'll raise this there.

PR #274, opened 2024-10-02
DM-46357: Add calibration pipelines for new IsrTaskLSST for LSSTComCam and LSSTComCamSim

erykoff:


pipelines/LSSTComCam/cpDefects.yaml
```diff
@@ -1,4 +1,29 @@
...
+description: cp_pipe LSSTComCam defect calibration construction.
 instrument: lsst.obs.lsst.LsstComCam
 imports:
-  - location: $CP_PIPE_DIR/pipelines/_ingredients/cpDefects.yaml
+  - location: $CP_PIPE_DIR/pipelines/_ingredients/cpDefectsLSST.yaml
+tasks:
+  cpMeasureBiasDefects:
+    class: lsst.cp.pipe.MeasureDefectsCombinedTask
+    config:
+      badPixelsToFillColumnThreshold: 1000
```
czwa, line 9: Do we expect these config options (for all three tasks) to differ between the three cameras?

pipelines/LSSTComCamSim/cpDefects.yaml
```diff
@@ -1,4 +1,29 @@
-description: cp_pipe DEFECT calibration construction from combined images.
+description: cp_pipe LSSTComCamSim defect calibration construction.
 instrument: lsst.obs.lsst.LsstComCamSim
 imports:
-  - location: $CP_PIPE_DIR/pipelines/_ingredients/cpDefects.yaml
+  - location: $CP_PIPE_DIR/pipelines/_ingredients/cpDefectsLSST.yaml
```
czwa, line 4: Have we ever run this on LSSTComCamSim data?

czwa (approved):
I think these are just questions for my own understanding.

pipelines/LSSTComCam/cpDefects.yaml
erykoff: I have left them the same for now, because of cargo culting.  But I could see these being different for LSSTCam based on defect research.

pipelines/LSSTComCamSim/cpDefects.yaml
erykoff: No, actually.  I was asked to leave it in just in case ComCamSim adds defects in the future.  But I actually think that it is more likely we will just use the existing defects and apply them and not "find" them so this may not be necessary.  But wanted to leave the option open.

PR #277, opened 2024-10-19
DM-46358: Add new IsrTaskLSST calibration pipelines for LSSTCam.

erykoff:
There is some code change on this ticket as well to better handle outliers on the initial linearity turnoff fit that came up with full camera testing.


pipelines/LSSTCam/cpFlatSingleChip.yaml
```diff
@@ -1,4 +1,4 @@
-description: cp_pipe FLAT calibration construction
+description: cp_pipe LSSTCam FLAT calibration construction, optimized for a single chip.
```
czwa, line 1: This file can probably be dropped, as we're unlikely to need it for the full camera.

pipelines/LSSTCam/legacy_202409/cpLinearizer.yaml
```diff
@@ -0,0 +1,20 @@
...
+imports:
+  - location: $CP_PIPE_DIR/pipelines/_ingredients/cpLinearizer.yaml
+tasks:
+  cpLinearizerSolve:
+    class: lsst.cp.pipe.LinearitySolveTask
+    config:
+      linearityType: Spline
+      usePhotodiode: true
+      splineGroupingColumn: CCOBCURR
+      maxLinearAdu: 120000
```
czwa, line 12: Is this parameter no longer needed?  It seems to be gone in the new pipelines, and I just wanted to check that was intentional.  Apologies if I didn't follow the inheritance correctly and just missed it.  I think the rest of the configs are consistent.

czwa (approved):
Minor comments and questions only.

pipelines/LSSTCam/cpFlatSingleChip.yaml
erykoff: I assumed we wanted these in case somebody wanted to run a test flat for a single chip.  If that's wrong I can remove them all.

PR #280, opened 2024-11-06
DM-45928: add parallel CTI and add serial/parallel turnoff calculations to DeferredChargeCalib

Alex-Broughton:


pipelines/LSSTComCamSim/cpCti.yaml
```diff
@@ -0,0 +1,4 @@
+description: LSSTCam cti calibration construction
+instrument: lsst.obs.lsst.LsstComCamSim
+imports:
+  - location: $CP_PIPE_DIR/pipelines/_ingredients/cpCtiLSST.yaml
```
erykoff, line 4: This is fine such as it is, but note that LSSTComCamSim doesn't have any CTI to my knowledge.

python/lsst/cp/pipe/cpCtiSolve.py
```diff
@@ -89,15 +90,42 @@ class CpCtiSolveConfig(pipeBase.PipelineTaskConfig,
         default=10000.0,
         doc="Upper flux limit to use for CTI fit (electron).",
     )
+    serialCtiRange = pexConfig.ListField(
+        dtype=float,
+        default=[-1.0e-5, 1.0e-5],
+        doc="Serial CTI range within containing serial turnoff.",
```
erykoff, line 96: This docstring does not parse.  Can you reword?

python/lsst/cp/pipe/cpCtiSolve.py
```diff
@@ -89,15 +90,42 @@ class CpCtiSolveConfig(pipeBase.PipelineTaskConfig,
...
+    )
+    parallelCtiRange = pexConfig.ListField(
+        dtype=float,
+        default=[-1.0e-5, 1.0e-5],
+        doc="Parallel CTI range within containing serial turnoff.",
```
erykoff, line 101: Same here, I don't understand this sentence.

python/lsst/cp/pipe/cpCtiSolve.py
```diff
@@ -89,15 +90,42 @@ class CpCtiSolveConfig(pipeBase.PipelineTaskConfig,
...
+    turnoffFinderSigmaClipMaxIters = pexConfig.Field(
+        dtype=int,
+        default=5,
+        doc="Maximum iterations for sigma clipping in turnoff finder.",
+    )
     globalCtiColumnRange = pexConfig.ListField(
+        dtype=int,
+        default=[1, 15],
+        doc="First and last serial overscan column to use for "
+            "serial EPER estimate.",
```
erykoff, line 117: Can you spell out EPER here and below?  (I think Extended Pixel Edge Response?)

python/lsst/cp/pipe/cpCtiSolve.py
```diff
@@ -437,6 +479,100 @@ def solveGlobalCti(self, inputMeasurements, calib, detector):
 
         return calib
 
+    def solveEper(self, inputMeasurements, calib, detector):
+        """Solve for serial and parallel EPER (estimator of CTI).
```
erykoff, line 483: Can you spell out EPER here as well?

python/lsst/cp/pipe/cpCtiSolve.py
```diff
@@ -437,6 +479,100 @@ def solveGlobalCti(self, inputMeasurements, calib, detector):
...
+
+        Parameters
+        ----------
+        inputMeasurements : `list` [`dict`]
+            List of overscan measurements from each input exposure.
+            Each dictionary is nested within a top level 'CTI' key,
```
erykoff, line 489: Double quote here?

python/lsst/cp/pipe/cpCtiSolve.py
```diff
@@ -437,6 +479,100 @@ def solveGlobalCti(self, inputMeasurements, calib, detector):
...
+
+            # Do parallel EPER calculation
+            signals, parallelEperEstimate = self.calcEper(
+                "PARALLEL",
+                inputMeasurements,
+                amp,
+            )
+
+            # Calculate the serial and parallel turnoffs
+            serialCtiTurnoff, serialCtiTurnoffSamplingErr = self.calcTurnoff(
```
erykoff, line 546: It's a big more verbose, but can you rename this function `calcCtiTurnoff()`?

python/lsst/cp/pipe/cpCtiSolve.py
```diff
@@ -611,3 +747,207 @@ def findTraps(self, inputMeasurements, calib, detector):
...
+
+    def calcEper(self, mode, inputMeasurements, amp):
+        """Solve for serial or parallel global CTI using the extended
+        pixel edge response (EPER) method.
+
+        Parameters
+        ----------
+        mode : `str`
+            The orientation of the calculation to perform. Can be
+            either `SERIAL` or `PARALLEL`.
```
erykoff, line 759: These (and CTI) should be double quotes, not backticks

python/lsst/cp/pipe/cpCtiSolve.py
```diff
@@ -611,3 +747,207 @@ def findTraps(self, inputMeasurements, calib, detector):
...
+
+        Returns
+        -------
+        calib : `lsst.ip.isr.DeferredChargeCalib`
+            Populated calibration.
+
+        Raises
+        ------
+        RuntimeError : Raised if no data remains after flux filtering or if
+            the mode string is not one of "SERIAL" or "PARALLEL".
```
erykoff, line 798: I understand raising if it's not SERIAL or PARALLEL.  But what happens if you have a bad amp?  Will this crash on LSSTCam detectors with bad amps?  (That would be bad; we want it to log and set bad amps to NaN or whatever but not Raise).

python/lsst/cp/pipe/cpCtiSolve.py
```diff
@@ -611,3 +747,207 @@ def findTraps(self, inputMeasurements, calib, detector):
...
+        # the first few columns or rows of the overscan.
+        overscan1 = data[:, 0]
+        overscan2 = data[:, 1]
+        ctiEstimate = (overscan1 + overscan2)/(nShifts*np.array(signal))
+
+        return signal, ctiEstimate
+
+    def calcTurnoff(self, signalVec, dataVec, ctiRange, amp):
+        """Solve for turnoff value in a sequenced dataset.
+
```
erykoff, line 853: There's an extra blank line here.

python/lsst/cp/pipe/cpCtiSolve.py
```diff
@@ -611,3 +747,207 @@ def findTraps(self, inputMeasurements, calib, detector):
...
+        )
+
+        # Retrieve the result
+        good = ~np.ma.getmask(cleanDataVecMaArray)
+        cleanDataVec = np.ma.getdata(cleanDataVecMaArray)
+
+        if cleanDataVec.size == 0:
+            self.log.warning("No data points after sigma clipping to compute turnoff "
+                             f" for amplifier {amp.getName()}. Setting turnoff point "
+                             "to 0 el.")
```
erykoff, line 929: el -> electron?

python/lsst/cp/pipe/cpCtiSolve.py
```diff
@@ -611,3 +747,207 @@ def findTraps(self, inputMeasurements, calib, detector):
...
+        turnoff = np.max(signalVec[good])
+
+        if cleanDataVec[good][-1] in ctiRange or turnoffIdx in [0, len(signalVec)-1]:
+            self.log.warning("Turnoff point is at the edge of the allowed range for "
+                             f"amplifier {amp.getName()}.")
+
+        self.log.info(f"Amp {amp.getName()}: There are {len(cleanDataVec[good])}/{len(dataVec)} data points "
+                      f"left to determine turnoff point.")
+
+        # Compute the sampliing error as one half the
```
erykoff, line 942: sampliiiiing

tests/test_deferredCharge.py
```diff
@@ -37,153 +38,231 @@ class CpCtiSolveTaskTestCase(lsst.utils.tests.TestCase):
     def setUp(self):
         self.camera = IsrMock().getCamera()
```
erykoff, line 40: Where did these data come from?

tests/test_deferredCharge.py
```diff
@@ -209,6 +293,16 @@ def test_task(self):
...
+            self.assertEqual(calib.signals[ampName][-2], calib.serialCtiTurnoff[ampName])
+            self.assertEqual(calib.signals[ampName][-1], calib.parallelCtiTurnoff[ampName])
+            self.assertEqual(
+                (calib.signals[ampName][-1] - calib.signals[ampName][-3]) / 2.0,
+                calib.serialCtiTurnoffSamplingErr[ampName],
+            )
+            self.assertEqual(
+                calib.signals[ampName][-1] - calib.signals[ampName][-2],
+                calib.parallelCtiTurnoffSamplingErr[ampName],
+            )
```
erykoff, line 305: Where did the expectations come from above?

python/lsst/cp/pipe/cpCtiSolve.py
czwa: I was about to leave a comment asking these to be fully spelled out as well.

tests/test_deferredCharge.py
czwa: I think the originals were just empirical values I had during testing, and suspect the same of the updated values.

tests/test_deferredCharge.py
Alex-Broughton: I added `SerialOscanMeansD`, which is roughly `20*SerialOscanMeansC`. The code should have correctly calculated that the serial turnoff is at the signal level associated with C and the serial overscan turnoff sampling error is roughly (signal D - signal B) / 2. 

I also set all the parallel overscan means to the same thing at all signal levels (which is similar to never sampling the turnoff, which is common in real data). In that case it shoul set the parallel turnoff to the last value and the parallel turnoff sampling error should be the spacing between the last value and the second-to-last value.

I can add a comment to clarify this.

python/lsst/cp/pipe/cpCtiSolve.py
erykoff: I'm still worried about this; I don't actually see where the flux filtering is happening, but we don't want this code raising an exception when given data with a bad amplifier.

Alex-Broughton: This function (calcEper) is only ever called from CpCtiSolveTask, where mode is only ever {"SERIAL", "PARALLEL"}. If an amp is labeled "BAD", it's imaging region is set to the mean, and it's overscan is unaffected, correct? At least that what it appears. 

So if a bad amplifier gets statistics generated and then has cpCtiSolve run on it, it will produce a solution, but it just won't be very good.

erykoff (approved):

czwa (approved):

PR #285, opened 2024-11-19
DM-47613: Investigate nans in combined dark variance planes

czwa:


python/lsst/cp/pipe/cpDark.py
```diff
@@ -114,19 +124,38 @@ def run(self, inputExp):
         # exposure, if possible.  Otherwise, use the cameraGeom value.
         gains = self._get_gains(inputExp)
 
+        # Is this gainContext load bearing?
```
erykoff, line 127: I don't understand this comment.

python/lsst/cp/pipe/cpDark.py
```diff
@@ -114,19 +124,38 @@ def run(self, inputExp):
...
         with gainContext(inputExp, inputExp.getVariance(), apply=True, gains=gains):
             # Scale the variance to match the image plane.  A similar
             # scaling happens during flat-field correction for science
             # images.
             self.log.debug("Median image and variance: %f %f",
                            np.median(inputExp.image.array), np.median(inputExp.variance.array))
+            crImage = inputExp.clone()
+            # Interpolate the crImage, so the CR code can ignore
+            # defects (which will now be interpolated).
```
erykoff, line 136: Are you putting this here so that we can use the new interpolation code if we turn it on?  Otherwise interpolation will be done by the CR task below if you supply the `defects`

python/lsst/cp/pipe/cpDark.py
```diff
@@ -114,19 +124,38 @@ def run(self, inputExp):
...
                 self.log.warning("CR rejection failed!")
+
+            # Copy results to input frame.
+            crBit = crImage.mask.getPlaneBitMask('CR')
+            crPixels = np.bitwise_and(crImage.mask.array, crBit)
+            inputExp.mask.array[crPixels] |= crBit
             self.log.info("Number of CR pixels: %d",
-                          np.count_nonzero(np.bitwise_and(inputExp.mask.array,
-                                                          inputExp.mask.getPlaneBitMask('CR'))))
+                          np.count_nonzero(np.bitwise_and(crImage.mask.array, crBit)))
```
erykoff, line 158: Is this just not `np.count_nonzero(crPixels)`?

python/lsst/cp/pipe/cpDark.py
czwa: Cosmic ray rejection was highly dependent on this `gainContext` with the old ISR task, and without this scaling, we'd end up with too many CRs.  I've filed a ticket to see if the improvements in the new task make this `gainContext` unnecessary or not.  DM-48754 (with comment updated to point there).

python/lsst/cp/pipe/cpDark.py
czwa: I think I confused myself over the course of this ticket's life and general confusion last week.  I've dropped the passing of constructed defects to the CR code, as this interpolation is more flexible (we can switch to the new interpolation if desired).  I also see that the code from last week wasn't using the `maskListToInterpolate` config option, so this version appears more robust.

python/lsst/cp/pipe/cpDark.py
```diff
@@ -114,19 +124,36 @@ def run(self, inputExp):
...
+            )
+
             try:
-                self.repair.run(inputExp, keepCRs=False)
+                self.repair.run(crImage, keepCRs=False)
             except LengthError:
                 self.log.warning("CR rejection failed!")
+
+            # Copy results to input frame.
+            crBit = crImage.mask.getPlaneBitMask('CR')
```
erykoff, line 152: This file uses double-quotes not single (for the most part...).

python/lsst/cp/pipe/cpDark.py
erykoff: Good point, the other less flexible method would only do BAD.  This is configurable, so better.

python/lsst/cp/pipe/cpDark.py
```diff
@@ -73,6 +73,16 @@ class CpDarkTaskConfig(pipeBase.PipelineTaskConfig,
         target=RepairTask,
         doc="Repair task to use.",
     )
+    maskListToInterpolate = pexConfig.ListField(
+        dtype=str,
+        doc="List of mask planes that should be interpolated.",
+        default=['SAT', 'BAD'],
```
erykoff, line 79: I think the SAT is fine, but in a dark anything that is SAT better be marked BAD!

erykoff (approved):

python/lsst/cp/pipe/cpDark.py
czwa: I've switched to double quotes.  They have slightly different meanings in perl, and I can't unlearn ~three decades of habit.

PR #286, opened 2024-11-25
DM-47365: new manual defects for LSSTComCam

aferte:


python/lsst/cp/pipe/cpDefects.py
```diff
@@ -1189,6 +1208,7 @@ def runQuantum(self, butlerQC, inputRefs, outputRefs):
                     self.chooseBest(inputs['inputDarkDefects']),
                     self.chooseBest(inputs['inputBiasDefects'])]
 
+        tempList.extend(inputs["inputManualDefects"])
```
czwa, line 1211: I think this needs a check to see if this exists.  If `doManualDefects=False`, this connection will get dropped, so there will be no entry in `inputs` with this key.  I don't think this needs a check on `doManualDefects` (although that wouldn't hurt), but checking `if "inputManualDefects" in inputs.keys():` should ensure that this won't raise.

czwa (approved):
One minor logic issue that I think might cause issues if we don't fix it now.

PR #287, opened 2024-12-04
DM-47636: Switch CROSSTALK masking to use subtrahend masking

czwa:
Only exclude mask planes we've asked to exclude.


erykoff (approved):
Ah I didn't realize we were setting the andMask and using it elsewhere and not here.  🤦

PR #288, opened 2024-12-18
DM-48197: Add additional efd connection checks to ensure tests only run where they can

erykoff:


python/lsst/cp/pipe/utilsEfd.py
```diff
@@ -89,6 +91,23 @@ def _getAuth(self, instanceAlias):
         else:
             raise RuntimeError(f"Could not connect to {url}")
 
+    def checkConnection(self):
+        """Check the connection to the EFD.
+
+        Raises
+        ------
+        RuntimError :
```
taranu, line 99: RuntimeError

tests/test_utilsEfd.py
```diff
@@ -29,12 +29,13 @@
...
     """Unit test for EFD access code."""
 
-    def setUp(self):
+    @classmethod
+    def setUpClass(cls):
         # Initialize a client.
         try:
-            self.client = CpEfdClient()
+            cls.client = CpEfdClient()
         except Exception as e:
```
taranu, line 37: Is this still supposed to give up on any Exception? It appears as if it should run if you can connect to the EFD, no?

erykoff: If it can't connect to the EFD with any error it should skip the tests.  I'm not sure what your question is here.

mfisherlevine: My question would be: why do this at all, and not just use `vcr` to test the code exactly the same way, but with no dependency on the EFD at all? That way the tests always do run, and you don't have to ever worry about connections.

taranu: All I meant was that you are raising a `RuntimeError` in `checkConnection` but here you catch any exception from `CpEfdClient.__init__`. But I suppose that's fine since it is reraising and not just continuing.

erykoff: Would you prefer that this just check for a RuntimeError?

taranu: Eh, never mind. Maybe reconsider if/when you switch to vcr and maybe it'll just be moot.

PR #290, opened 2025-01-17
DM-48375: cpCombine should optionally erase all mask planes except NO_DATA

czwa:


python/lsst/cp/pipe/cpCombine.py
```diff
@@ -349,13 +363,26 @@ def run(self, inputExpHandles, inputScales=None, inputDims=None):
...
+            # CROSSTALK plane from printing into the calibration.  Run
+            # after the checkNoData pass so that we don't censor what
+            # we want to raise on.
+            mask = combinedExp.mask
+            for plane in mask.getMaskPlaneDict().keys():
+                if plane != self.config.noGoodPixelsMask:
+                    mask.clearMaskPlane(plane)
 
-        self.interpolateNans(combined)
+        self.interpolateNans(combined, self.noGoodPixelsMask)
```
erykoff, line 385: Can you explicitly use the keyword here, `maskPlane=self.noGoodPixelsMask`?

python/lsst/cp/pipe/cpCombine.py
```diff
@@ -653,13 +680,23 @@ def interpolateNans(self, exp):
...
-        bad = np.isnan(array)
+        array = exp.image.array
+        mask = exp.mask.array
+        variance = exp.variance.array
+        badMaskValue = exp.mask.getPlaneBitMask(maskPlane)
+
+        bad = np.isnan(array) | np.isnan(variance)
         if np.any(bad):
             median = np.median(array[np.logical_not(bad)])
+            medianVariance = np.median(variance[np.logical_not(bad)])
```
erykoff, line 694: This can be `variance[~bad]`.

python/lsst/cp/pipe/cpCombine.py
```diff
@@ -701,6 +738,9 @@ def calibStats(self, exp, calibrationType):
             bad = ((ampImage.mask.array & noGoodPixelsBit) > 0)
             key = f"LSST CALIB {calibrationType.upper()} {amp.getName()} BADPIX-NUM"
             metadata[key] = bad.sum()
+            self.log.warn(f"Found {metadata[key]} pixels with "
+                          f"mask plane {self.config.noGoodPixelsMask} "
+                          f"for amp {amp.getName()}.")
```
erykoff, line 743: Two things.  First, this must be `self.log.warning()` and not `self.log.warn()`.  Second, this should only trigger if `metadata[key] > 0` or else we'll get flooded with non-warnings.

python/lsst/cp/pipe/cpCombine.py
czwa: Yes, and I've fixed the missing `.config`.

erykoff:
The code as it is looks good.  I do wonder about tests, however.

PR #291, opened 2025-01-22
DM-48468: Turn off gaussian histogram Ks testing by default.

erykoff:


python/lsst/cp/pipe/ptc/cpPtcExtract.py
```diff
@@ -201,6 +201,12 @@ class PhotonTransferCurveExtractConfig(pipeBase.PipelineTaskConfig,
             'FULL': 'Second order correction.'
         }
     )
+    doKsHistMeasurement = pexConfig.Field(
+        dtype=bool,
+        doc="Do the Ks test of gaussianity? If False, will be filled with all 0.0; "
```
jchiang87, line 206: It would be good to state explicitly that it's the ks test pvalues that will be filled with zeros.   I'm wondering though if it would make more sense to fill them with 1.0 so that all of the points will be accepted no matter what pvalue threshold is used.

jchiang87 (approved):
I think this is fine as-is, aside from updating the doc for `doKsHistMeasurement`, but I'd also consider setting the p-values to 1.0 instead if that config is false.

PR #292, opened 2025-01-22
DM-46562: Add bps templates for LSSTCam/LSSTComCam/LATISS calibration creation.

erykoff:


bps/templates/bps_flat.yaml
```diff
@@ -0,0 +1,11 @@
...
+
+project: ${TICKET}
+campaign: ${TICKET}
+
+payload:
+  payloadName: ${INSTRUMENT}_${TICKET}
+  output: ${USER_CALIB_PREFIX}${INSTRUMENT}/calib/${TICKET}/${TAG}/flatGen-${FLAT_BAND}.${RERUN}
+  butlerConfig: ${REPO}
+  inCollection: ${USER_CALIB_PREFIX}${INSTRUMENT}/calib/${TICKET}/${TAG},${RAW_COLLECTION},${CALIB_COLLECTIONS}
+  dataQuery: ${!SELECTION_FLAT_BAND}
```
czwa, line 11: I think this exclamation point is a typo.

bps/templates/README.md
```diff
@@ -0,0 +1,61 @@
...
+
+A Note on Flat Generation
+-------------------------
+
+In order to use the template to generate flats, you must have the `$FLAT_BAND` and `$SELECTION_FLAT_BAND` env vars set, as well as (e.g.) `$SELECTION_FLAT_g`.  For example:
+
+```
+export FLAT_BAND=g
+export SELECTION_FLAT_BAND=SELECTION_FLAT_g
+export SELECTION_FLAT_g="instrument='LSSTComCam' and selection-string"
```
czwa, line 57: This uses `selection-string` instead of `selection_string` used elsewhere.  Not really a "bug" as this is documentation.

bps/templates/bps_flat.yaml
erykoff: This is bash "indirection" which should allow one template to be used for all the flats with different selection variables.  That's the idea, at least.  I haven't gotten to testing this part yet!

bps/templates/README.md
czwa: Actual comment:  If we have a band variable, we should constrain the query to just that band.  This will prevent future accidents where we generate a 15 input g flat and a 1 input r flat in the same collection because an endpoint was misread.

erykoff: True but not trivial because we need to know the physical filter in the constraint, so I can't easily template that generally for LATISS/LSSTComCam/LSSTCam.  But we can put strong recommendation that `SELECTION_FLAT_g` and friends include a physical filter constraint!

bps/templates/bps_linearizer.yaml
```diff
@@ -0,0 +1,11 @@
...
+
+project: ${TICKET}
+campaign: ${TICKET}
+
+payload:
+  payloadName: ${INSTRUMENT}_${TICKET}
+  output: ${USER_CALIB_PREFIX}${INSTRUMENT}/calib/${TICKET}/${TAG}/linearizerGen.${RERUN}
+  butlerConfig: ${REPO}
+  inCollection: ${USER_CALIB_PREFIX}${INSTRUMENT}/calib/${TICKET}/${TAG},${RAW_COLLECTION},${CALIB_COLLECTIONS}
+  dataQuery: ${SELECTION_PTC_LINEARIZER-$SELECTION_PTC}
```
czwa, line 11: This has a typo as well.

bps/templates/README.md
```diff
@@ -0,0 +1,61 @@
...
+the env vars must be explicitly exported.
+
+The calibrations also assume that the caller is continuously certifying
+calibrations for the next step into a collection
+`${USER_CALIB_PREFIX}${INSTRUMENT}/calib/${TICKET}/${TAG}`.
+
+Environment Variables
+---------------------
+
+These are the environment variables that must be used. Examples are substituted in.
```
czwa, line 26: This should also note that `selection_string` is a dummy variable here, and that it needs to be filled with something explicit.  I was going to suggest putting in the information being used on DM-46562 (the ticket used below), but I see that's this ticket adding the BPS templates, not a generation ticket.

bps/templates/bps_linearizer.yaml
erykoff: More bash magic that is supposed to use `$SELECTION_PTC_LINEARIZER` if it is set and otherwise use `$SELECTION_PTC`.  Also not yet tested if the bps vars support this...

bps/templates/README.md
czwa: I don't think we need to know the `physical_filter`.  I assume `pipetask` and `butler query-dimension-records` use the same resolver, so something like this should work:
```
butler query-dimension-records /repo/main exposure --where "instrument='LSSTComCam' AND physical_filter.band='g'" --limit 1
instrument       id      day_obs         group         physical_filter        obs_id        exposure_time    dark_time    observation_type observation_reason seq_num seq_start seq_end target_name science_program tracking_ra tracking_dec sky_angle azimuth zenith_angle has_simulated can_see_sky               timespan (TAI)              
---------- ------------- -------- -------------------- --------------- -------------------- ------------- --------------- ---------------- ------------------ ------- --------- ------- ----------- --------------- ----------- ------------ --------- ------- ------------ ------------- ----------- ------------------------------------------
LSSTComCam 2021040100003 20210401 CALSET_20210401_1601            g_07 CC_O_20210401_000003           1.0 1.5343017578125             flat               flat       3         3       3     UNKNOWN         unknown        None         None      None    None         None         False        None [2021-04-01T16:24:36, 2021-04-01T16:24:37)
```

erykoff: When getting a flat from a calib repo that doesn't work, so I guess I had assumed we needed it in the constraint as well.

bps/templates/bps_biasBootstrap.yaml
```diff
@@ -0,0 +1,15 @@
+pipelineYaml: "${CP_PIPE_DIR}/pipelines/${INSTRUMENT}/cpBiasBootstrap.yaml"
+
+project: "${TICKET}"
+campaign: "${TICKET}"
+
+biasBootstrapCollection: "${USER_CALIB_PREFIX}$INSTRUMENT/calib/$TICKET/$TAG/biasBootstrapGen.$RERUN"
+biasBootstrapRun: "{biasBootstrapCollection}/run${BOOTSTRAP_RUN_NUMBER}"
```
czwa, line 7: Is this missing a `$` for `biasBootstrapCollection`?

bps/templates/bps_darkBootstrap.yaml
```diff
@@ -0,0 +1,16 @@
+pipelineYaml: "${CP_PIPE_DIR}/pipelines/${INSTRUMENT}/cpDarkBootstrap.yaml"
+
+project: "${TICKET}"
+campaign: "${TICKET}"
+
+darkBootstrapCollection: "${USER_CALIB_PREFIX}$INSTRUMENT/calib/$TICKET/$TAG/darkBootstrapGen.$RERUN"
+biasBootstrapRun: "${USER_CALIB_PREFIX}$INSTRUMENT/calib/$TICKET/$TAG/biasBootstrapGen.$RERUN/run${BOOTSTRAP_RUN_NUMBER}"
+darkBootstrapRun: "{darkBootstrapCollection}/run${BOOTSTRAP_RUN_NUMBER}"
```
czwa, line 8: Same question as on biasBootstrap re `$`.

czwa: I see `inCollection` is the same, so maybe I'm just confused about variable scope.  Those defined within the yaml only need braces, but those that are inherited from the environment need `${}`?

bps/templates/bps_flat_g.yaml
```diff
@@ -0,0 +1,11 @@
+pipelineYaml: "${CP_PIPE_DIR}/pipelines/${INSTRUMENT}/cpFlat.yaml"
+
+project: "${TICKET}"
+campaign: "${TICKET}"
+
+payload:
+  payloadName: "${INSTRUMENT}_${TICKET}"
```
czwa, line 7: Does the `payloadName` need to be unique?  Will it complain if you launch flats in all filters simultaneously?  I supposed this could apply to any of the steps, but flats feel more parallelizable than the others.

czwa (approved):
I think most of my comments are me not fully understanding how the variables work here.

bps/templates/bps_biasBootstrap.yaml
erykoff: This is a bps variable, and must be quoted because yaml gets confused otherwise.

PR #293, opened 2025-01-24
DM-47079: Use TRIMMED_SUM as default pd integration method for Run 7 data

jchiang87:


erykoff (approved):
Looks good, but can you also update the docstring below that `photodiodeCurrentScale` also applies to `TRIMMED_SUM`?

PR #298, opened 2025-02-14
DM-48960: Remove deprecated element-wise numpy 64-bit conversion from cpCtiSolve

Alex-Broughton:


python/lsst/cp/pipe/cpCtiSolve.py
```diff
@@ -929,8 +929,9 @@ def calcCtiTurnoff(self, signalVec, dataVec, ctiRange, amp):
                              "to 0 electrons.")
             return 0.0, 0.0
 
-        turnoffIdx = np.argwhere(good)[-1]
-        turnoff = np.max(signalVec[good])
+        # turnoffIdx = np.argwhere(good)[-1]
+        turnoffIdx = good.size - np.argmax(good[::-1]) - 1
```
erykoff, line 933: I do not understand this line.

Alex-Broughton: This returns the last true value in the boolean array

erykoff (approved):

PR #299, opened 2025-02-18
DM-48955: Add illumination correction creation pipeline and BPS template files

erykoff:


pipelines/_ingredients/cpIlluminationCorrectionLSST.yaml
```diff
@@ -0,0 +1,67 @@
...
+      connections.stars: preSource
+      connections.exposure: calexp
+      connections.background: calexpBackground
+      python: |
+        # TODO[DM-47320]: make output catalog as similar to the old 'src' as
+        # possible; we'll try to go back to calibrateImage defaults with a
+        # smaller catalog after it's integrated.
+        # Goal is to remove all of these overrides except the aperture radii,
+        # which we instead want to set to the same as the compensated tophat
+        # radii below for lsst.fgcmcal.fgcmOutputProducts.FgcmOutputProductsTask
```
TallJimbo, line 22: It might be worth it to try to do the copy of this TODO for this variance of `calibrateImage` now: I was being really careful to rock the boat as little as possible when switching to `calibrateImage` in DRP so I could separate structural changes from algorithmic changes as much as possible.  But I was especially trying to make sure I didn't break any analysis tooling that was expecting a deeper `src` catalog than the default `calibrateImage` produces, and you don't have to worry about that here.  And we do know the default `calibrateImage` configs work fine in terms of robustness, since that's basically what AP runs.

TallJimbo (approved):

pipelines/_ingredients/cpIlluminationCorrectionLSST.yaml
erykoff: I can't run the variant; I need the fgcmcal inputs.  Though I can cut down the apertures.  But I tried removing plugins and it crashed because the transformation schema expected all these outputs and at that point I didn't want to open that can of hornets.

TallJimbo: Oh, yuck. It hadn't occurred to me before that doing the standardization before the consolidate step and doing the consolidate step before the association meant that we can't easily run variants of `calibrateImage` -> `{fgcm, gbdes}` without writing new functor files.  It's going on my list.

(My list of things to start daydreaming about ripping out and wholly replacing, that is; it's not quite the same as your List.)

For now, maybe just change the TODO comment to say that we can't let this configuration diverge meaningfully from the DRP one until we add a new functor file for it (and then make a ticket to add a new functor file).  I think that's actually moderately important, not because we care so much about running unnecessary algorithms here, but because having to sync this with the DRP configuration is problematic.

erykoff: I defined a new functor file and overrode the appropriate configs.

schemas/PreSourceIlluminationCorrection.yaml
```diff
@@ -0,0 +1,189 @@
...
+    raErr:
+        functor: RAErrColumn
+    decErr:
+        functor: DecErrColumn
+    ra_dec_Cov:
+        functor: RADecCovColumn
+    psfFlux:
+        functor: Column
+        args:
+            - slot_PsfFlux_flux
```
czwa, line 45: The others are all scalar entries, but these two are single element lists?  I see Gaussian fluxes are the same below, so maybe this is just how flux/flux errors work?

czwa (approved):

schemas/PreSourceIlluminationCorrection.yaml
erykoff: Yeah.  This is just copied from the official functor file, slimmed down for our needs.

PR #300, opened 2025-02-19
DM-47681: ITL edge bleed masking 

aferte:


python/lsst/cp/pipe/_configure.py
```diff
@@ -15,6 +15,7 @@ def configureIsrTaskLSSTForCalibrations(config):
     config.doCorrectGains = False
     config.doSaturation = False
     config.doSuspect = False
+    config.doITLEdgeBleedMask = False
```
erykoff, line 18: This needs to be moved to the new location in application order.  Before `doBias` I guess?

PR #301, opened 2025-03-16
DM-49175: Adjust configs/warnings for LSSTCam dense PTC

Alex-Broughton:


python/lsst/cp/pipe/ptc/cpPtcSolve.py
```diff
@@ -1211,6 +1211,7 @@ def errFunc(p, x, y):
                         & (np.abs(np.nan_to_num(sigResids)) < sigmaCutPtcOutliers)
                         & mask
                     )
+
```
erykoff, line 1214: Please remove this newline (there are no changes to this file on this PR)

python/lsst/cp/pipe/cpCtiSolve.py
```diff
@@ -699,7 +713,14 @@ def findTraps(self, inputMeasurements, calib, detector):
                     Nskipped += 1
             self.log.info(f"Skipped {Nskipped} exposures brighter than {self.config.maxImageMean}.")
             if len(signal) == 0 or len(data) == 0:
-                raise RuntimeError("All exposures brighter than config.maxImageMean and excluded.")
+                # There are no exposures left, so set trap so that
+                # there is no correction
+                self.log.warning("All exposures brighter than config.maxImageMean are excluded. "
+                                 "Setting zero-sized serial trap for amp {ampName}.")
```
erykoff, line 719: Please fix the warning to `"amp %s", ampName)`

python/lsst/cp/pipe/cpCtiSolve.py
```diff
@@ -433,7 +443,11 @@ def solveGlobalCti(self, inputMeasurements, calib, detector):
                     Nskipped += 1
             self.log.info(f"Skipped {Nskipped} exposures brighter than {self.config.maxSignalForCti}.")
             if len(signal) == 0 or len(data) == 0:
-                raise RuntimeError("All exposures brighter than config.maxSignalForCti and excluded.")
+                # There are no exposures left, set globalCTI to 0
+                self.log.warning("All exposures brighter than config.maxSignalForCti are excluded. "
+                                 f"Setting {ampName} global CTI to zero.")
```
erykoff, line 448: Please fix warning to `"Setting %s global CTI to zero.", ampName)`

python/lsst/cp/pipe/cpCtiSolve.py
```diff
@@ -314,7 +314,17 @@ def solveLocalOffsets(self, inputMeasurements, calib, detector):
                     Nskipped += 1
             self.log.info(f"Skipped {Nskipped} exposures brighter than {self.config.maxImageMean}.")
             if len(signal) == 0 or len(data) == 0:
-                raise RuntimeError("All exposures brighter than config.maxImageMean and excluded.")
+                # All exposures excluded, set the calibration so that
+                # there is no correction
+                self.log.warning("All exposures brighter than config.maxImageMean are excluded. "
+                                 "Setting local offset drift scale to zero for amp {ampName}.")
```
erykoff, line 320: Please fix warning to `"...for amp %s.", ampName`

bps/templates/bps_ptc.yaml
```diff
@@ -2,10 +2,17 @@ pipelineYaml: "${CP_PIPE_DIR}/pipelines/${INSTRUMENT}/cpPtc.yaml"
...
+  isrExposure:
+    pipetasks: cpPtcIsr
+    dimensions: exposure
 
 payload:
   payloadName: "${INSTRUMENT}_${TICKET}_ptc"
   output: "${USER_CALIB_PREFIX}${INSTRUMENT}/calib/${TICKET}/${TAG}/ptcGen.${RERUN}"
   butlerConfig: "${REPO}"
-  inCollection: "${USER_CALIB_PREFIX}${INSTRUMENT}/calib/${TICKET},${RAW_COLLECTION},${CALIB_COLLECTIONS}"
+  inCollection: ${USER_CALIB_PREFIX}${INSTRUMENT}/calib/${TICKET}/${TAG},${RAW_COLLECTION},${CALIB_COLLECTIONS}"
```
erykoff, line 17: You lost a quote here, it will crash.  Also this should not have the tag.

python/lsst/cp/pipe/cpCtiSolve.py
```diff
@@ -304,17 +304,29 @@ def solveLocalOffsets(self, inputMeasurements, calib, detector):
             # leaks into the overscan region.
             signal = []
             data = []
-            Nskipped = 0
+            NSkipped = 0
```
erykoff, line 307: `nSkipped`.  Must start lower case.

erykoff:
Please update the name of the PR to be more descriptive (e.g., add clustering to bps templates and LSSTCam config overrides)

bps/templates/bps_bias.yaml
```diff
@@ -1,7 +1,14 @@
 pipelineYaml: "${CP_PIPE_DIR}/pipelines/${INSTRUMENT}/cpBias.yaml"
 
+clusterAlgorithm: lsst.ctrl.bps.quantum_clustering_funcs.dimension_clustering
```
czwa, line 3: Is there a two sentence explanation for this?  My understanding was that this was needed for the PTC extract step to ensure that things get paired correctly, but I didn't think we needed it for ISR processing as well.  Is it as simple as "makes it go faster"?

bps/templates/bps_defects.yaml
```diff
@@ -2,10 +2,11 @@ pipelineYaml: "${CP_PIPE_DIR}/pipelines/${INSTRUMENT}/cpDefects.yaml"
 
 project: "${TICKET}"
 campaign: "${TICKET}"
+submitPath: "$SCRATCH/submit/{outputRun}"
 
-biasBootstrapRun: "${USER_CALIB_PREFIX}$INSTRUMENT/calib/$TICKET/$TAG/biasBootstrapGen.$RERUN/run${BOOTSTRAP_RUN_NUMBER}"
-darkBootstrapRun: "${USER_CALIB_PREFIX}$INSTRUMENT/calib/$TICKET/$TAG/darkBootstrapGen.$RERUN/run${BOOTSTRAP_RUN_NUMBER}"
-flatBootstrapRun: "${USER_CALIB_PREFIX}$INSTRUMENT/calib/$TICKET/$TAG/flatBootstrapGen.$RERUN/run${BOOTSTRAP_RUN_NUMBER}"
+biasBootstrapRun: "${USER_CALIB_PREFIX}$INSTRUMENT/calib/$TICKET/$TAG/biasBootstrapGen.${RERUN}/run${BOOTSTRAP_RUN_NUMBER}"
```
czwa, line 7: Do any of the `$RERUN` variables elsewhere need braces as well?  I see above in the `bps_darkBootstrap.yaml` file that we have `$RERUN`.  Similarly in the `bps_flatBootstrap.yaml`.

python/lsst/cp/pipe/cpBfk.py
```diff
@@ -440,34 +445,43 @@ def run(self, inputPtc, dummy, camera, inputDims):
                 # Get a mask for a 3px picture frame
                 valid = True
                 if kernelSize >= 7:
+                    # Last 1px
```
czwa, line 448: I think this needs an explanatory comment.  We use a 3-pixel radius "annulus" to calculate the threshold, but then only check the 1-pixel outer frame?

bps/templates/bps_bias.yaml
erykoff: Having a short explanation of the clustering configs would be great, yes!  But I do know that Jim Chiang told us "use these when doing any LSSTCam ISR tasks".

bps/templates/bps_defects.yaml
erykoff: I think that the braces are optional, but I meant to standardize on using it (like the quotes) in which case `$TICKET/$TAG` should be updated as well (here and below)

bps/templates/bps_darkBootstrap.yaml
```diff
@@ -2,6 +2,13 @@ pipelineYaml: "${CP_PIPE_DIR}/pipelines/${INSTRUMENT}/cpDarkBootstrap.yaml"
 
 project: "${TICKET}"
 campaign: "${TICKET}"
+submitPath: "$SCRATCH/submit/{outputRun}"
```
erykoff, line 5: If we are standardizing on braces then it should be `${SCRATCH}` here and elsewhere

erykoff (approved):
A couple of minor formatting issues, otherwise looks good

bps/templates/bps_bias.yaml
Alex-Broughton: It's there because ISR is so fast that HTCondor can't keep up when there are a lot of input exposures, and many quanta running in parallel. While other pipelines won't have as many input exposures as the PTC and linearizer pipelines, this is there to force it to cluster by exposure and specifically "slow it down" (it's not changing the total run time, just moving from many fast quanta to a small number of slower quanta). It also makes bps report faster since it only needs walk through less sub directories in submit/

I'll add a comment to the bps templates

PR #302, opened 2025-03-17
DM-49421: Add new PhotonTransferCurveExtractPairTask for parallelized extraction

erykoff:


python/lsst/cp/pipe/ptc/cpPtcExtract.py
```diff
@@ -79,16 +86,147 @@ class PhotonTransferCurveExtractConnections(pipeBase.PipelineTaskConnections,
...
+                adjuster.remove_quantum(quantumId)
+                exposures.remove(exposure)
+                quantumIds.remove(quantumId)
+                continue
+
+        if self.config.matchExposuresType == "TIME":
+            # Extract the exposure times.
+            exposureTimes = []
+            for dataId in quantumIds:
+                exposureTimes.append(adjuster.expand_quantum_data_id(dataId).exposure.exposure_time)
```
TallJimbo, line 126: Seems like this is going to get you one copy of every exposure time for every detector.

Unless I've missed some reason that's not the case, I think you might have an easier time working with dicts-of-lists with exposure ID keys, rather than flattened lists that are sorted by exposure, because it would you do some iterations one exposure at a time rather than one quantum at a time.

erykoff: Oh wait there may be a misunderstanding here on my part (very possible; even likely) that requires some significant refactoring on my part.  I've only tested this when solving per-detector, but this actually pools together *all* the detectors I guess?

TallJimbo: Yes, it will pool together all of the detectors.

erykoff: @TallJimbo I've done a big refactor to take this into account, please have another look.

@Alex-Broughton one of the surprising things is that there seem to be some exposures that are there for detector 117 and not for 76.  Is this expected? I think it's handling this all correctly, but we need to do a run over the 4 test detectors and ensure the final output is what we expect.

PR #303, opened 2025-03-19
DM-49497: Add check for valid BFK

Alex-Broughton:


python/lsst/cp/pipe/cpBfk.py
```diff
@@ -415,7 +415,31 @@ def run(self, inputPtc, dummy, camera, inputDims):
...
+                edges += (yv < 3)
+                edges += (yv > kernelSize - 3 - 1)
+
+                kernelEdges = bfk.ampKernels[ampName][edges].ravel()
+                kernelEdgeStd = np.std(kernelEdges)
+
+                # Edges should converge to zero
+                valid = np.all(np.isclose(kernelEdges, 0, rtol=5*kernelEdgeStd))
+                # Kernel should be negative
+                valid = np.all(bfk.ampKernels[ampName] <= 5*kernelEdgeStd)
```
erykoff, line 438: Should this be a `*=` here?

python/lsst/cp/pipe/cpBfk.py
```diff
@@ -415,7 +415,31 @@ def run(self, inputPtc, dummy, camera, inputDims):
...
-            bfk.valid[ampName] = True
+
+            # Check for validity
+            kernelSize = bfk.ampKernels[ampName].shape[0]
+            kernelCenterValue = bfk.ampKernels[ampName][kernelSize//2, kernelSize//2]
+
+            xv, yv = np.meshgrid(range(kernelSize), range(kernelSize))
+            # Get a mask for a 3px picture frame
+            valid = True
+            if kernelSize >= 7:
```
erykoff, line 426: Should any of this be configurable (the 3x3, the 7 size, the 5 times std) or you think these are fixed-fixed?

python/lsst/cp/pipe/cpBfk.py
Alex-Broughton: Yep! Thanks :)

python/lsst/cp/pipe/cpBfk.py
Alex-Broughton: The 5 sigma can be a config, but the 3px shouldn't be. The kernel should be just big enough to have 3 px on the edge that go to zero. 7 is then just the minimum size (3*2 + 1)

python/lsst/cp/pipe/cpBfk.py
```diff
@@ -415,7 +430,52 @@ def run(self, inputPtc, dummy, camera, inputDims):
...
+                    kernelEdgeStd = np.std(kernelEdges)
+
+                    # Edges should converge to zero
+                    valid = np.all(np.isclose(
+                        kernelEdges, 0,
+                        rtol=self.config.nSigmaTolForValidEdge*kernelEdgeStd,
+                    ))
+                    if not valid:
+                        self.log.info(f"{ampName}: the kernel edges did not converge to 0 within "
+                                      f"+/- {self.config.nSigmaTolForValidEdge*kernelEdgeStd} sigma")
```
erykoff, line 458: It is against the dev guide and python best practices to use f-strings with logging (I know, very annoying).
You have to do `self.log.info("%s: the kernel...", ampName, ...)`

Also, why is this not a warning?

python/lsst/cp/pipe/cpBfk.py
```diff
@@ -415,7 +430,52 @@ def run(self, inputPtc, dummy, camera, inputDims):
...
+                    if not np.all(bfk.ampKernels[ampName][~edges] <= 5*kernelEdgeStd):
+                        self.log.info(f"{ampName}: the kernel is not negative")
+
+                    # The center should be the minimum
+                    valid *= np.all(bfk.ampKernels[ampName] >= kernelCenterValue)
+                    if not np.all(bfk.ampKernels[ampName] >= kernelCenterValue):
+                        self.log.info(f"{ampName}: the kernel center is not the absolute minimum.")
+                else:
+                    self.log.info(f"{ampName}: The kernel is too small for validity check. Setting "
+                                  "bfk.valid=False.")
```
erykoff, line 471: This is a config failure, isn't it?  This should be checked in the config validation and not run if you combine these settings.  Basically a RuntimeError not an info log.

python/lsst/cp/pipe/cpBfk.py
```diff
@@ -415,7 +430,52 @@ def run(self, inputPtc, dummy, camera, inputDims):
...
+                    bfk.valid[ampName] = False
+
+                bfk.valid[ampName] = valid
```
erykoff, line 474: This will clobber that previous line, but we shouldn't need that check anyway and it should raise on config.

erykoff (changes requested):

erykoff (approved):

PR #307, opened 2025-04-07
DM-49926: Propagate MJDs of input exposure through ptcExtract

Alex-Broughton:


python/lsst/cp/pipe/ptc/cpPtcExtract.py
```diff
@@ -634,6 +639,14 @@ def _extractOneFlatPair(
         partialPtcDataset = PhotonTransferCurveDataset(
             ampNames, 'PARTIAL',
             covMatrixSide=self.config.maximumRangeCovariancesAstier)
+
+        # Get the following statistics for each amp
+        readNoise1 = dict()
```
erykoff, line 644: Why are these dicts, do we do anything with them?

PR #310, opened 2025-04-15
DM-49159: Add configuration to add flat source to flat metadata.

erykoff:
The test for this functionality is in ci_cpp_gen3

### DIFF
--- a/pipelines/LSSTComCam/cpFlat.yaml
+++ b/pipelines/LSSTComCam/cpFlat.yaml
@@ -7,3 +7,7 @@ tasks:
     class: lsst.ip.isr.IsrTaskLSST
     config:
       doDeferredCharge: true
+  cpFlatCombine:
+    class: lsst.cp.pipe.CalibCombineByFilterTask
+    config:
+      flatSource: "TWILIGHT"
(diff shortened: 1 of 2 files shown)


pipelines/LSSTComCam/cpFlat.yaml
```diff
@@ -7,3 +7,7 @@ tasks:
     class: lsst.ip.isr.IsrTaskLSST
     config:
       doDeferredCharge: true
+  cpFlatCombine:
+    class: lsst.cp.pipe.CalibCombineByFilterTask
+    config:
+      flatSource: "TWILIGHT"
```
cmsaunders, line 13: Is this the only possible option for comCam? (Sorry I missed all the comCam data taking!) If not I wouldn't set it here.

cmsaunders (approved):
I don't love the "backwards looking" (as you said in the slack thread) way this sets the metadata, since it seems pretty susceptible to user error, but it's reasonable in the circumstances. Would it be possible to put in a check (or TODO for one) to warn if the `flatSource` disagrees with the header? 

One other small comment below.

pipelines/LSSTComCam/cpFlat.yaml
erykoff: Yup.  We didn't have a flat field screen, and so we made twilight flats.  I also don't anticipate ComCam going back on the telescope anytime soon.
In a sense this is academic since we also aren't regenerating twilight flats for ComCam but I thought it would be useful to have this for completeness.

erykoff:
I've filed a ticket to update the code when we have finally converged on our metadata conventions for calibrations.

PR #312, opened 2025-04-17
DM-49847: Turn on CTI for PTC

Alex-Broughton:


bps/templates/bps_cti.yaml
```diff
@@ -15,5 +15,5 @@ payload:
   payloadName: "${INSTRUMENT}_${TICKET}_cti"
   output: "${USER_CALIB_PREFIX}${INSTRUMENT}/calib/${TICKET}/${TAG}/ctiGen.${RERUN}"
   butlerConfig: "${REPO}"
-  inCollection: "${USER_CALIB_PREFIX}${INSTRUMENT}/calib/${TICKET},${RAW_COLLECTION},${CALIB_COLLECTIONS}"
+  inCollection: "${LINEARIZER_GEN},${USER_CALIB_PREFIX}${INSTRUMENT}/calib/${TICKET},${RAW_COLLECTION},${CALIB_COLLECTIONS}"
```
erykoff: Is this the correct order; or should the `calib/TICKET` come first (it does everywhere else).  Also ... is `LINEARIZER_GEN` guaranteed to exist as an env var?
The problem is that we need the `linearizerPtc` I guess.  In which case does that need to be a certified calib ... I fear this is fragile.

python/lsst/cp/pipe/cpCtiSolve.py
```diff
@@ -514,8 +521,22 @@ def solveGlobalCti(self, inputMeasurements, calib, detector):
...
+            if np.isclose(10**result.params['ctiexp'].value, 10**(-5), rtol=.01):
+                self.log.warning(f"Global CTI for detector {detector.getId()} ({amp.getName()}) is "
+                                 "has hit the fitter's upper bound (10^-5).")
+
             # Only the global CTI term is retained from this fit.
-            calib.globalCti[ampName] = 10**result.params['ctiexp'].value
+            # If the CTI is within 1% of the lower bound, just set
+            # the CTI to zero and don't perform a global CTI
+            # correction.
+            if np.isclose(10**result.params['ctiexp'].value, 1e-8, rtol=0.01):
```
erykoff: I think double quotes here?

python/lsst/cp/pipe/cpCtiSolve.py
```diff
@@ -197,12 +201,16 @@ def run(self, inputMeasurements, camera, inputDims, inputPtc):
...
+                before calculating CTI statistics. (`float`).
         camera : `lsst.afw.cameraGeom.Camera`
             Camera geometry to use to find detectors.
         inputDims : `list` [`dict`]
             List of input dimensions from each input exposure.
-        inputPtc : `lsst.ip.isr.PhotonTransferCurveDataset`
-            Input PTC.
+        inputCalib : `lsst.ip.isr.PhotonTransferCurveDataset` or
+                     `lsst.ip.isr.Linearizer`
+            Input PTC or linearizer.
```
erykoff: Mention this is only used for metadata tracking.

python/lsst/cp/pipe/cpLinearitySolve.py
```diff
@@ -1427,6 +1427,7 @@ def run(
         for amp in detector:
             ampName = amp.getName()
 
+            linearizer.inputGain[ampName] = inputPtc.gain[ampName]
```
erykoff: I think you need to add this also to the other linearity solve task (not just the new relative one).

erykoff (approved):

PR #313, opened 2025-04-18
DM-50259: Turn off new ISR check functionality when doing calibration production.

erykoff:


python/lsst/cp/pipe/_configure.py
```diff
@@ -11,8 +13,10 @@ def configureIsrTaskLSSTForCalibrations(config):
     """
     # These are defined in application/run order.
     config.doBootstrap = False
+    config.doCheckUnprocessableData = False
     config.doDiffNonLinearCorrection = False
     config.doCorrectGains = False
+    config.ampNoiseThreshold = np.inf
```
Alex-Broughton, line 19: Do we want to set this to the current default, e.g. 25el?

erykoff: So when creating calibrations we want the default to be "off" because otherwise we can't measure the noise.  And we also have a policy that all pixels should be filled in with calibrations and so we don't want to pre-mask anything.

Alex-Broughton (approved):

PR #314, opened 2025-04-18
DM-50295: Add vignetting functionality for LSSTCam flats

erykoff:


.gitignore
```diff
@@ -24,3 +24,4 @@ bin/
 .pytest_cache
 pytest_session.txt
 python/*.dist-info/
+*~
```
czwa, line 27: I thought these were automatically suppressed by emacs (although maybe only if the file is added?).  No objection to including it if you ran into it.

pipelines/LSSTCam/cpFlat.yaml
```diff
@@ -2,3 +2,13 @@ description: cp_pipe LSSTCam flat calibration construction.
...
+tasks:
+  cpFlatCombine:
+    class: lsst.cp.pipe.CalibCombineByFilterTask
+    config:
+      doVignetteMask: true
+      # The radius is derived from the Batoid vignetting
+      # polynomial, with a threshold of 0.1 (90% vignetted).
+      # This corresponds to roughly 80% vignetting in the dome
+      # flats.
+      vignette.radius: 364.51
```
czwa, line 14: Do we know why there's a difference between Batoid and the dome flats?  Idealized perfect beams compared to a reality with complicated stray light?  I don't think it actually matters, but this is a larger difference than I would have naively expected.

pipelines/_ingredients/cpFlatLSST.yaml
```diff
@@ -14,11 +14,12 @@ tasks:
         config.crosstalk.doQuadraticCrosstalkCorrection = True
         config.doApplyGains = True
         config.doLinearize = True
-        # TODO DM-46426: Add cpCtiLSST pipeline so that this can be True.
-        config.doDeferredCharge = False
+        config.doDeferredCharge = True
         config.doBias = True
         config.doDark = True
-        config.doBrighterFatter = True
+        # TODO DM-50263: Investigate if this can be turned on, looking
```
czwa, line 20: Same typo question as on cp_verify.

python/lsst/cp/pipe/cpCombine.py
```diff
@@ -389,13 +411,21 @@ def run(self, inputExpHandles, inputScales=None, inputDims=None):
             polygon = inputExpHandles[0].get(component="validPolygon")
             maskVignettedRegion(combined, polygon=polygon, vignetteValue=0.0)
 
+        # Set the detector
+        combinedExp.setDetector(inputDetector)
+
+        if self.config.doVignetteMask:
```
czwa, line 417: I noticed during today's A&P meeting DM-47522: https://rubinobs.atlassian.net/browse/DM-47522 "Use VIGNETTED mask plane in ISR".  Looking at the code, it appears we hard code this to `NO_DATA` here: https://github.com/lsst/ip_isr/blob/main/python/lsst/ip/isr/vignette.py#L105, despite having a config option.  I think setting `NO_DATA` is correct, but setting `VIGNETTED` may also be useful for downstream purposes (a coadd could have regions that should be `NO_DATA`+`VIGNETTED` for pixels that fall on a detector but received no light and also regions with `NO_DATA` because that region was not covered by any detector).  
As we're likely going to be iterating flats for the near future, I don't think it's essential to get this fixed on this ticket, but I wanted to make sure this was discussed.
The comment on `ip_isr` was there because I didn't have a better place to put a reference to this review comment, but knew it was coming.

czwa (approved):
Some questions, and a reference to another vignetting ticket to be aware of.

.gitignore
erykoff: These are created by emacs and `git status` complains about them.  Because of all the pipelines I've worked on I have tons of these here, and it finally broke me.

pipelines/LSSTCam/cpFlat.yaml
erykoff: The batoid vignetting function was a polynomial fit to the focused light.  The flat tracks focused + scattered light (roughly along with the background) so I'm not surprised there's a difference here especially so far out where it's falling off very rapidly ... and I don't know how good a fit the Batoid polynomial was at the far radii.

PR #315, opened 2025-04-21
DM-50336: Add new PTC attributes for overscanMedian and overscanMedianAbsDev

Alex-Broughton:


python/lsst/cp/pipe/ptc/cpPtcSolve.py
```diff
@@ -30,6 +30,7 @@
 
 from scipy.signal import fftconvolve
 from scipy.optimize import least_squares
+from scipy.stats import median_abs_deviation as mad
```
erykoff, line 33: Can you just leave this as `median_abs_deviation`?

erykoff (changes requested):

bps/templates/bps_ptc.yaml
```diff
@@ -18,5 +18,5 @@ payload:
   payloadName: "${INSTRUMENT}_${TICKET}_ptc"
   output: "${USER_CALIB_PREFIX}${INSTRUMENT}/calib/${TICKET}/${TAG}/ptcGen.${RERUN}"
   butlerConfig: "${REPO}"
-  inCollection: "${USER_CALIB_PREFIX}${INSTRUMENT}/calib/${TICKET},${RAW_COLLECTION},${CALIB_COLLECTIONS}"
+  inCollection: "${RAW_COLLECTION},${CALIB_COLLECTIONS}"  # ${USER_CALIB_PREFIX}${INSTRUMENT}/calib/${TICKET}
```
erykoff, line 21: We need to discuss the right way of doing a bps run with these templates that doesn't involve doing the full stack.  (I have a hack that I don't love).  But please revert this change in the meantime.

erykoff (changes requested):

erykoff (approved):

PR #318, opened 2025-05-06
DM-50377: Add special defect handling for E2V midline break.

erykoff:


python/lsst/cp/pipe/cpDefects.py
```diff
@@ -188,6 +188,24 @@ class MeasureDefectsTaskConfig(pipeBase.PipelineTaskConfig,
              "then no saturated column filling will be performed."),
         default=-1,
     )
+    e2vMidlineBreakNRow = pexConfig.Field(
+        dtype=int,
+        doc="E2V midline break number of midline break rows at the bottom of the top amps "
+            "or the top of the bottom amps to always mask or ignore. Number of rows will "
+            "be twice this config value. Only used if detector is E2V type. Set to <=0 "
+            "to not treat the E2V midline break specially.",
```
Alex-Broughton, line 196: I think this setup is pretty confusing. I actually think that "IGNORE" should mean do nothing and `_e2vMidlineMark` should return and do nothing, "MEDIAN" should set it to the median, and "MASK" should set it to the illegal value.

I think it wouldn't make sense for someone to e.g. pass a negative or zero value for `e2vMidlineBreakNRow`, or at least if someone passes a zero, it should default to IGNORE. 

I think as-is this doc string is also a bit unclear on first read.

Also, is it the number of rows x 2 or is it 2n+1?

Alex-Broughton: i.e. is the midline read out with the top row of amplifiers or the bottom row

erykoff: Clearly my docstring needs a lot of work.  And now it's much better, thanks!

PR #327, opened 2025-06-30
DM-51234: Add two-LED (red/blue) flat pipelines for LSSTCam

erykoff:


bps/templates/bps_flat_singleled_white.yaml
```diff
@@ -0,0 +1,21 @@
...
+    dimensions: exposure
+
+payload:
+  payloadName: "${INSTRUMENT}_${TICKET}_flat_white_${LED_NAME}"
+  output: "${USER_CALIB_PREFIX}${INSTRUMENT}/calib/${TICKET}/${TAG}/flatGen-white.${RERUN}"
+  butlerConfig: "${REPO}"
+  inCollection: "${USER_CALIB_PREFIX}${INSTRUMENT}/calib/${TICKET},${RAW_COLLECTION},${CALIB_COLLECTIONS}"
+  dataQuery: "${SELECTION_FLAT_white} and band='white' and exposure.observation_reason='single_led_flat_${LED_NAME}'"
+
+extraQgraphOptions: "--config cpFlatCombine:connections.outputData=flat_${LED_NAME} --config cpFlatBin8:connections.inputExp=flat_${LED_NAME} --config cpFlatBin8:connections.outputExp=flat_${LED_NAME}_bin8 --config cpFlatMosaic8:connections.inputExps=flat_${LED_NAME}_bin8 --config cpFlatMosaic8:connections.outputData=flat_${LED_NAME}_mosaic8 --config cpFlatBin64:connections.inputExp=flat_${LED_NAME} --config cpFlatBin64:connections.outputExp=flat_${LED_NAME}_bin64 --config cpFlatMosaic64:connections.inputExps=flat_${LED_NAME}_bin64 --config cpFlatMosaic64:connections.outputData=flat_${LED_NAME}_mosaic64"
```
czwa, line 21: I thought we had templates for connections to handle things like this.

python/lsst/cp/pipe/cpCombine.py
```diff
@@ -860,3 +887,158 @@ def setFlatSource(self, exp):
...
+        default=(10000, 200),
+    )
+    distributionPercentiles = pexConfig.ListField(
+        dtype=float,
+        default=[0, 5, 16, 50, 84, 95, 100],
+        doc="Percentile levels to measure on the final combined calibration.",
+    )
+
+
+class CalibCombineTwoFlatsByFilterTask(CalibCombineTask):
```
czwa, line 961: I think I'm fine with this, but if the scaling was applied at the `cpFlatMeasure` step, then we could just use the standard `cpCombine`.  I was also thinking about an `SED-model` that would hold a set of weights for a given filter that would make this extendable to arbitrary "monochromatic" inputs.  Maybe that's more complicated than it needs to be for the two-input case.

czwa (approved):

bps/templates/bps_flat_singleled_white.yaml
erykoff: Oh, please explain!  I'd love to simplify this.  But this template is just for the "single-LED no-filter flats" which I don't think we plan on taking regularly.

python/lsst/cp/pipe/cpCombine.py
erykoff: I didn't want to get too far down the monochromatic line since I don't know how we're going to do that and don't want to guess.  But as commented on Jira due to the different gradients I think we're going to need to insert a gradient-removal step here.

PR #328, opened 2025-07-01
DM-51630: Add option to crop out strongly vignetted parts of amplifiers when running a PTC

erykoff:


pipelines/LSSTCam/cpBfk.yaml
```diff
@@ -11,6 +11,26 @@ tasks:
       photodiodeIntegrationMethod: MEAN
       auxiliaryHeaderKeys: ["TEMPAVG"]
       matchExposuresType: TIME
+      doVignetteFunctionRegionSelection: true
```
Alex-Broughton, line 14: Do we want to turn this on for measureCTI stats as well?

erykoff: You mean in `ip_isr` isrFunctions?  That may be useful, but definitely needs to be a different ticket and probably needs some additional thought because of the diagonal nature of the cut.

Alex-Broughton (approved):

PR #339, opened 2025-09-19
DM-52350: Add A23 BF Correction

Alex-Broughton:


pipelines/_ingredients/cpElectrostaticBFDistortionMatrixLSST.yaml
```diff
@@ -0,0 +1,89 @@
...
+      maximumRangeCovariancesAstierFullCovFit: 15
+  cpBfPtcAdjustGainRatios:
+    class: lsst.cp.pipe.ptc.PhotonTransferCurveAdjustGainRatiosTask
+    config:
+      connections.exposures: "cpBfIsrExp"
+      connections.input_ptc: "bfPtcUnadjusted"
+      connections.output_ptc: "bfPtc"
+  # TODO DM-46439: This can be renamed back to cpBfSolve when repos
+  # are cleaned up for the previous dimensionality error.
+  cpBfSolveX:
```
erykoff: Has this always been called X?  You can use new labels for the electrostatic mode.  In fact, you should; these are different and not substitutable tasks.

python/lsst/cp/pipe/cpElectrostaticBF.py
```diff
@@ -0,0 +1,498 @@
...
+        doc="Input BF PTC dataset.",
+        storageClass="PhotonTransferCurveDataset",
+        dimensions=("instrument", "detector"),
+        isCalibration=True,
+    )
+
+    output = cT.Output(
+        name="electroBfDistortionMatrix",
+        doc="Output measured brighter-fatter electrostatic model.",
+        storageClass="ElectrostaticBrighterFatterDistortionMatrix",
```
erykoff: This is incorrect.  It should be `IsrCalib` *everywhere* in the storage class (inputs/outputs/isr/etc)

czwa (approved):
As with ip_isr, I don't see any problems, other than docstrings.

PR #343, opened 2025-10-03
DM-52726: Remove unnecessary u-band flat specializations.

erykoff:


czwa (approved):
Always glad when special cases can be dropped.

PR #370, opened 2026-06-26
DM-55092: add ability to mask picture frame according to detector type

aferte:


python/lsst/cp/pipe/cpDefects.py
```diff
@@ -477,16 +499,32 @@ def _findHotAndColdPixels(self, exp):
...
-                    ampImg = ampImg[:-self.config.nPixBorderLeftRight, :, afwImage.LOCAL]
-            if self.config.nPixBorderUpDown:
-                if ampImg.getY0() == 0:
-                    ampImg = ampImg[:, self.config.nPixBorderUpDown:, afwImage.LOCAL]
-                else:
-                    ampImg = ampImg[:, :-self.config.nPixBorderUpDown, afwImage.LOCAL]
+            # and depending on detector type to mask picture frame effect
+            if detectorType == 'E2V':
+                nPixBorderLeftRight = self.config.nPixBorderLeftRightE2V
+            elif detectorType == 'ITL':
```
erykoff: The WF sensors will have `ITL-WF` and not `ITL` as the type.  So this should be an `in` comparison...

erykoff (commented):
We need tests for the WF sensor types, and all the ITL comparisons should be substring aware.

erykoff (approved):

